PR #2, opened 2023-11-13
[IMP] util/inherit.py: leverage the new ir.model.inherit model

KangOl:


KangOl:

KangOl:
@robodoo check

KangOl:
@robodoo rebase-ff r+

PR #3, opened 2023-11-14
record.py: regard noupdate in ensure_xmlid_match_record

cawo-odoo:
The change in f8c41cc7a4c8314dc31c6415a98be512c90c522c introduced additional `ir_model_data` updates for rows that point to an existing record that doesn't match the values used to lookup a suitable record, if such a matching record exists. It will update the row in that case not regarding the `noupdate` column.
    
Tune this behavior by adding an additional check that returns immediately with the original `res_id`, if an `ir_model_data` row for the xmlid exists that points to an existing record and the `noupdate` column is `True`. This way, the update for existing, non-matching records will only happen if `noupdate` is `False`.

Also, in a second commit, add a test for this.

This was https://github.com/odoo/upgrade/pull/4793, already approved there by @aj-fuentes but not merged in time for the utils extraction.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -782,10 +782,17 @@ class TestRecords(UnitTestCase):
         newtx1 = util.ref(cr, "base.TX1")
         self.assertEqual(newtx1, tx1.id)
 
-        # break one res_id
-        cr.execute("UPDATE ir_model_data SET res_id=%s WHERE module='base' AND name='TX1'", [tx2.id])
+        # break one res_id, with noupdate
+        cr.execute("UPDATE ir_model_data SET res_id=%s,noupdate=TRUE WHERE module='base' AND name='TX1'", [tx2.id])
 
-        # case: `base.TX1` points to ResCurrency(169) but doesn't match values {'name': 'TX3'}; no other matches found.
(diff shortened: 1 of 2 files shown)


cawo-odoo:
```
$ cd ~/src/upgrade ; rg -l ensure_xmlid_match_record | cut -d/ -f2 | sort -u | xargs echo upgradeci retry with always only
upgradeci retry with always only auth_signup base data_merge fleet helpdesk hr hr_payroll_account hr_recruitment l10n_be_hr_payroll l10n_de l10n_nl l10n_th l10n_vn mail mass_mailing mass_mailing_sms point_of_sale portal_anonymous product purchase sale_amazon sales_team stock stock_account training_base uom
```

KangOl:
I'm not sure whether this is something we actually want.
It goes against the intent of this function that is to *ensure* that the xmlid match some values.

aj-fuentes:
From what was discussed here https://github.com/odoo/upgrade/pull/4793#discussion_r1233857230
I think we can let it with force update always and revisit this if we actually see more reasons why it shouldn't be the case. In the original motivation the issue seemed to revolve around one particular _standard_ record that we could pre-check beforehand. I think it's indeed safer to let this function always "enforce" the values.

PR #4, opened 2023-11-15
[IMP] util/records: prepare arch for parser also for jsonb arch

cawo-odoo:
In a recent commit c7c95cfc8ee2f955413fd1f89cc6dd56a9d14b7a, some code to fixup views in preparation for the `lxml.etree` parser has been added to `edit_vew()`. So far, it only happens in the code path for non-jsonb columns. This patch introduces a sub-function and calls it from both the code branch for jsonb and non-jsonb.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -199,6 +199,11 @@ def edit_view(cr, xmlid=None, view_id=None, skip_if_not_noupdate=True, active=Tr
         )
         [arch] = cr.fetchone() or [None]
         if arch:
+
+            def parse(arch):
+                arch = arch.encode("utf-8") if isinstance(arch, unicode) else arch
+                return lxml.etree.fromstring(arch.replace(b"&#13;\n", b"\n"))
+
             if jsonb_column:
 
                 def get_trans_terms(value):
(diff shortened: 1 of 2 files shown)


aj-fuentes:
B904 should be ignored in pyproject.toml, it's Python2 incompatible

src/util/records.py
```diff
@@ -231,6 +229,18 @@ def get_trans_terms(value):
...
+def _parse(arch, view_id):
+    try:
+        arch = arch.encode("utf-8") if isinstance(arch, unicode) else arch
+        return lxml.etree.fromstring(arch.replace(b"&#13;\n", b"\n"))
+    except lxml.etree.XMLSyntaxError as e:
+        if e.msg.startswith("Opening and ending tag mismatch"):
+            # this view is already wrong, we don't change it
+            _logger.warning("Skipping edit of invalid view (id=%s):\n%s", view_id, e.msg)
+            raise Skip()
+        raise
```
KangOl, line 241: Why move this to the `edit_view` method?
Now every consumer also needs to handle the `Skip` exception.
I would let the `XMLSyntaxError` bubble up.

cawo-odoo: Well, what I said in the commit message, plus:
I did not find any caller handling `XMLSyntaxError` explicitly[^1] so far. So my feeling was that existing callers do not care what the exception is exactly, while new ones could make a distinction (or not) if it makes sense to them like it does for `adapt_domains` (or not). Maybe if we raise `Skip(str(e))`, the difference will be even smaller?

[^1]: `rg XMLSyntaxError` on all the scripts.

cawo-odoo: @KangOl, after discussing with @aj-fuentes I went back to the simpler patch from the original PR.

KangOl:
upgradeci retry with always only base report sale theme_common web website website_enterprise website_helpdesk website_sale

KangOl:
@robodoo rebase-ff r+

KangOl:
@robodoo rebase-ff r+

PR #5, opened 2023-11-16
[FIX] fields: Handle invalid domains

nseinlet:
Some filters are simply invalid, and leads to a traceback when removing fields in them.


src/util/fields.py
```diff
@@ -147,7 +147,12 @@ def clean_context(context):
         [model, r"\y{}\y".format(fieldname)],
     )
     for id_, name, context_s in cr.fetchall():
-        context = safe_eval(context_s or "{}", SelfPrintEvalContext(), nocopy=True)
+        try:
+            context = safe_eval(context_s or "{}", SelfPrintEvalContext(), nocopy=True)
+        except SyntaxError:
+            cr.execute("DELETE FROM ir_filters WHERE id = %s", [id_])    
+            add_to_migration_reports(("ir.filters", id_, name), "Filters/Dashboards")
```
KangOl, line 154: The report actually says that the filter has been corrected.
It also adds a 404 link to the removed record.
Also, why removing the filter? I would simply skip it (with a `logger.warning`).

src/util/fields.py
```diff
@@ -147,7 +147,12 @@ def clean_context(context):
         [model, r"\y{}\y".format(fieldname)],
     )
     for id_, name, context_s in cr.fetchall():
-        context = safe_eval(context_s or "{}", SelfPrintEvalContext(), nocopy=True)
+        try:
+            context = safe_eval(context_s or "{}", SelfPrintEvalContext(), nocopy=True)
+        except SyntaxError:
+            cr.execute("DELETE FROM ir_filters WHERE id = %s", [id_])
```
KangOl, line 153:
```suggestion
            cr.execute("DELETE FROM ir_filters WHERE id = %s", [id_])
```

Don't your editor automatically remove the trailing spaces?

KangOl:
upgradeci retry with always only hr project stock

PR #8, opened 2023-11-17
[FIX] util/orm.py: fix `custom_module_field_as_manual` transient access

erl-odoo:
From Odoo version 14.0 transient models require access rights instead of
implicit access (see related odoo/odoo#43306). This can lead to access
errors in `custom_module_field_as_manual`, particularly in the upgrade
TestCrawler, leading to blocked upgrades.

To reproduce:
- have a v13 db with a custom menu which opens a custom model's view,
having a relational field (e.g. one2many) to a custom transient model,
and have existing records
- run the upgrade test prepare
- upgrade to v14
- run the upgrade test check, the TestCrawler will fail with an
AccessError

To fix that, create temporary access rights for custom transient models
while using the context manager.

upg-1157201

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -424,6 +424,27 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
         )
         mock_x_email_fields = [r[0] for r in env.cr.fetchall()]
 
+    temp_ir_model_access_ids = []
+    if version_gte("14.0") and custom_models:
+        # 1.2 Version 14.0 and above requires new access rights for transient models,
+        # and the rights are yet to be added post standard upgrade for custom modules.
+        # To avoid access errors, create some temporary access rights for these models.
+        env.cr.execute(
+            """
+                INSERT INTO ir_model_access (name, model_id, active, perm_read)
+                     SELECT 'tmp_access', ir_model.id, true, true
+                       FROM ir_model
+                  LEFT JOIN ir_model_access
+                         ON ir_model_access.model_id = ir_model.id
+                      WHERE ir_model.model IN %s
+                        AND ir_model.transient
+                        AND ir_model_access.id IS NULL
+                  RETURNING id
+            """,
+            [custom_models],
+        )
+        temp_ir_model_access_ids = [r[0] for r in env.cr.fetchall()]
+
     # 2. Convert fields which are not in the registry to `manual` fields
     # and list the fields that were converted, to restore them back afterwards.
     # Also temporarily disable rules (ir.rule) that come from custom modules.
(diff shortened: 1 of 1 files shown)


erl-odoo:
@aj-fuentes 
Related to specific PR https://github.com/odoo/upgrade-specific/pull/9944, example access error in comments there

src/util/orm.py
```diff
@@ -424,6 +426,37 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
...
+        # 1.2 Version 14.0 and above requires new access rights for transient models,
+        # and the rights are yet to be added post standard upgrade for custom modules.
+        # To avoid access errors, create some temporary access rights for these models.
+        env.cr.execute(
+            """
+                SELECT id
+                  FROM ir_model
+                 WHERE model IN %s
+                   AND transient
+                   AND id NOT IN (
```
aj-fuentes, line 440: use a left join for this

src/util/orm.py
```diff
@@ -424,6 +426,37 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
...
+                            WHERE active
+                       )
+            """,
+            [custom_models],
+        )
+
+        no_access_custom_transient_ids = env.cr.fetchall()
+
+        if no_access_custom_transient_ids:
+            insert_query = """
```
aj-fuentes, line 452: this should be unified into the previous query (with a cte), return only the created ids.

aj-fuentes (changes requested):

KangOl:

src/util/orm.py
erl-odoo: indeed looks much better, thanks

src/util/orm.py
```diff
@@ -424,6 +424,28 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
...
+        # and the rights are yet to be added post standard upgrade for custom modules.
+        # To avoid access errors, create some temporary access rights for these models.
+        env.cr.execute(
+            """
+                INSERT INTO ir_model_access (
+                                name, model_id, active, perm_read, perm_write, perm_create, perm_unlink
+                            )
+                     SELECT 'tmp_access', ir_model.id, true, true, true, true, true
+                       FROM ir_model
+                  LEFT JOIN ir_model_access ON ir_model_access.model_id = ir_model.id
```
aj-fuentes, line 439:
```suggestion
                  LEFT JOIN ir_model_access 
                         ON ir_model_access.model_id = ir_model.id
```

src/util/orm.py
```diff
@@ -424,6 +424,28 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
...
+    if version_gte("14.0") and custom_models:
+        # 1.2 Version 14.0 and above requires new access rights for transient models,
+        # and the rights are yet to be added post standard upgrade for custom modules.
+        # To avoid access errors, create some temporary access rights for these models.
+        env.cr.execute(
+            """
+                INSERT INTO ir_model_access (
+                                name, model_id, active, perm_read, perm_write, perm_create, perm_unlink
+                            )
+                     SELECT 'tmp_access', ir_model.id, true, true, true, true, true
```
aj-fuentes, line 437: As discussed in DM. I wonder if this would be enough
```suggestion
                     SELECT 'tmp_access', ir_model.id, true, false, false, false, false
```
@KangOl any advice here?

KangOl: Yes, it should be enough. It's also safer.

erl-odoo: For the TestCrawler to pass, it is enough indeed, I've tested up to v16. But in that case I dropped the `false`s as it's not necessary.

aj-fuentes (approved):
LGTM @KangOl ?

KangOl:
@robodoo r+

PR #9, opened 2023-11-17
[IMP] util/{records,domains}: improve edit_view context manager

aj-fuentes:
The generator based context manager, using `yield`, has multiple shortcomings. First it requires an extra `skippable_cm` magic to be able to deal with missing views. Second, there was not clear way to signal that we want to skip the edition of the view to avoid spurious writes if in the end the view is unchanged.

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -6,12 +6,6 @@ import re
 import lxml
 
 try:
-    from contextlib import suppress
-except ImportError:
-    # python2 code, use the openerp vendor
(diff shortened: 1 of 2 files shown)


KangOl:
upgradeci retry with always only base report sale theme_common web website website_enterprise website_helpdesk website_sale

KangOl:
It doesn't skip the `with` body anymore. It just now bind the return value to `None` and ignore the "NoneType has no attribute" errors that are raised.

PR #10, opened 2023-11-20
[IMP] util/testing: improve skipping tests for demo data

aj-fuentes:
Now `skip_if_demo` is both a decorator and a method.

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -90,6 +90,19 @@ class UnitTestCase(TransactionCase, _create_meta(10, "upgrade_unit")):
             util.ENVIRON["__base_version"] = parse_version(bv)
 
 
+def skip_if_demo(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        assert args
+        self = args[0]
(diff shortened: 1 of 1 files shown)


src/testing.py
```diff
@@ -90,6 +90,19 @@ def setUpClass(cls):
             util.ENVIRON["__base_version"] = parse_version(bv)
 
 
+def skip_if_demo(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        assert args
+        self = args[0]
```
KangOl, line 97:
```suggestion
    def wrapper(self, *args, **kwargs):
        assert isinstance(self, UpgradeCommon)
```

src/testing.py
```diff
@@ -90,6 +90,19 @@ def setUpClass(cls):
...
+        self.env.cr.execute("SELECT 1 FROM ir_module_module WHERE name='base' AND demo")
+        if self.env.cr.rowcount:
+            self.skipTest("This test is not guaranteed with demo data.")
+        return f(*args, **kwargs)
```
KangOl, line 101:
```suggestion
        return f(self, *args, **kwargs)
```

KangOl (changes requested):

PR #11, opened 2023-11-22
[FIX] records: remove double quote_ident

jjmaksoud:
The results of get_fk are quote_idented by [default](https://github.com/odoo/upgrade-util/blob/7681facc4492d71720efaf587aaa10b03ae9b154/src/util/pg.py#L435). Adding more quotes may result in sql syntax errors when the table names or column names are not standard.

affects upg-1172577
```
psycopg2.errors.SyntaxError: zero-length delimited identifier at or near """"
LINE 1: SELECT COUNT(*) FROM "groups_team" WHERE ""groupID"" = 79
                                                 ^
```

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -505,7 +505,7 @@ def remove_group(cr, xml_id=None, group_id=None):
         return
 
     # Get all fks from table res_groups
-    fks = get_fk(cr, "res_groups")
+    fks = get_fk(cr, "res_groups", quote_ident=False)
 
     # Remove records referencing the group_id from the referencing tables (restrict fks)
     standard_tables = ["ir_model_access", "rule_group_rel"]


diagnoza (approved):

src/util/records.py
```diff
@@ -514,15 +514,15 @@ def remove_group(cr, xml_id=None, group_id=None):
...
                 count = cr.fetchone()[0]
                 if count:
                     custom_tables.append((foreign_table, foreign_column, count))
                 continue
 
-            query = 'DELETE FROM "{}" WHERE "{}" = %s'.format(foreign_table, foreign_column)
+            query = "DELETE FROM {} WHERE {} = %s".format(foreign_table, foreign_column)
             query = cr.mogrify(query, (group_id,)).decode()
 
             if column_exists(cr, foreign_table, "id"):
```
KangOl, line 528: Will be always false for quoted table names.
We have multiple option here:
 - ignore it and don't explode the query in this case.
 - unquote the table name
 - call `get_fk` with `quote_ident=False`

diagnoza: What about using identifiers from `psycopg2.sql`?

KangOl: It's a more intrusive change that needs more tests that this simple patch.
But feel free to start experimenting.

KangOl:
@robodoo r+

PR #13, opened 2023-11-24
[FIX] records: skip inactive, invalid views in adapt_domains()

cawo-odoo:
Do not fail when the XML parser fails to parse an inactive view. Inactive views are probably abandoned and do not cause immediate errors in the upgraded db, so just skip them when they are invalid.

Also, log the view_id before (re)raising a parser exception. This should save some time when manual fixing of invalid (active) views is needed.

upg-1171125

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -309,8 +309,8 @@ def adapt_domains(cr, model, old, new, adapter=None, skip_inherit=(), force_adap
         if column_exists(cr, "ir_ui_view", "arch_db")
         else "arch"
     )
-    cr.execute("SELECT id, model FROM ir_ui_view WHERE {} ~ %s".format(arch_db), [match_old])
-    for view_id, view_model in cr.fetchall():
+    cr.execute("SELECT id, model, active FROM ir_ui_view WHERE {} ~ %s".format(arch_db), [match_old])
+    for view_id, view_model, view_active in cr.fetchall():
         # Note: active=None is important to not reactivate views!
         try:
             with suppress(_Skip), edit_view(cr, view_id=view_id, active=None) as view:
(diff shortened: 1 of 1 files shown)


src/util/domains.py
```diff
@@ -354,6 +354,7 @@ def adapt_domains(cr, model, old, new, adapter=None, skip_inherit=(), force_adap
                 # this view is already wrong, we don't change it
                 _logger.warning("Skipping domain adpatation for invalid view (id=%s):\n%s", view_id, e.msg)
                 continue
+            _logger.exception("Failing domain adpatation of invalid view (id=%s)", view_id)
```
aj-fuentes, line 357: There was a typo here and the warning above: `adpatation`
Sill I'm not sure if we want to log the whole error... the view is already broken, we are not supposed to fix it. We fix views that are correct but do not work due to something done during the upgrade.

src/util/domains.py
```diff
@@ -350,10 +350,16 @@ def adapt_domains(cr, model, old, new, adapter=None, skip_inherit=(), force_adap
...
                 # this view is already wrong, we don't change it
-                _logger.warning("Skipping domain adpatation for invalid view (id=%s):\n%s", view_id, e.msg)
+                _logger.warning(
+                    "Skipping domain adaptation for %sinvalid view (id=%s):\n%s",
+                    "" if view_active else "inactive, ",
+                    view_id,
+                    e.msg,
+                )
                 continue
+            _logger.error("Failing domain adaptation of invalid view (id=%s)", view_id)  # noqa: TRY400
```
aj-fuentes, line 362:
```suggestion
            _logger.error("Cannot adapt domain of invalid view (id=%s)", view_id)  # noqa: TRY400
```

KangOl:

aj-fuentes:
@robodoo r+

PR #14, opened 2023-11-27
[FIX] util/const : track rename field for view_fixer

thni-odoo:
This store the field name being renamed using  update_field_usage/references method for the view_fixer
we store field name in a temporary constant __temp_rename_update_ref and once the old field is removed we transfer it into __renamed_fields to add a safety check also avoid any unforeseen issues.


aj-fuentes:
As we discussed in DM: I'm not sure this is a safe assumption to make.

I don't think we can safely assume that using `update_field_references` + `remove_field` is the same as `rename_field`.

@KangOl wdyt?

PR #18, opened 2023-12-04
[FIX] util/fields : update field in context of custom search view

thni-odoo:
This commit aims to make views compatible which are generated when a search filter is created using studio, view looks like
```xml
<data>
  <xpath expr="xyz" position="after">
    <filter name="filter1"  context="{'group_by': 'renamed field'}"/>
    <filter name="filter2"  context="{'group_by': 'removed field'}"/>
  </xpath>
</data>
```
Currently after upgrade this view will be disabled because we are not updating the field in this view.

After this fix

view:
```xml
<data>
  <xpath expr="xyz" position="after">
    <filter name="filter1"  context="{'group_by': 'New Name'}"/>
    <filter name="filter2"  context="{}"/>
  </xpath>
</data>
```
idk whether update_custom_filter is in correct file any suggestion where it should be ?


KangOl:

src/util/fields.py
```diff
@@ -1121,3 +1138,48 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
...
+               AND active = 't'
+        """.format(
+            arch_db
+        ),
+        [match_old, model],
+    )
+    for view_id in cr.fetchall():
+        try:
+            with suppress(_Skip), edit_view(cr, view_id=view_id, active=None) as view:
+                for elem in view.xpath("//filter[contains(@context, '{0}')]".format(old)):
```
aj-fuentes, line 1166: Should be moved around here https://github.com/odoo/upgrade-util/blob/3c7205a11504e0e018ca48ca6d0df65a2152f7ce/src/util/domains.py#L338

KangOl: Yes, don't move it to `domains.py`.

src/util/fields.py
```diff
@@ -1121,3 +1138,48 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
...
+                                "Field %s was removed during the upgrade, context has been updated from filter %s to avoid disabling of custom view",
+                                old,
+                                elem.attrib["name"],
+                            )
+                        elem.attrib["context"] = unicode(context)
+        except lxml.etree.XMLSyntaxError as e:
+            if e.msg.startswith("Opening and ending tag mismatch"):
+                # this view is already wrong, we don't change it
+                _logger.warning("Skipping custom filter adaptaiton for invalid view (id=%s):\n%s", view_id, e.msg)
+                continue
```
KangOl, line 1184: Again, code copy/pasted from `domains.py`. Should be factorized.

KangOl (changes requested):

src/util/fields.py
```diff
@@ -1121,3 +1138,48 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
         ) % {"src_model": src_model, "dst_model": dst_model, "actions": ", ".join(action_names)}
 
         add_to_migration_reports(message=msg, category="Server Actions")
+
+
+def update_custom_filter(cr, old, model, new=None):
```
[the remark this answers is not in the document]
thni-odoo: Can you please provide a reference for callback method, I didn't get it.

src/util/fields.py
```diff
@@ -1121,3 +1138,48 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
...
+    )
+    match_old = r"\y{}\y".format(re.escape(old))
+    cr.execute(
+        """
+            SELECT id
+              FROM ir_ui_view
+             WHERE {} ~ %s
+               AND type = 'search'
+               AND model = %s
+               AND active = 't'
```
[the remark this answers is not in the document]
thni-odoo: I wanted to avoid False negative cases.

PR #21, opened 2023-12-11
[FIX] util/fields: update depends on rename

thni-odoo:
Currently we are not updating the dependencies for custom compute fields which leads to test crawler failure when it tries to access the view and give: KeyError: 'old_field_name'

Full error 
```bash
2023-12-11 06:32:37,756 926 ERROR thni_14.0 odoo.addons.base.maintenance.migrations.base.testsodoo.upgrade.base.tests.test_mock_crawl: Adding menu ('account.menu_action_move_out_invoice_type', 126, 'Accounting > Customers > Invoices', 206) to the failing menus 
  File "/tmp/tmpmzpbd93k/migrations/base/tests/test_mock_crawl.py", line 238, in crawl_menu
    self.mock_action(action_vals)
  File "/tmp/tmpmzpbd93k/migrations/base/tests/test_mock_crawl.py", line 394, in mock_action
    mock_method(model, view, fields_list, domain, group_by)
  File "/tmp/tmpmzpbd93k/migrations/base/tests/test_mock_crawl.py", line 423, in mock_view_form
    [data] = record.read(fields_list)
  File "/home/odoo/src/odoo/14.0/odoo/models.py", line 3024, in read
    f = self._fields[dotname.split('.')[0]]
KeyError: 'invoice_payment_state'

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -1011,6 +1011,7 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
             # skip all inherit, they will be handled by the resursive call
             adapt_domains(cr, model, old, new, adapter=domain_adapter, skip_inherit="*", force_adapt=True)
             adapt_related(cr, model, old, new, skip_inherit="*")
+            adapt_depends(cr, model, old, new, skip_inherit="*")
 
         inherited_models = tuple(
             inh.model for model in only_models for inh in for_each_inherit(cr, model, skip_inherit)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by KangOl** [comment]
Don't inject variables into the SQL via `.format()`.
And you can't blindly replace the value into the `depends` column. It's a comma separated list of dotted path to fields.

**Comment by KangOl** [comment]

**Comment by thni-odoo** [comment]
@KangOl  Changes Done!

**Comment by KangOl** [comment]
No. As said, you can't just blindly replace the matching part without verifying it apply on the correct model.

Actually, it should be part of the `_update_field_usage_multi` function, like done with `adapt_related`. I don't think a dedicated method is needed, but you should use the same technique as `adapt_related` and use `_adapt_one_domain` to correctly rename the field names.

**Comment by thni-odoo** [comment]
upgradeci retry

**Comment by thni-odoo** [comment]
@KangOl Changes are done as suggested currently i have made a method can you please suggest me where to embed the code.

**Comment by thni-odoo** [comment]
@KangOl Gentle Reminder !

File: src/util/fields.py [hunk 1]
```diff
@@ -1008,6 +1008,34 @@ def adapt_dict(d):
             )
 
 
+# adapt depends for custom compute fields
+def adapt_depends(cr, model, old, new):
```
KangOl, line 1012: - missing loop on inherits
 - verify the model

```suggestion
def adapt_depends(cr, model, old, new, skip_inherit=()):
    _validate_model(model)
```

src/util/fields.py
```diff
@@ -1008,6 +1008,34 @@ def adapt_dict(d):
...
+         WHERE state = 'manual'
+           AND depends ~ %s
+        """,
+        [match_old],
+    )
+    for id_, model, depends in cr.fetchall():
+        temp_depends = depends.split(",")
+        for i in range(len(temp_depends)):
+            domain = _adapt_one_domain(cr, target_model, old, new, model, [(temp_depends[i], "=", "depends")])
+            temp_depends[i] = domain[0][0] if domain else temp_depends[i]
```
KangOl, line 1033: Useless write itself.
```suggestion
            if domain:
                temp_depends[i] = domain[0][0]
```

KangOl (changes requested):

thni-odoo:
@KangOl Suggested Changes Done!

thni-odoo:
@KangOl Gentle Reminder!

thni-odoo:
@KangOl Gentle Reminder!

KangOl:
@robodoo r+

PR #22, opened 2023-12-11
[FIX] util.rename_xmlid: also rename `_inherits` xmlids

KangOl:


src/util/records.py
```diff
@@ -597,6 +597,17 @@ def rename_xmlid(cr, old, new, noupdate=None, on_collision="fail"):
...
+        for inh in inherit_parents(cr, model, recursive=False):
+            if inh.via:
+                parent = inh.model.replace(".", "_")
+                rename_xmlid(
+                    cr,
+                    "{}_{}".format(old, parent),
+                    "{}_{}".format(new, parent),
+                    noupdate=noupdate,
+                    on_collision=on_collision,
+                )
```
aj-fuentes, line 610: Do we need the `recursive=False` only for this?
```suggestion
        inh = next(inherit_parents(cr, model), None)
        if inh and inh.via:
            parent = inh.model.replace(".", "_")
            rename_xmlid(
                cr,
                "{}_{}".format(old, parent),
                "{}_{}".format(new, parent),
                noupdate=noupdate,
                on_collision=on_collision,
            )
```

aj-fuentes: Actually from the error in runbot maybe `parent = next(inherit_parents(cr, model), None)` is enough?

KangOl: Nothing forbid models to have multiple `_inherits`.

KangOl:

src/util/records.py
aj-fuentes: I see...

aj-fuentes:
upgradeci retry

aj-fuentes (approved):
LGTM

aj-fuentes:
upgradeci retry with always only crm in versions 7.0 8.0 9.0 10.0 11.0

KangOl:
However, it bothers me that I have to change the API of the method.
Maybe another method should be added.

aj-fuentes (approved):

KangOl:
@robodoo rebase-ff r+

PR #23, opened 2023-12-12
[REF] util.views: refactor converter code, regroup views helpers

andreabak:
Overview of changes (see commits for details):
- refactored BootstrapConverter code, separating conversion code and operations from bootstrap-specific code. This way other kind of converters can be implemented, reusing pre-defined ElementOperations into lists (eg. ReplaceClasses/RenameAttributes/etc.)
  - this will also be useful in the (near) future to consolidate into util similar helpers in PS [custom-util](https://github.com/odoo-ps/custom-util/blob/master/migrations/custom_util/views/operations.py), commonly used to quickly fix studio and COW views (without all the manual lxml operations boilerplate)
- generalized helpers and moved from odoo/upgrade `base/16.0/pre-90-convert-bootstrap5.py` so that they can be reused for other similar kind of operations using converters (ie. apply converter to all html fields / templates, etc.)
- re-organized views-related helpers into a subpackage `util.views` (converter, bootstrap, records)
- misc fixes and tweaks


andreabak:
upgradeci retry with always all modules in all versions

KangOl:
upgradeci skip

KangOl:
upgradeci retry with always only base mass_mailing point_of_sale test_themes theme_common theme_treehouse web web_editor website website_blog website_enterprise website_event website_mass_mailing website_sale in all versions

src/util/views/records.py
```diff
@@ -0,0 +1,239 @@
...
+try:
+    from odoo.tools.translate import xml_translate
+except ImportError:
+    xml_translate = lambda callback, value: value
+
+from ..pg import column_exists, column_type
+from ..report import add_to_migration_reports
+
+_logger = logging.getLogger(__name__)
+
```
KangOl, line 23: __all__ = ["remove_view", "edit_view", "add_view"]

src/util/views/__init__.py
```diff
@@ -0,0 +1 @@
+from .records import *
```
KangOl, line 1:
```suggestion
```

src/util/records.py
```diff
@@ -12,12 +11,12 @@
     from odoo import release
     from odoo.tools.convert import xml_import
     from odoo.tools.misc import file_open
-    from odoo.tools.translate import xml_translate
 except ImportError:
     from openerp import release
     from openerp.tools.convert import xml_import
     from openerp.tools.misc import file_open
 
+from . import views
```
KangOl, line 19:
```suggestion
from .views.records import *
```

src/util/modules.py
```diff
@@ -39,7 +39,8 @@
 from .models import delete_model
 from .orm import env, flush
 from .pg import column_exists, table_exists, target_of
-from .records import ref, remove_menus, remove_records, remove_view, replace_record_references_batch
+from .records import ref, remove_menus, remove_records, replace_record_references_batch
+from .views import remove_view
```
KangOl, line 43:
```suggestion
from .views.records import remove_view
```

src/util/__init__.py
```diff
@@ -16,3 +16,4 @@
 from .records import *
 from .report import *
 from .specific import *
+from .views import *
```
KangOl, line 19:
```suggestion
```

src/util/domains.py
```diff
@@ -36,7 +36,7 @@
 from .inherit import for_each_inherit
 from .misc import SelfPrintEvalContext
 from .pg import column_exists, get_value_or_en_translation, table_exists
-from .records import edit_view
+from .views import edit_view
```
KangOl, line 39:
```suggestion
from .views.records import edit_view
```

KangOl (changes requested):

src/util/views/records.py
aj-fuentes: `__all__` for the record.

src/util/convert_bootstrap.py
KangOl, line 1: As shown by [the matt errors](https://upgradeci.odoo.com/upgradeci/run/122222), deleting this file is not retro-compatible.
I would keep a `convert_bootstrap.py` file which reexport functions from `.views.bootstrap` and emit a deprecation warning.

andreabak: Yeah, I just saw. I wasn't aware of the new saas~17.2 script using it, and forgot to re-check usages while rebasing this old commit from the linked upgrade PR https://github.com/odoo/upgrade/pull/5431.

But considering it's just a couple references in this repo, and possibly very few (to none) in third party code (since util was private until recently), why not just fix in the PR above? (so breaking compatibilty)

src/util/records.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
+# ruff: noqa: F403, F405  # ruff is unable to detect * imports
```
KangOl, line 2: This is unfortunate. This avoids catching real undefined local variable errors.
```suggestion
```

src/util/records.py
```diff
@@ -42,7 +41,7 @@
     table_exists,
     target_of,
 )
-from .report import add_to_migration_reports
+from .views.records import *
```
KangOl, line 44:
```suggestion
from .views.records import add_view, edit_view, remove_view
```

KangOl (changes requested):

src/util/records.py
andreabak: True. AFAIK Ruff still doesn't support multi-file analysis :disappointed:, hopefully cases like these will be fixed in the future.

src/util/records.py
andreabak: Doesn't this somewhat defeat the * import + `__all__` purpose though?
The alternative is that I add the noqa to every single line where the imported names are used, which is kinda spammy.

src/util/convert_bootstrap.py
KangOl: The saas~17.2 script will still need to be corrected to avoid the deprecation warning (as the runbot won't like it).

However, I would still have a reexport, as `convert_bootstrap` may be used in `upgrade-specific` and `odoo-ps` scripts.

andreabak: I will fix the saas~17.2 script import in https://github.com/odoo/upgrade/pull/5431.

I've found 2 in `upgrade-specific`, >6 mo old, I could fix those too.

I've also checked references in odoo-ps repos and there are none. (in general for PS upgrades it's less of a problem, we closely follow the projects upgrades, so breaking changes will be noticed, and also I can communicate with the whole team to adapt their code if necessary. But that's not the case here).

Still no? :innocent:

src/util/records.py
KangOl: > Doesn't this somewhat defeat the * import + `__all__` purpose though?

Yes, `__all__` is not strictly needed then. But we can keep it.

src/util/convert_bootstrap.py
KangOl: It will also avoid all the upgrade PR to be rebased to pass the CI

andreabak:
upgradeci retry

andreabak:
upgradeci retry

andreabak:
@KangOl I've rebased this on the latest master commit, resolved conflicts, etc.. CI checks seem good.
Is it ready to go? (together with https://github.com/odoo/upgrade/pull/5431)

FYI @aj-fuentes

PR #24, opened 2023-12-13
[FIX] adapt_domains: compatibility with old versions

KangOl:
The `active` column on `ir_ui_view` doesn't exists before Odoo 9.0


KangOl:
upgradeci retry with always only event in 8.0 17.0

aj-fuentes (approved):

KangOl:
@robodoo r+

PR #27, opened 2023-12-20
[IMP] util.alter_column_type

KangOl:
Detect the case where there is a lot of rows with a NULL value and alwasy change the type using a temporary column.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -435,27 +435,19 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
         using = "{{0}}::{}".format(type)
 
     # else, create a new column and parallel update queries.
-    cr.execute(
-        """
-        ALTER TABLE "{table}" RENAME COLUMN "{column}" TO "_{column}_upg";
-        ALTER TABLE "{table}" ADD COLUMN "{column}" {type};
-        """.format(**locals())
-    )
-    using = using.format('"_{}_upg"'.format(column))
-    parallel_execute(
+    tmp_column = "_{}_upg".format(column)
+    cr.execute(format_query(cr, "ALTER TABLE {} RENAME COLUMN {} TO {}", table, column, tmp_column))
+    cr.execute(format_query(cr, "ALTER TABLE {} ADD COLUMN {} {}", table, column, sql.SQL(type)))
+
+    using = sql.SQL(format_query(cr, using, tmp_column))
+    explode_execute(
         cr,
-        explode_query_range(
-            cr,
-            """
-            UPDATE "{table}"
-               SET "{column}" = {using}
-             WHERE "_{column}_upg" IS NOT NULL
-            """.format(**locals()),
-            table=table,
-        ),
+        format_query(cr, "UPDATE {} SET {} = {} WHERE {} IS NOT NULL", table, column, using, tmp_column),
+        table=table,
         logger=logger,
     )
-    cr.execute('ALTER TABLE "{table}" DROP COLUMN "_{column}_upg" CASCADE'.format(**locals()))
+
+    cr.execute(format_query(cr, "ALTER TABLE {} DROP COLUMN {} CASCADE", table, tmp_column))
 
 
 def table_exists(cr, table):
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -420,9 +420,19 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
             [model_of_table(cr, table), column],
         )
     if not using:
-        # Simple case. Use general SQL syntax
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only account delivery_sendcloud industry_fsm mail project

src/util/pg.py
```diff
@@ -361,9 +361,18 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
...
     if not using:
-        # Simple case. Use general SQL syntax
-        cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" TYPE %s' % (table, column, type))
-        return
+        # if there is a high number of NULL entries, it will be faster to just ignore those
+        cr.execute(
+            "SELECT null_fract FROM pg_stats WHERE schemaname = current_schema() AND tablename = %s AND attname = %s",
+            [table, column],
+        )
+        [null_frac] = cr.fetchone()
```
aj-fuentes, line 369: `pg_stats` may return nothing.
```suggestion
        [null_frac] = cr.fetchone() or (0.0,)
```

KangOl: Only if the column does not exist. And in this case, that's the next query that will fail.

KangOl: Actually, it may be missing without calling `ANALYZE` before.

cawo-odoo: The ANALYZE by itself can be expensive. Usually, we call ANALYZE on the whole DB very early in the upgrade and stats should already be there. How about only doing it here, if there are no stats for the column already?

aj-fuentes: > Actually, it may be missing without calling ANALYZE before.

Yes, and even with a full ANALYZE at the beginning of the upgrade, newly created tables/columns are not reflected in pg_stats until next checks.

I think analyzing the column is a good approach. We are going to perform a potentially heavy operation on the column anyways but the question is: would this defeat the purpose of this optimization? @KangOl I guess you have some test run we can get more data from. If the analyze is as expensive as calling alter type then maybe a good compromise is analyzing only if there is nothing in pg_stats as @cawo-odoo suggested.

src/util/pg.py
```diff
@@ -361,9 +361,19 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
             [model_of_table(cr, table), column],
         )
     if not using:
-        # Simple case. Use general SQL syntax
-        cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" TYPE %s' % (table, column, type))
-        return
+        # if there is a high number of NULL entries, it will be faster to just ignore those
+        cr.execute(sql.SQL("ANALYZE {}({})").format(sql.Identifier(table), sql.Identifier(column)))
```
aj-fuentes, line 365: Ideally we could wait to merge `util.format` and reuse it, since it's less verbose.

src/util/pg.py
cawo-odoo: > That's why I only `analyze`d this column.

I noticed that, my thought was just that in the vast majority of cases[^1] it should be superfluous to call it at all, such that whatever the cost, it is wasted. Of course you can argue that the cost is so small compared to the actual operation's, that we don't care. You decide.

[^1]: I know that internal disables the full ANALYZE sometimes, but in the platform, all columns that exist in the original DB should already have stats. And of course now that this repo is public, there are also users where we don't know about that.

nseinlet: Internal don' t analyse because stats are accurate at the beginning of the upgrade. But it' s a specific use case

src/util/pg.py
```diff
@@ -407,9 +407,19 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
...
+            "SELECT null_frac FROM pg_stats WHERE schemaname = current_schema() AND tablename = %s AND attname = %s",
+            [table, column],
+        )
+        [null_frac] = cr.fetchone() or (0.0,)
+        if null_frac > 0.70:
+            using = "{{0}}::{}".format(type)
+        else:
+            # Simple case. Use general SQL syntax
+            cr.execute(format_query(cr, "ALTER TABLE {} ALTER COLUMN {} TYPE {}", table, column, sql.SQL(type)))
+            return
```
aj-fuentes, line 422:
```suggestion
        if null_frac < 0.70:
            # Simple case. Use general SQL syntax
            cr.execute(format_query(cr, "ALTER TABLE {} ALTER COLUMN {} TYPE {}", table, column, sql.SQL(type)))
            return
        using = "{{0}}::{}".format(type)
```

aj-fuentes (approved):

KangOl:
@robodoo rebase-ff r+

PR #28, opened 2023-12-21
[IMP] util.create_column: create FK

KangOl:
Allow to directly set the FK constraint when creating a new column.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -29,6 +29,8 @@ from .misc import log_progress
 
 _logger = logging.getLogger(__name__)
 
+ON_DELETE_ACTIONS = frozenset(("SET NULL", "CASCADE", "RESTRICT", "NO ACTION", "SET DEFAULT"))
+
 
 class PGRegexp(str):
     """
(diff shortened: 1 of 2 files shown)


src/util/pg.py
```diff
@@ -290,8 +290,23 @@ def create_column(cr, table, column, definition, **kwargs):
...
+    fk = kwargs.pop("fk", "")
     if kwargs:
         raise TypeError("create_column() got an unexpected keyword argument %r" % kwargs.popitem()[0])
+
+    if fk:
+        if isinstance(fk, str):
+            fk = (fk, "id", "SET NULL")
+        if not (isinstance(fk, tuple) and len(fk) == 3 and all(isinstance(e, str) for e in fk)):
+            raise ValueError("`fk` is expected to be a table name or a 3-tuple (table_name, column, on_delete_action)")
+        if fk[2].upper() not in ("SET NULL", "CASCADE", "RESTRICT"):
```
nseinlet, line 302: Extract those values in a constant, to be consistent with " create_fk" definition below.

src/util/pg.py
```diff
@@ -318,7 +333,7 @@ def create_column(cr, table, column, definition, **kwargs):
             parallel_execute(cr, explode_query_range(cr, query, table=table))
```
nseinlet, line 333: You should probably call "create_fk" if needed.

nseinlet (approved):

nseinlet:
@robodoo r+ rebase-ff

PR #29, opened 2023-12-21
[IMP] util/fields: improve rex to be more precise

diagnoza:
So far we've been matching very broadly, now we only look for actual references of the field, eg `.data` or `"data"`.

[opw-3637908](https://www.odoo.com/web#id=3637908&cids=1&model=project.task&view_type=form)


src/util/fields.py
```diff
@@ -833,6 +833,7 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
 
     p = {
         "old": r"\y%s\y" % (re.escape(old),),
+        "old_pattern": r"\.\y%s\y|\"\y%s\y\"|\'\y%s\y\'" % (re.escape(old), re.escape(old), re.escape(old)),
```
aj-fuentes, line 836: I think it could be simpler.
```suggestion
        "old_pattern": r"[\."']%s\y" % (re.escape(old), re.escape(old), re.escape(old)),
```

diagnoza: What about dict keys "<pattern> something something"?

KangOl:

src/util/fields.py
aj-fuentes: > What about dict keys ">pattern< something something"?

`r"([\."'])%s\1"` ?

aj-fuentes: actually only for the ' or " but it won't be more readable than what you already have so...

aj-fuentes: Still the `\y` are not needed, so I'd recommend to put the cases explicitly:
`r"""\.%s\y|"%s"|'%s'"""`
In principle you could pass a mapping to `%` to avoid escaping the same thing three times.

diagnoza: I've changed it to something more appealing

src/util/fields.py
```diff
@@ -833,6 +833,7 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
 
     p = {
         "old": r"\y%s\y" % (re.escape(old),),
+        "old_pattern": r"""\.{0}\y|"{0}"|'{0}'""".format(re.escape(old)),
```
KangOl, line 836: Matching the ending quote will ignore patterns like `.filtered("foo_id.bar")`.
```suggestion
        "old_pattern": r"""[.'"]{0}\y""".format(re.escape(old)),
```

aj-fuentes: Same I proposed [here](https://github.com/odoo/upgrade-util/pull/29#discussion_r1434221635)

I agree it would be enough...

src/util/fields.py
```diff
@@ -833,6 +833,7 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
 
     p = {
         "old": r"\y%s\y" % (re.escape(old),),
+        "old_pattern": r"""[\.'"]{0}\y""".format(re.escape(old)),
```
KangOl, line 836: No need to escape `.` in `[]`.
```suggestion
        "old_pattern": r"""[.'"]{0}\y""".format(re.escape(old)),
```

KangOl:
```
[local] chs@t=# select 'blah.foo_id[0]' ~ '[.''"]foo_id\y';
┌──────────┐
│ ?column? │
├──────────┤
│ t        │
└──────────┘
(1 row)

Time: 4.612 ms
```

aj-fuentes (approved):

KangOl:
@robodoo r+

PR #31, opened 2024-01-02
[IMP] skip useless queries

nseinlet:
When renaming models, indirect references should be uypdated In some cases, there's no record to update, but system spend time to execute the useless queries.


KangOl:
upgradeci retry with always only hr_contract_salary hr_holidays industry_fsm l10n_pl l10n_pl_jpk l10n_pl_reports mail mass_mailing payment point_of_sale stock timesheet_grid voip

KangOl:
@robodoo r+

PR #32, opened 2024-01-02
[chore] format code using ruff instead of black 

KangOl:


aj-fuentes (approved):
LGTM, the comment could be ignored.

KangOl:
Question:
Do we keep the `[tools.black]` section in the `pyproject.toml` file?
It is used by the `tools/generate-inherit.py` script.
Options:
 - keep it
 - change the script to either use the value from `[tools.ruff]` or hardcode the line-length.
 - We can also not format the output of the script ourself and let `ruff` doing it via `pre-commit`.

aj-fuentes:
> We can also not format the output of the script ourself and let ruff doing it via pre-commit.

I think that's the best option. As you said, regenerate the script then let pre-commit reformat it. I'd prefer to decouple the generating script from the linting.

KangOl:

KangOl:
@robodoo rebase-ff r+

PR #33, opened 2024-01-03
[FIX] util/records: ensure no duplicates for m2m when replacing refs

KangOl:
There are two ways an update of references could fail in an m2m table. If there is a record already in the table matching the currently updated row. Example:
```
 mapping old new
          1   3
          2   3
 Fails for:
 m2m fk col2
     1   x  -updated-> 3  x !error
     3   x
```
Or, if we have multiple ids mapped to the same new one. Example:
```
 mapping old new
          1   3
          2   3
 Fails for:
 m2m fk col2
     1   x  -updated-> 3  x
     2   x  -updated-> 3  x !error
```


KangOl:
upgradeci retry with always only helpdesk

src/util/records.py
```diff
@@ -1047,13 +1047,27 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
                 cols = get_columns(cr, table, ignore=(fk,))
```
KangOl, line 1047: Use the new toys.
```suggestion
                cols = get_columns(cr, table, ignore=(fk,)).iter_unquoted()
```

Then, the `get_fk()` call above should pass the `quote_ident=False` argument, and we should use `format_query()` instead of `query.format()`

KangOl:
Wouldn't it be cleaner with multiple calls to `cr.execute` instead of one call with multiple semicolon-separated queries?

aj-fuentes:
> Wouldn't it be cleaner with multiple calls to `cr.execute` instead of one call with multiple semicolon-separated queries?

I totally agree. I was actually thinking about that, I'll change it.

KangOl:
matt is red for a good reason: some commit is done at some point.

KangOl:
@robodoo rebase-ff r+

PR #34, opened 2024-01-08
[FIX] util/fields: fix cascade ttype update afu

aj-fuentes:
### [FIX] util/fields: fix missing function parameter

`convert_field_to_html` expects a cursor as firts parameter

### [FIX] util/fields: also convert inherits fields

When converting fields types we need to continue to inherits. Before we
avoided that because the column may be missing but it leads to situations
where custom manual fields are not updated properly.

For example if we add a custom manual field in `product.packaging.x_pdesc`
related to `product_id.description`. Without this patch the upgrade `x_pdesc`
is still of `ttype=text`. With this patch:
```
test_14_15=> select id,model,name,state,related,related_field_id,ttype from ir_model_fields where name like 'x\_%' or (model like 'product.%' and name='description')
+------+-------------------+-------------+--------+-----------------------------+------------------+-------+
| id   | model             | name        | state  | related                     | related_field_id | ttype |
|------+-------------------+-------------+--------+-----------------------------+------------------+-------|
| 3068 | product.packaging | x_pdesc     | manual | product_id.description      | 2635             | html  |
| 2516 | product.template  | description | base   | <null>                      | <null>           | html  |
| 2635 | product.product   | description | base   | product_tmpl_id.description | <null>           | html  |
+------+-------------------+-------------+--------+-----------------------------+------------------+-------+
```

There are many specific fixes to update manual related fields to
`product.product.description`.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -488,23 +488,24 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
 def convert_field_to_html(cr, model, field, skip_inherit=()):
     _validate_model(model)
     table = table_of_model(cr, model)
-    jsonb_column = column_type(cr, table, field) == "jsonb"
-    if jsonb_column:
-        query = """
-            WITH html_values AS (
-                SELECT id, jsonb_object_agg(t.key, {2}) AS value
-                  FROM "{0}", jsonb_each_text("{1}") AS t
-                 WHERE {{parallel_filter}}
-              GROUP BY id
-            )
-            UPDATE "{0}"
-               SET "{1}" = h.value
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always account approvals base calendar crm event fleet gamification helpdesk hr hr_appraisal hr_contract hr_holidays hr_payroll hr_recruitment hr_referral l10n_be_hr_contract_salary l10n_multilang lunch maintenance mrp mrp_plm product purchase purchase_requisition repair sale sale_management sale_subscription sale_temporal stock website_sale_stock website_slides website_twitter_wall

KangOl:
@robodoo r+

PR #35, opened 2024-01-09
[IMP] tests: avoid duplicate tests

Xavier-Do:
Linked to odoo/odoo#148660

In the corresponding community pr, a test is not executed twice when importing a testClass, a common mistake when writing tests.

Upgrade tests were relying on this side effect to run tests. A flag is added on the class to enable this behavior in this case only.


KangOl:
@robodoo r+

PR #38, opened 2024-01-16
[IMP] util/pg: allow multiple parallel filters

aj-fuentes:
When exploding a query we may need to add the same parallel filter in multiple places.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -212,10 +212,11 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
         parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
         return [query.format(parallel_filter=parallel_filter)]
 
-    parallel_filter = "{alias}.id BETWEEN %s AND %s".format(alias=alias)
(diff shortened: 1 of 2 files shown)


src/util/pg.py
```diff
@@ -212,10 +212,11 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
         parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
         return [query.format(parallel_filter=parallel_filter)]
 
-    parallel_filter = "{alias}.id BETWEEN %s AND %s".format(alias=alias)
+    parallel_filter = "{alias}.id BETWEEN %(low)s AND %(high)s".format(alias=alias)
```
KangOl, line 215: Nitpicking on variable name:
```suggestion
    parallel_filter = "{alias}.id BETWEEN %(lower-bound)s AND %(upper-bound)s".format(alias=alias)
```

aj-fuentes: Done

KangOl:

KangOl:
@robodoo r+

PR #39, opened 2024-01-25
[FIX] util/records: correct query construction

aj-fuentes:
Check that there are unique columns to expand the extra conditions.


KangOl:
What about simply
```diff
-                    % {"ands": "AND".join('"%s"=t."%s"' % (col, col) for col in uniq_cols)}
+                    % {"ands": "AND".join('"%s"=t."%s"' % (col, col) for col in uniq_cols) or "true"}

aj-fuentes:
Checking the other error...

aj-fuentes:
@KangOl this looks good, only error is in pos_blackbox

KangOl:
@robodoo r+

PR #43, opened 2024-02-13
[FIX] util/fields : add force_adapt in adapt_depends

thni-odoo:
Adding `force_adapt` to consider whole path instead of just the leaf part
same as related fields.


KangOl:
upgradeci skip
@robodoo r+

PR #44, opened 2024-02-21
[ADD] util/specific.py: add helper to rename module

sagu-odoo:
Whenever we changes the custom module name every time we have to add migration report statically for reducing line of code [reference](https://github.com/odoo/upgrade-util/commit/7786de7d7f4e8333553b3d62b2451fe43d5f838a).

### DIFF
--- a/src/util/specific.py
+++ b/src/util/specific.py
@@ -3,6 +3,7 @@ import logging
 
 from .helpers import _validate_table
 from .misc import _cached
+from .modules import rename_module
 from .pg import column_exists, rename_table
 from .report import add_to_migration_reports
 
@@ -40,6 +41,17 @@ def dispatch_by_dbuuid(cr, version, callbacks):
         func = callbacks[uuid]
         _logger.info("calling dbuuid-specific function `%s`", func.__name__)
         func(cr, version)
+
+
+def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
+    rename_module(cr, old_module_name, new_module_name)
+
+    add_to_migration_reports(
+        category="Custom modules",
+        message="The custom module '{old_module_name}' was renamed to '{new_module_name}'. {report_details}".format(
+            **locals()
+        ),
+    )
 
 
 def rename_custom_table(


KangOl:
upgradeci skip

src/util/specific.py
```diff
@@ -60,6 +61,17 @@ def rename_custom_table(
...
 
 
+def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
+    rename_module(cr, old_module_name, new_module_name)
+
+    add_to_migration_reports(
+        category="Rename Module",
+        message="The custom module '{old_module_name}' was renamed to '{new_module_name}'. {report_details}".format(
+            **locals()
+        ),
```
Pirols, line 71:
```suggestion
        message="The custom module %r was renamed to %r. %s" % (old_module_name, new_module_name, report_details)
```

Using `locals()` seems an overkill here.

src/util/specific.py
```diff
@@ -60,6 +61,17 @@ def rename_custom_table(
     )
 
 
+def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
+    rename_module(cr, old_module_name, new_module_name)
+
+    add_to_migration_reports(
+        category="Rename Module",
+        message="The custom module '{old_module_name}' was renamed to '{new_module_name}'. {report_details}".format(
```
asno-odoo, line 69: wouldn't it be better to add the version the module was renamed in.

src/util/specific.py
```diff
@@ -60,6 +61,17 @@ def rename_custom_table(
...
+            **locals()
+        ),
+    )
```
Pirols, line 72:
```suggestion
def rename_custom_module(cr, old_module_name, new_module_name, report_details="", skip_check=False):
    if skip_check or module_installed(cr, old_module_name):
        rename_module(cr, old_module_name, new_module_name)

        add_to_migration_reports(
            category="Rename Module",
            message="The custom module '{old_module_name}' was renamed to '{new_module_name}'. {report_details}".format(
                **locals()
            ),
        )
```

Avoid writing to migration report if the module is not found in the current db.

src/util/specific.py
```diff
@@ -60,6 +61,17 @@ def rename_custom_table(
     )
 
 
+def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
```
Pirols, line 64:
```suggestion
def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
    """DOCSTRING"""
```

In principle, AFAI understood when we touch (or - like in this case - create) a util function, we should add/correct a DOCSTRING. So that, at some point, we will be able to automatically generate documentation. 

OTOH, I'm not sure if we have established a format yet, have we? @aj-fuentes @KangOl

Pirols (commented):
My 2c

src/util/specific.py
Pirols, line 64: I would also move this utility definition to before `rename_custom_table()` to list them sorted by magnitude (`module` > `model` ~~ `table` > `column`).

src/util/specific.py
aj-fuentes: The module must be renamed even if not installed in many cases.

src/util/specific.py
aj-fuentes: This function is simple enough.

aj-fuentes (approved):
LGTM

src/util/specific.py
sagu-odoo: changes done as per suggestion thanks for the review

aj-fuentes:
upgradeci skip

src/util/specific.py
```diff
@@ -42,6 +43,17 @@ def dispatch_by_dbuuid(cr, version, callbacks):
         func(cr, version)
 
 
+def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
+    rename_module(cr, old_module_name, new_module_name)
+
+    add_to_migration_reports(
+        category="Rename Module",
```
aj-fuentes, line 50: To be in line with the other categories for custom tables/columns
```suggestion
        category="Custom modules",
```

aj-fuentes:
upgradeci skip

aj-fuentes:
@robodoo r+

PR #46, opened 2024-02-22
[FIX] util.replace_record_references_batch: handle unique constraints

KangOl:
We already do it for indirect references (res_model/res_id), why not for the direct ones?

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1061,6 +1061,19 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
                   FROM _upgrade_rrr r
                  WHERE r.old = t.{fk}
             """
+            unique_indexes = _get_unique_indexes_with(cr, table, fk)
+            if unique_indexes:
+                conditions = [""]
+                for _, uniq_cols in unique_indexes:
+                    uniq_cols = set(uniq_cols) - {fk}  # noqa: PLW2901
+                    ands = (
+                        " AND ".join(format_query(cr, "u.{col} = t.{col}", col=col) for col in uniq_cols)
+                        if uniq_cols
+                        else "true"
+                    )
+                    conditions.append("NOT EXISTS(SELECT 1 FROM {table} u WHERE u.{fk} = r.new AND %s)" % ands)
+
+                query += " AND ".join(conditions)
 
             if not column_exists(cr, table, "id"):
                 # seems to be a m2m table. Avoid duplicated entries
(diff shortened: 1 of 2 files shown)


src/util/records.py
```diff
@@ -1061,6 +1061,19 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
                   FROM _upgrade_rrr r
                  WHERE r.old = t.{fk}
             """
+            unique_indexes = _get_unique_indexes_with(cr, table, fk)
+            if unique_indexes:
+                conditions = [""]
+                for _, uniq_cols in unique_indexes:
+                    uniq_cols = set(uniq_cols) - {fk}  # noqa: PLW2901
```
aj-fuentes, line 1068: What if the unique constraint is _only_ for FK? Do we care?

aj-fuentes (approved):
Generally speaking it LGTM. Only one small question.

KangOl:
```
upgradeci retry with always only account account_edi_ubl_cii analytic base calendar crm crm_iap_lead documents event helpdesk hr hr_appraisal hr_appraisal_survey hr_payroll hr_timesheet knowledge l10n_be l10n_bo l10n_cn l10n_it l10n_nl l10n_pa l10n_pl l10n_ro loyalty payment pos_blackbox_be project project_todo rating_project sale_subscription sign stock website_forum website_slides in all versions
```

src/util/records.py
KangOl: It's handled by the `or "true"` part a few lines below.

aj-fuentes: I meant, in the case the fk had itself a uniq column, this breaks if:
1. There was another entry for the same target value
2. If we replace two distinct ids to the same target.

```
replace 1->4 2->5 3->5
table fk (unique)
      1   # fails update for case 1: collision with 4   
      2   
      3   # fails update for case 2: 2 and 3 both map to 5
      4
```

I'm still not sure we can handle this correctly (not just for the fk uniq). We are dropping updates that could conflict due to unique constraints but how could we know which updates are safe to drop? Maybe it's better to let this fail and be handled specifically.

KangOl: The patch is actually there to fix the `1 -> 4` case.
The other case cannot be easily detected. This however doesn't change the current behavior.

aj-fuentes: Yes, I see it after you told me, albeit it's a bit hard to follow it is indeed there.

Still we have the issue about dropping an update that previously would just error out. Would it cause issues?

A common flow is to replace references then remove the source records, which sets null in most cases (for FKs). We don't know if that null could cause issues. Before this patch the update would just fail. With this patch setting null doesn't fail (at least we expect so), and we let the upgrade continue. Post upgrade the null value may be incorrect. This could unlock hidden issues.

In other words we would be breaking the expectation that replacing all references actually replaces _all_ references.

KangOl:
@robodoo r+

PR #49, opened 2024-02-28
[IMP] util/pg: serialize queries upon DeadlockDetected/SerializationFailure

Pirols:
OPW-3875155

As recently noted in https://github.com/odoo/upgrade/pull/5753, it is possible that `parallel_execute` generated concurrency issues. For instance, if a `DELETE` query is run in `//` on a table with a self-referencing foreign key, based on the actual records in the db, the following error may be encountered:

```
  File "/home/odoo/src/odoo/15.0/odoo/service/server.py", line 1260, in preload_registries
  File "/home/odoo/src/odoo/15.0/odoo/modules/registry.py", line 87, in new
  File "/home/odoo/src/odoo/15.0/odoo/modules/loading.py", line 415, in load_modules
  File "/home/odoo/src/odoo/15.0/odoo/modules/loading.py", line 174, in load_module_graph
  File "/home/odoo/src/odoo/15.0/odoo/modules/migration.py", line 186, in migrate_module
  File "/tmp/tmpjgh5os94/migrations/base/0.0.0/pre-clean-assets.py", line 7, in migrate
    util.parallel_execute(
  File "/tmp/tmpjgh5os94/migrations/util/pg.py", line 103, in parallel_execute
    return sum(
  File "/tmp/tmpjgh5os94/migrations/util/misc.py", line 215, in log_progress
    for i, e in enumerate(it, 1):
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
    yield fs.pop().result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/tmpjgh5os94/migrations/util/pg.py", line 95, in execute
    cr.execute(query)
  File "<decorator-gen-5>", line 2, in execute
  File "/home/odoo/src/odoo/15.0/odoo/sql_db.py", line 90, in check
  File "/home/odoo/src/odoo/15.0/odoo/sql_db.py", line 311, in execute
psycopg2.errors.DeadlockDetected: deadlock detected
DETAIL:  Process 1418247 waits for ShareLock on transaction 21800495; blocked by process 1418237.
Process 1418237 waits for ShareLock on transaction 21800496; blocked by process 1418247.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (450,6) in relation "ir_attachment"
SQL statement "UPDATE ONLY "public"."ir_attachment" SET "doc_next_name_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "doc_next_name_id""
```

So far the solution has always been to deparallelize the query, maybe exploding it and running the sub queries sequentially.

This commit implements a mechanism to automatically collect queries that fail due to concurrency issues and rerun them sequentially with no further failure control.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -14,7 +14,8 @@ from functools import partial, reduce
 from multiprocessing import cpu_count
 
 try:
-    from concurrent.futures import ThreadPoolExecutor
+    from concurrent.futures import ThreadPoolExecutor  # noqa: I001
+    import concurrent
 except ImportError:
     ThreadPoolExecutor = None
 
@@ -29,7 +30,7 @@ except NameError:
     pass
 
 import psycopg2
-from psycopg2 import sql
+from psycopg2 import errorcodes, sql
 
 try:
     from odoo.sql_db import db_connect
@@ -117,17 +118,34 @@ if ThreadPoolExecutor is not None:
 
         cr.commit()
 
+        CONCURRENCY_ERRORCODES = {errorcodes.DEADLOCK_DETECTED}
+        failed_queries = []
+        tot_cnt = 0
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
-            return sum(
-                log_progress(
-                    executor.map(execute, queries),
-                    logger,
-                    qualifier="queries",
-                    size=len(queries),
-                    estimate=False,
-                    log_hundred_percent=True,
-                )
-            )
+            future_queries = {executor.submit(execute, q): q for q in queries}
+            for future in log_progress(
+                concurrent.futures.as_completed(future_queries),
+                logger,
+                qualifier="queries",
+                size=len(queries),
+                estimate=False,
+                log_hundred_percent=True,
+            ):
+                try:
+                    tot_cnt += future.result()
+                except psycopg2.OperationalError as exc:
+                    if exc.pgcode not in CONCURRENCY_ERRORCODES:
+                        raise
+
+                    # to be retried without concurrency
+                    failed_queries.append(future_queries[future])
+
+        if failed_queries:
+            logger.warning("Serialize queries that failed due to concurrency issues")
+            tot_cnt += _parallel_execute_serial(cr, failed_queries, logger=logger)
+            cr.commit()
+
+        return tot_cnt
 
 else:
     _parallel_execute_threaded = _parallel_execute_serial
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -118,7 +118,7 @@ if ThreadPoolExecutor is not None:
                 log_hundred_percent=True,
             ):
                 try:
-                    tot_cnt += future.result()
+                    tot_cnt += future.result() or 0
                 except psycopg2.OperationalError as exc:
                     if exc.pgcode not in CONCURRENCY_ERRORCODES:
                         raise
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -84,20 +84,6 @@ def _parallel_execute_serial(cr, queries, logger=_logger):
 if ThreadPoolExecutor is not None:
 
     def _parallel_execute_threaded(cr, queries, logger=_logger):
-        """
-        Execute queries in parallel.
-
-        Use a maximum of 8 workers (but not more than the number of CPUs)
-        Side effect: the given cursor is commited.
-        As example, on `**REDACTED**` (using 8 workers), the following gains are:
-            +---------------------------------------------+-------------+-------------+
-            | File                                        | Sequential  | Parallel    |
-            +---------------------------------------------+-------------+-------------+
-            | base/saas~12.5.1.3/pre-20-models.py         | ~8 minutes  | ~2 minutes  |
-            | mail/saas~12.5.1.0/pre-migrate.py           | ~10 minutes | ~4 minutes  |
-            | mass_mailing/saas~12.5.2.0/pre-10-models.py | ~40 minutes | ~18 minutes |
-            +---------------------------------------------+-------------+-------------+
-        """
         if not queries:
             return None
 
@@ -152,6 +138,32 @@ else:
 
 
 def parallel_execute(cr, queries, logger=_logger):
+    """
+    Execute queries in parallel.
+
+    .. example::
+       .. code-block:: python
+
+          util.parallel_execute(cr, [util.format_query(cr, "REINDEX TABLE {}", t) for t in tables])
+
+    .. tip::
+       If looking to speedup a single query, see :func:`~odoo.upgrade.util.pg.explode_execute`.
+
+    :param list(str) queries: list of queries to execute concurrently
+    :param `~logging.Logger` logger: logger used to report the progress
+    :return: the sum of `cr.rowcount` for each query run
+    :rtype: int
+
+    .. warning::
+       - Due to the nature of `cr.rowcount`, the return value of this function may represent an
+         underestimate of the real number of affected records. For instance, when some records
+         are deleted/updated as a result of an `ondelete` clause, they won't be taken into account.
+
+       - As a side effect, the cursor will be committed.
+
+    .. note::
+       If a concurrency issue occurs, the *failing* queries will be retried sequentially.
+    """
     parallel_execute_impl = (
         _parallel_execute_serial
         if getattr(threading.current_thread(), "testing", False)
@@ -330,6 +342,7 @@ def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_log
        It's up to the caller to ensure the queries do not update the same records in
        different buckets. It is advised to never use this function for `DELETE` queries on
        tables with self references due to the potential `ON DELETE` effects.
+       For more details see :func:`~odoo.upgrade.util.pg.parallel_execute`.
     """
     return parallel_execute(
         cr,
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -104,7 +104,10 @@ if ThreadPoolExecutor is not None:
 
         cr.commit()
 
-        CONCURRENCY_ERRORCODES = {errorcodes.DEADLOCK_DETECTED}
+        CONCURRENCY_ERRORCODES = {
+            errorcodes.DEADLOCK_DETECTED,
+            errorcodes.SERIALIZATION_FAILURE,
+        }
         failed_queries = []
         tot_cnt = 0
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -525,6 +525,50 @@ class TestPG(UnitTestCase):
         self.test_parallel_rowcount()
         threading.current_thread().testing = True
 
+    def test_parallel_execute_retry_on_serialization_failure(self):
+        TEST_TABLE_NAME = "_upgrade_serialization_failure_test_table"
+        N_ROWS = 10
+
+        cr = self.env.cr
+
+        cr.execute(
+            util.format_query(
+                cr,
+                """
+                DROP TABLE IF EXISTS {table};
+
+                CREATE TABLE {table} (
+                    id SERIAL PRIMARY KEY,
+                    other_id INTEGER,
+                    FOREIGN KEY (other_id) REFERENCES {table} ON DELETE CASCADE
+                );
+
+                INSERT INTO {table} SELECT GENERATE_SERIES(1, %s);
+
+                -- map odd numbers `n` to `n + 1` and viceversa (`n + 1` to `n`)
+                UPDATE {table} SET other_id = id + (MOD(id, 2) - 0.5)*2;
+                """
+                % N_ROWS,
+                table=TEST_TABLE_NAME,
+            )
+        )
+
+        threading.current_thread().testing = False
+        # exploded queries will generate a SerializationFailed error, causing some of the queries to be retried
+        util.explode_execute(
+            cr, util.format_query(cr, "DELETE FROM {}", TEST_TABLE_NAME), TEST_TABLE_NAME, bucket_size=1
+        )
+        threading.current_thread().testing = True
+
+        if hasattr(self, "_savepoint_id"):
+            # `explode_execute` causes the cursor to be committed, losing the automatic checkpoint
+            # Force a new one to avoid issues when cleaning up
+            self.addCleanup(cr.execute, f"SAVEPOINT test_{self._savepoint_id}")
+            self.addCleanup(cr.execute, util.format_query(cr, "DROP TABLE IF EXISTS {}", TEST_TABLE_NAME))
+
+        cr.execute(util.format_query(cr, "SELECT 1 FROM {}", TEST_TABLE_NAME))
+        self.assertFalse(cr.rowcount)
+
     def test_create_column_with_fk(self):
         cr = self.env.cr
         self.assertFalse(util.column_exists(cr, "res_partner", "_test_lang_id"))


Pirols:
The original idea (courtesy of @cawo-odoo) was to retry the queries a number of times (as done in https://github.com/odoo/odoo/blob/3bd399e2ca9d03364965aa7f252cfdd51ad4623d/odoo/service/model.py#L117).

src/util/pg.py
```diff
@@ -97,27 +97,55 @@ def _parallel_execute_threaded(cr, queries, logger=_logger):
...
-                return cnt
+                try:
+                    cr.execute(query)
+                    cnt = cr.rowcount
+                    cr.commit()
+                except psycopg2.OperationalError as exc:
+                    if exc.pgcode not in PG_CONCURRENCY_ERRORS_TO_RETRY:
+                        raise
+                    cr.rollback()
+                    return query
```
aj-fuentes, line 120: I don't like this mixed returns. I'd just define a `conflict_queries = []` outside `execute` and add queries there when they fail. The use that  `_parallel_execute_serial(cr, conflict_queries, ...)`

Pirols: > That idea doesn't sound thread safe.

I think it is python-thread safe (due to the GIL). Still, it's probably safer not to do that.

aj-fuentes: You can use a synchronized queue, but well maybe not worth the hassle.

Pirols: Yeah, I'm not sure it is. I will do a fixup commit, then we choose.

cawo-odoo: > I think it is python-thread safe (due to the GIL)

That's an implementation detail I would not make assumptions or rely on. You could probably get away with a bit of manual locking.

```python
from threading import Lock

def execute():
    ...
    with lock:
        conflict_queries.append(query)

conflict_queries = []
lock = Lock()

with ThreadPoolExecutor(max_workers=max_workers) as executor:
....
```

src/util/pg.py
```diff
@@ -97,27 +97,55 @@ def _parallel_execute_threaded(cr, queries, logger=_logger):
...
+                log_hundred_percent=True,
             )
 
+        total_affected_rows = 0
+        with cursor() as cr:
+            for query_or_cnt in out:
+                if isinstance(query_or_cnt, int):
+                    total_affected_rows += query_or_cnt
+                else:
+                    # queries that failed to run in //, try running them sequentially
```
cawo-odoo, line 142: Maybe log an info message here?

src/util/pg.py
Pirols: Update: I think that the synchronized queue makes for a nice solution here.

src/util/pg.py
Pirols: Yes, that's probably helpful.

Pirols: Current output:

```
...
2024-02-28 15:24:26,609 28 INFO pied_1192107_15.0 odoo.modules.migration: module base: Running migration [>0.0.0] pre-clean-assets
2024-02-28 15:24:27,642 28 ERROR ? odoo.sql_db: bad query:
            DELETE FROM ir_attachment
                  WHERE res_model = 'ir.ui.view'
                    AND COALESCE(res_id, 0) = 0
                    AND res_field IS NULL
                    AND mimetype IN ('text/css', 'application/javascript')
                    AND url ~ '^/web/(content|assets)/\d+-[0-9a-f]+/.*\.(css|js)$'
             AND ir_attachment.id BETWEEN 10001 AND 20000
ERROR: deadlock detected
DETAIL:  Process 2628328 waits for ShareLock on transaction 49515341; blocked by process 2628321.
Process 2628321 waits for ShareLock on transaction 49515339; blocked by process 2628328.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (456,13) in relation "ir_attachment"
SQL statement "UPDATE ONLY "public"."ir_attachment" SET "doc_next_name_id" = NULL WHERE $1 OPERATOR(pg_catalog.=) "doc_next_name_id""

2024-02-28 15:24:27,727 28 INFO pied_1192107_15.0 odoo.upgrade.util.pg: Execute sequentially queries that failed due to concurrency issues
...
```

src/util/pg.py
```diff
@@ -97,17 +102,36 @@ def _parallel_execute_threaded(cr, queries, logger=_logger):
...
+                try:
+                    cr.execute(query)
+                    cnt = cr.rowcount
+                    cr.commit()
+                except psycopg2.OperationalError as exc:
+                    if exc.pgcode not in PG_CONCURRENCY_ERRORS_TO_RETRY:
+                        raise
+                    cr.rollback()
+                    failed_queries.put_nowait(query)
+                    return 0
```
aj-fuentes, line 127:
```suggestion
                    return -1  # error
```

src/util/pg.py
```diff
@@ -97,17 +102,36 @@ def _parallel_execute_threaded(cr, queries, logger=_logger):
...
+                else:
+                    return cnt
 
         cr.commit()
 
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
-            return sum(
+            tot_cnt = sum(
```
aj-fuentes, line 134:
```suggestion
            res = list(
```

src/util/pg.py
```diff
@@ -118,6 +142,13 @@ def execute(query):
                 )
             )
 
+        if not failed_queries.empty():
+            logger.info("Execute sequentially queries that failed due to concurrency issues")
+            tot_cnt += _parallel_execute_serial(cr, list(failed_queries.queue), logger=logger)
+            cr.commit()
```
aj-fuentes, line 148:
```suggestion
        tot_count = sum(q_count for q_count in res if q_count >= 0)
        failed_queries = [q for q, q_res in zip(queries, res) if q_res == -1]
        if failed_queries:
            logger.info("Execute sequentially queries that failed due to concurrency issues")
            tot_cnt += _parallel_execute_serial(cr, list(failed_queries.queue), logger=logger)
            cr.commit()
```

aj-fuentes (commented):
After seeing the alternatives, I think you original solution is simpler. I'd still recommend returning an error code instead of different types -- str vs int. `-1` could be used as error for the count per query.

Pirols:
> After seeing the alternatives, I think you original solution is simpler. I'd still recommend returning an error code instead of different types -- str vs int. `-1` could be used as error for the count per query.

This was actually my first idea, it didn't make the cut for me because `-1` is a potential outcome of rowcount[^1], for instance on `CREATE INDEX` queries ([example](https://github.com/odoo/upgrade/blob/0e8e41f4ecb922b33e8b34f683aff314bb276a12/migrations/base/0.0.0/pre-10-create-idx-fk.py#L497)).

Now, can we use `-2`? AFAICT: yes. OTOH IMO it seems an arbitrary and questionable choice that would at least require a comment to clarify. On top of that, as you suggested, we would need to zip queries to their output, implying the assumption that the output from `ThreadPoolExecutor` would be sorted in the right order, which would be correct but might add an extra bit of cognitive load.

All in all, I find the current implementation that leverages a synchronized queue to be better: it uses the proper tool for the job which is self-documenting in the intent and simplifies the code substantially.

[^1]: https://www.psycopg.org/docs/cursor.html#cursor.rowcount - I was considering publishing a follow up PR adding a check for `None` values (e.g. `sum(x for x in res if x is not None)`, since the docs say that the output might change at some point.

KangOl:
I would use the same idiom as in `matt.py` and return a `Result` object[^1][^2]


[^1]: https://github.com/odoo/upgrade/blob/5b82305461e1f8821ed5474be0a9c60dcadceb7e/tools/matt.py#L35-L62
[^2]: minus the typing, to be python2 compatible

aj-fuentes:
> Now, can we use -2? AFAICT: yes. OTOH IMO it seems an arbitrary and questionable choice that would at least require a comment to clarify.

You could use any other value `-100`? Note that the values are clearly documented here, it's either positive, `-1`, or a "future" None. To me it looks OK to use any other value.

> On top of that, as you suggested, we would need to zip queries to their output, implying the assumption that the output from ThreadPoolExecutor would be sorted in the right order, which would be correct but might add an extra bit of cognitive load.

There is no assumption here. [`ThreadPoolExecutor.map`](https://docs.python.org/3.8/library/concurrent.futures.html#concurrent.futures.Executor.map) must work like that, otherwise even their [own examples](https://docs.python.org/3.10/library/concurrent.futures.html#processpoolexecutor-example) are wrong. To put in another way, how would a `map` implementation be correct if it doesn't respect the order? It won't be a map otherwise.

> it uses the proper tool for the job which is self-documenting in the intent and simplifies the code substantially.

It adds extra locks and to some extent potential deadlocks that we actually don't know about. In other words, it is not clearly documented whether the executor code can safely interact with data not handled by the `Executor` class -- I think it's safe, especially if using the standard synchronized queues, but that's the main argument I have for the change. Note that this implementation would not work for a hypothetical `ProcessPoolExecutor` case. 

IMO your original implementation was better aligned with the idea of using `map` as a black box to which we pass a function and get back only its result. My only extra suggestion is to get rid of mixed return objects.

All that being said, current implementation seems correct also. Maybe @KangOl can fully check this.

aj-fuentes:
Nit: deparallelize -> serialize ?

aj-fuentes (approved):

Pirols:
I pushed an alternative version we worked on with @cawo-odoo.
It's not 100% polished yet.

I (we) appreciate the fact that the deadlock is caught and dealt with by the main thread, requiring no special behaviour from the workers callbacks. Opinions? @aj-fuentes @KangOl

KangOl:
Can you try to add a test to prove it works as expected?

aj-fuentes:
To simplify review: you implemented this example https://docs.python.org/3.10/library/concurrent.futures.html#threadpoolexecutor-example

It looks correct. I also like that it avoids the mixed returns in the right way.

src/util/pg.py
```diff
@@ -70,13 +72,13 @@ def _parallel_execute_serial(cr, queries, logger=_logger):
...
 
     def _parallel_execute_threaded(cr, queries, logger=_logger):
         """
-        Execute queries in parallel
-        Use a maximum of 8 workers (but not more than the number of CPUs)
-        Side effect: the given cursor is commited.
+        Try to execute queries concurrently using `min([8, len(queries), n_cpus])` workers.
+        If concurrency issues are detected, some of the queries will be executed sequentially.
+        Side effect: the given cursor is committed.
         As example, on `**REDACTED**` (using 8 workers), the following gains are:
```
aj-fuentes, line 82: The example table below could be removed. The REDACTED here looks awkward.

Note that if you are documenting this you could instead add docstrings below in `parallel_execute` -- the public API.

Pirols:
> Can you try to add a test to prove it works as expected?

Yes, on it.

src/util/pg.py
Pirols: Yes, I will take the chance to add some documentation to the public APIs.

cawo-odoo:
> It looks correct. I also like that it avoids the mixed returns in the right way.

Me too.
Note that PG automatically roles back when this exception is raised, so it is OK that this version removed the explicit `cr.rollback()`

Pirols:
In trying to reproduce the `DeadlockDetected` error in a test, I actually got a `SerializationError` instead. While considering whether[^1] to catch this error as well, I noticed that it would introduce an inconsistency in the returned cnt value.

TL; DR: Since the cursor sets rowcount to the number of explicitly (e.g. not deleted due to a cascade effect) affected rows the total count from `parallel_execute` may be an underestimate of that quantity.

### Detailed explanation

Assume the following db:

```SQL
CREATE TABLE my_table (id SERIAL PRIMARY KEY, other_id INTEGER, FOREIGN KEY (other_id) REFERENCES my_table ON DELETE CASCADE);

INSERT INTO my_table SELECT GENERATE_SERIES(1, 2);

UPDATE my_table SET other_id = 2 WHERE id = 1;
UPDATE my_table SET other_id = 1 WHERE id = 2;
```

Cursors do not account for the number of rows deleted by CASCADE-effect in rowcount. In fact, running `cr.execute("DELETE FROM my_table WHERE id = 1")` deletes both rows but it sets `cr.rowcount` to 1 (more generally to the number of explicitly affected rows).
This, combined with the `failed_queries` retry mechanism, results in `explode_execute(cr, "DELETE FROM my_table", bucket_size=1)` returning 1 while removing both rows.

To be more precise, `explode_execute(cr, "DELETE FROM my_table", bucket_size=1)` tries to execute two queries, one per row, in parallel. One of the two queries manages to acquire the lock and deletes both rows and - as we saw - sets `rowcount` to 1. The other query fails and is thus added to `failed_queries`; to be retried. However, when the query is eventually re-executed, it runs against an empty table, adding 0 to the total count. The returned total count is thus 1, which underestimates the number of rows explicitly removed.

### Considerations

This does not seem like a terribly big issue:

* The inconsistency is mainly due to how the cursor computes `rowcount` and is not really consistent with the total number of affected (though I only tried with `DELETE FROM`) rows even if not run in `//`.
* So far our migration scripts *very* rarely read the output value of `parallel_execute` and, when they do, the value is used as a flag and not as an actual count.

Wdyt?

[^1]: Even if we don't, the same issue might affect `DeadlockDetected` raising scenarios as well (to be determined).

src/util/pg.py
```diff
@@ -257,6 +272,10 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
 
 
 def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_logger):
+    """
+    Explode the provided query and execute it in parallel.
+    See `explode_query_range()` and `parallel_execute` for more details
```
KangOl, line 277:
```suggestion
    See `explode_query_range()` and `parallel_execute()` for more details
```
or 
```suggestion
    See `explode_query_range` and `parallel_execute` for more details
```

cawo-odoo (approved):
The new choice of exceptions to handle seems good. Reference:
https://www.postgresql.org/docs/current/mvcc-serialization-failure-handling.html
Also, the "miscounting" you describe is not new to this PR, it is just more likely to happen in these situations, and I agree it does not seem harmful.

LGTM

src/util/pg.py
```diff
@@ -112,23 +100,57 @@ def execute(query):
...
+                log_hundred_percent=True,
+            ):
+                try:
+                    tot_cnt += future.result() or 0
+                except psycopg2.OperationalError as exc:
+                    if exc.pgcode in CONCURRENCY_ERRORCODES:
+                        # to be retried without concurrency
+                        failed_queries.append(future_queries[future])
+                    else:
+                        raise
```
cawo-odoo, line 126:
```suggestion
                    if exc.pgcode not in CONCURRENCY_ERRORCODES:
                        raise
                    # to be retried without concurrency
                    failed_queries.append(future_queries[future])
```

Pirols:
Updated the docstring-adding commit to avoid conflicts with https://github.com/odoo/upgrade-util/pull/66

src/util/pg.py
```diff
@@ -112,23 +100,57 @@ def execute(query):
...
+
+    Potential gains example (using 8 workers):
+        +---------------------------------------------+-------------+-------------+
+        | File                                        | Sequential  | Parallel    |
+        +---------------------------------------------+-------------+-------------+
+        | base/saas~12.5.1.3/pre-20-models.py         | ~8 minutes  | ~2 minutes  |
+        | mail/saas~12.5.1.0/pre-migrate.py           | ~10 minutes | ~4 minutes  |
+        | mass_mailing/saas~12.5.2.0/pre-10-models.py | ~40 minutes | ~18 minutes |
+        +---------------------------------------------+-------------+-------------+
+    """
```
aj-fuentes, line 153: We don't want this to be on the online docs if the automatic docs generation is finally merged.
```suggestion
    :meta private: exclude from online docs
    """
```

Once done, then we can revisit this docstring and harmonize it with the docs PR. The same for any other utils that do not have yet clear docs, i.e. those not touched by the docs PR.

cawo-odoo (approved):
This LGTM even more now.

KangOl:
Can you please rebase your branch(es) and fix the conflicts?

Pirols:
> Can you please rebase your branch(es) and fix the conflicts?

Done.

src/util/pg.py
```diff
@@ -117,23 +104,69 @@ def execute(query):
...
 def parallel_execute(cr, queries, logger=_logger):
+    """
+    Execute queries in parallel, using a maximum of 8 workers.
+
+    The actual number of workers is computed as the minimum between the number of queries
+    and that of the availablue cpu cores, capped at 8.
+
+    .. example::
+        .. code-block:: python
+            util.parallel_execute(cr, ["REINDEX TABLE {}".format(t) for t in tables])
```
KangOl, line 152: At least use the correct way of formatting a query.
```suggestion
            util.parallel_execute(cr, [util.format_query(cr, "REINDEX TABLE {}", t) for t in tables])
```

Pirols:
How the docstring looks: 
image

src/util/pg.py
```diff
@@ -117,23 +104,71 @@ def execute(query):
...
+
+        return tot_cnt
 
 else:
     _parallel_execute_threaded = _parallel_execute_serial
 
 
 def parallel_execute(cr, queries, logger=_logger):
+    """
+    Execute queries in parallel, using a maximum of 8 workers.
```
aj-fuentes, line 145:
```suggestion
    Execute queries in parallel.
```

src/util/pg.py
```diff
@@ -117,23 +104,71 @@ def execute(query):
...
+
+    The actual number of workers is computed as the minimum between the number of queries
+    and that of the availablue cpu cores, capped at 8.
```
aj-fuentes, line 148:
```suggestion
```

src/util/pg.py
```diff
@@ -117,23 +104,71 @@ def execute(query):
...
+
+    .. example::
+       .. code-block:: python
+
+          util.parallel_execute(cr, [util.format_query(cr, "REINDEX TABLE {}", t) for t in tables])
+
+    :param list(str) queries: list of queries to execute concurrently
+    :param logger: logger used to report the progress
+    :type logger: :class:`logging.Logger`
```
aj-fuentes, line 157:
```suggestion
    :param `~logging.Logger` logger: logger used to report the progress
```

src/util/pg.py
```diff
@@ -117,23 +104,71 @@ def execute(query):
...
+    :return: the sum of `cr.rowcount` for each query run
+    :rtype: int
+
+    .. warning::
+       Due to the nature of `cr.rowcount`, the return value of this function may represent an
+       underestimate of the real number of affected records. For instance, when some records
+       are deleted/updated as a result of an `ondelete` clause, they won't be taken into account.
+
+    .. note::
+       If a known concurrency issue occurs, the failing queries will be retried without parallelism
```
aj-fuentes, line 167:
```suggestion
       If a known concurrency issue occurs, the *failing* queries will be retried sequentially.
```

aj-fuentes (commented):
The workers used is an implementation detail that we don't need in the documentation. Some other minimal comments.

Pirols:
> The workers used is an implementation detail that we don't need in the documentation. Some other minimal comments.

Suggestions taken and screenshot updated. Also added a tip box.

aj-fuentes:
> How the docstring looks:

For info: we can see the latest docs in the documentation build directly

KangOl (approved):

src/util/pg.py
```diff
@@ -117,23 +104,69 @@ def execute(query):
...
+                    tot_cnt += future.result() or 0
+                except psycopg2.OperationalError as exc:
+                    if exc.pgcode not in CONCURRENCY_ERRORCODES:
+                        raise
+
+                    # to be retried without concurrency
+                    failed_queries.append(future_queries[future])
+
+        if failed_queries:
+            logger.info("Serialize queries that failed due to concurrency issues")
```
nseinlet, line 133:
```suggestion
            logger.warning("Serialize queries that failed due to concurrency issues")
```
Don't let dev makes queries that silently failed to execute in //

nseinlet (approved):

KangOl:
@robodoo rebase-ff r+

PR #50, opened 2024-02-29
[FIX] util/records: do not fail editing views on extra whitespace

aj-fuentes:
XML parser would fail if there are extra context before the document declaration. Even though it's invalid XML we can still attempt to edit the views.


KangOl:
upgradeci retry with always only base report sale theme_common web website website_enterprise website_helpdesk website_sale in 9.0 10.0 12.0 13.0 15.0 16.0 17.0

KangOl:
@robodoo r+

PR #52, opened 2024-03-07
[FIX] util/fields: clean sort in filters

asno-odoo:
In the remove_field, sort is not cleaned up when filters are cleaned. It sometimes can lead to post upgrade issues if someone uses a removed field for their custom filter.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -153,6 +153,28 @@ def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inh
         if changed:
             add_to_migration_reports(("ir.filters", id_, name), "Filters/Dashboards")
 
+    if column_exists(cr, "ir_filters", "sort"):
+        cr.execute(
+            """
+               WITH to_update AS (
+                   SELECT f.id,
+                          COALESCE(ARRAY_TO_JSON(ARRAY_AGG(s.sort_item ORDER BY rn) filter (WHERE s.sort_item not in %s)), '[]') AS sort
+                     FROM ir_filters f
+                 JOIN LATERAL JSONB_ARRAY_ELEMENTS_TEXT(f.sort::jsonb)
+                       WITH ORDINALITY AS s(sort_item, rn)
+                          ON true
+                       WHERE f.model_id = %s
+                         AND f.sort ~ %s
+                  GROUP BY id
+               )
+               UPDATE ir_filters f
+                  SET sort = t.sort
+                 FROM to_update t
+                WHERE f.id = t.id
+            """,
+            [(fieldname, fieldname + " desc"), model, r"\y{}\y".format(fieldname)],
+        )
+
     def adapter(leaf, is_or, negated):
         # replace by TRUE_LEAF, unless negated or in a OR operation but not negated
         if is_or ^ negated:


src/util/fields.py
```diff
@@ -153,6 +153,17 @@ def clean_context(context):
         if changed:
             add_to_migration_reports(("ir.filters", id_, name), "Filters/Dashboards")
 
+    fieldnameQuote = '"{}"'.format(fieldname)
```
KangOl, line 156: Why quoting the field name?

src/util/fields.py
```diff
@@ -153,6 +153,17 @@ def clean_context(context):
...
+    cr.execute(
+        """
+            UPDATE ir_filters
+              SET sort='[' || ARRAY_TO_STRING(ARRAY_REMOVE(STRING_TO_ARRAY(substring(TRIM(sort),2,LENGTH(TRIM(sort))-2),','),%s),',') || ']'
+             WHERE model_id=%s
+               AND sort ~ %s
+        """,
+        [fieldnameQuote, model, fieldnameQuote],
+    )
+
```
KangOl, line 166: Actually, the `sort` column contains a JSON array[^1] (since the start[^2]).
So we can parse it using PostgreSQL JSON functions.
Also, we should handle the "desc" sorts.

```suggestion
    if column_exists(cr, "ir_filters", "sort"):
        cr.execute(
            """
with to_update as (
    select id,
           coalesce(array_to_json(array_agg(sort_item ORDER BY rn) filter (where sort_item not in %s)), '[]') as sort
    from (

        select f.id, s.*
          from ir_filters f
  join lateral jsonb_array_elements_text(f.sort::jsonb)
         with ordinality as s(sort_item, rn)
           on true
        where f.model_id = %s
          and f.sort ~ %s
    ) as matches
   group by id
)
update ir_filters f
   set sort = t.sort
  FROM to_update t
 WHERE f.id = t.id
            """,
            [(fieldname, fieldname + " desc"), model, r"\y{}\y".format(fieldname)],
        )

```



[^1]: in 17.0: https://github.com/odoo/odoo/blob/696389439de6087c643162cf96a3d3f9f4f944a9/addons/web/static/src/search/search_model.js#L2221

[^2]: in 9.0: https://github.com/odoo/odoo/blob/7132a15c38a3b5e5ec232404d21aacfa2ac5874e/addons/web/static/src/js/views/search_menus.js#L234

src/util/fields.py
```diff
@@ -153,6 +153,31 @@ def clean_context(context):
...
+                          COALESCE(array_to_json(array_agg(sort_item ORDER BY rn) filter (WHERE sort_item not in %s)), '[]') AS sort
+                   FROM (
+                       SELECT f.id, s.*
+                         FROM ir_filters f
+                 JOIN LATERAL jsonb_array_elements_text(f.sort::jsonb)
+                       WITH ORDINALITY AS s(sort_item, rn)
+                          ON true
+                       WHERE f.model_id = %s
+                         AND f.sort ~ %s
+                   ) AS matches
```
KangOl, line 170: I simplified the query afterward. See last version https://github.com/odoo/upgrade-util/pull/52#discussion_r1516498360

aj-fuentes: Do we need the sorting complexity? Why not something like below?
```sql
SELECT f.id,
        min(f.sort),
        array_to_json(array_remove(array_remove(array_agg(e.value), 'login'), 'login desc'))::text
   FROM ir_filters f
   JOIN LATERAL json_array_elements_text(f.sort::JSON) e
     ON TRUE
  WHERE f.model_id='res.users'
    AND f.sort ~ '\ylogin\y'
  GROUP BY f.id
```
This avoids ordering and such. The `min(f.sort)` was just for quick test
```
+----+------------------------------+-----------------+
| id | min                          | array_to_json   |
|----+------------------------------+-----------------|
| 1  | ["login","lang","name"]      | ["lang","name"] |
| 3  | ["login desc","lang","name"] | ["lang","name"] |
| 2  | ["login","lang","name"]      | ["lang","name"] |
+----+------------------------------+-----------------
```

```

KangOl: - `array_agg` will use an arbitrary order if none specified.
- `array_remove` can be used but `filter` have the same effect.

KangOl:
@robodoo r+

PR #53, opened 2024-03-07
[IMP] util/{domains,fields}: replace fields in server actions' update_path

aj-fuentes:
`update_path` is used by server actions to know what field to update. It
needs to be updated when renaming fields.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -7,6 +7,7 @@ import warnings
 
 import psycopg2
 from psycopg2 import sql
+from psycopg2.extras import Json
 
 try:
     from odoo import release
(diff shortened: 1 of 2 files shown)


src/util/fields.py
```diff
@@ -855,6 +856,32 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
+            SELECT a.id,
+                   a.update_path,
+                   m.model
+              FROM ir_act_server a
+              JOIN ir_model m
+                ON a.model_id = m.id
+             WHERE a.state = 'object_write'
+               AND a.update_path ~ %s
+            """,
+            [r"\y{}\y".format(re.escape(old))],
```
KangOl, line 871: Consistence with other queries in this function
```suggestion
               AND a.update_path ~ %(old)s
            """,
            p,
```

Or at least (but less readable)

```suggestion
               AND a.update_path ~ %s
            """,
            [p["old"]],
```

KangOl (changes requested):

PR #56, opened 2024-03-14
[FIX] util/pg: ensure the ranges contain at least bucket_size elements

aj-fuentes:
Before this patch when we exploded a query by range we could end up
creating too many queries for which there is no actual record in the
range. In extreme cases this causes a memory error doe to the huge list
of generated queries.

Example:
```
=> select min(id),max(id),count(id) from account_journal
+-----+----------+-------+
| min | max      | count |
|-----+----------+-------|
| 1   | 10000003 | 56    |
+-----+----------+-------+
```

All the journals could be processed in one bucket of >=56 elements. The
high max id causes the generation of many queries that would not process
any actual record.

```
...
  File "/tmp/tmpl7nu4jh5/migrations/account/saas~16.2.1.2/pre-migrate.py", line 365, in migrate
    util.explode_execute(cr, query, table="account_journal", alias="journal", bucket_size=1)
  File "/tmp/tmpl7nu4jh5/migrations/util/pg.py", line 226, in explode_execute
    explode_query_range(cr, query, table, alias=alias, bucket_size=bucket_size),
  File "/tmp/tmpl7nu4jh5/migrations/util/pg.py", line 217, in explode_query_range
    return [
  File "/tmp/tmpl7nu4jh5/migrations/util/pg.py", line 218, in <listcomp>
    cr.mogrify(query, {"lower-bound": index, "upper-bound": index + bucket_size - 1}).decode()
  File "/home/odoo/src/odoo/17.0/odoo/sql_db.py", line 316, in mogrify
    return self._obj.mogrify(query, params)
MemoryError
```

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -485,6 +485,28 @@ class TestPG(UnitTestCase):
             cr.execute(q)
             self.assertTrue(all(x for (x,) in cr.fetchall()))
 
+    @mute_logger("odoo.upgrade.util.pg.explode_query_range")
+    def test_explode_query_range(self):
+        cr = self.env.cr
+
+        cr.execute("SELECT count(id) FROM res_partner_title")
+        count = cr.fetchone()[0]
+        # ensure there start with at least 10 records
+        for _ in range(10 - count):
+            count += 1
+            self.env["res.partner.title"].create({"name": "x"})
+
+        # set one record with very high id
+        tid = self.env["res.partner.title"].create({"name": "x"}).id
+        count += 1
+        cr.execute("UPDATE res_partner_title SET id = 10000000 WHERE id = %s", [tid])
+
+        qs = util.explode_query_range(cr, "SELECT 1", table="res_partner_title", bucket_size=count)
+        self.assertEqual(len(qs), 1)  # one bucket should be enough for all records
+
+        qs = util.explode_query_range(cr, "SELECT 1", table="res_partner_title", bucket_size=count - 1)
+        self.assertEqual(len(qs), 1)  # 10% rule for second bucket, 1 <= 0.1(count - 1) since count >= 11
+
     def test_parallel_rowcount(self):
         cr = self.env.cr
         cr.execute("SELECT count(*) FROM res_lang")
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -21,6 +21,11 @@ try:
 except ImportError:
     from UserList import UserList
 
+try:  # noqa: SIM105
+    range = xrange  # noqa: A001
+except NameError:
+    pass
+
 import psycopg2
 from psycopg2 import sql
 
@@ -36,6 +41,7 @@ from .misc import log_progress
 _logger = logging.getLogger(__name__)
 
 ON_DELETE_ACTIONS = frozenset(("SET NULL", "CASCADE", "RESTRICT", "NO ACTION", "SET DEFAULT"))
+MAX_BUCKETS = int(os.getenv("MAX_BUCKETS", "150000"))
 
 
 class PGRegexp(str):
@@ -196,27 +202,57 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
 
     alias = alias or table
 
-    cr.execute("SELECT min(id), max(id) FROM {}".format(table))
-    min_id, max_id = cr.fetchone()
-    if min_id is None:
-        return []  # empty table
-
     if "{parallel_filter}" not in query:
         sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
         query += sep_kw + "{parallel_filter}"
 
-    if ((max_id - min_id + 1) * 0.9) <= bucket_size:
-        # If there is less than `bucket_size` records (with a 10% tolerance), no need to explode the query.
-        # Force usage of `prefix` in the query to validate it correctness.
-        # If we don't the query may only be valid if there is no split. It avoid scripts to pass the CI but fail in production.
+    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
+    min_id, max_id = cr.fetchone()
+    if min_id is None:
+        return []  # empty table
+    count = (max_id + 1 - min_id) // bucket_size
+    if count > MAX_BUCKETS:
+        _logger.getChild("explode_query_range").warning(
+            "High number of queries generated (%s); switching to a precise bucketing strategy", count
+        )
+        cr.execute(
+            format_query(
+                cr,
+                """
+                WITH t AS (
+                    SELECT id,
+                           mod(row_number() OVER(ORDER BY id) - 1, %s) AS g
+                      FROM {table}
+                     ORDER BY id
+                ) SELECT array_agg(id ORDER BY id) FILTER (WHERE g=0),
+                         min(id),
+                         max(id)
+                    FROM t
+                """,
+                table=table,
+            ),
+            [bucket_size],
+        )
+        ids, min_id, max_id = cr.fetchone()
+    else:
+        ids = list(range(min_id, max_id + 1, bucket_size))
+
+    assert min_id == ids[0] and max_id + 1 != ids[-1]  # sanity checks
+    ids.append(max_id + 1)  # ensure last bucket covers whole range
+    # `ids` holds a list of values marking the interval boundaries for all buckets
+
+    if (max_id - min_id + 1) <= 1.1 * bucket_size or (len(ids) == 3 and ids[2] - ids[1] <= 0.1 * bucket_size):
+        # If we return one query `parallel_execute` skip spawning new threads. Thus we return only one query if we have
+        # only two buckets and the second would have at most 10% of bucket_size records.
+        # Still, since the query may only be valid if there is no split, we force the usage of `prefix` in the query to
+        # validate its correctness and avoid scripts that pass the CI but fail in production.
         parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
         return [query.format(parallel_filter=parallel_filter)]
 
     parallel_filter = "{alias}.id BETWEEN %(lower-bound)s AND %(upper-bound)s".format(alias=alias)
     query = query.replace("%", "%%").format(parallel_filter=parallel_filter)
     return [
-        cr.mogrify(query, {"lower-bound": index, "upper-bound": index + bucket_size - 1}).decode()
-        for index in range(min_id, max_id, bucket_size)
+        cr.mogrify(query, {"lower-bound": ids[i], "upper-bound": ids[i + 1] - 1}).decode() for i in range(len(ids) - 1)
     ]
 
 


src/util/pg.py
```diff
@@ -196,27 +196,41 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
         return []  # empty table
 
     if "{parallel_filter}" not in query:
         sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
         query += sep_kw + "{parallel_filter}"
 
-    if ((max_id - min_id + 1) * 0.9) <= bucket_size:
-        # If there is less than `bucket_size` records (with a 10% tolerance), no need to explode the query.
-        # Force usage of `prefix` in the query to validate it correctness.
+    if cr.rowcount == 1:
```
aj-fuentes, line 223: To keep the same "intended" behavior, we could check if there are only 2 buckets then count if the total number of records is at most 10% more than bucket_size. @KangOl should we do it?

src/util/pg.py
```diff
@@ -196,27 +196,41 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
-    cr.execute("SELECT min(id), max(id) FROM {}".format(table))
-    min_id, max_id = cr.fetchone()
-    if min_id is None:
+    cr.execute(
+        format_query(
+            cr,
+            """
+            SELECT min(id), max(id)
+              FROM (SELECT id,
+                           ROW_NUMBER() OVER (ORDER BY id) / %s
```
KangOl, line 205: `row_number` start at 1.
```suggestion
                           (ROW_NUMBER() OVER () - 1) / %s
```

src/util/pg.py
KangOl: Now that we have exactly the correct number of records in the buckets, the 10% tolerance is not necessary.

src/util/pg.py
```diff
@@ -196,27 +196,41 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+            table=table,
+        ),
+        [bucket_size],
+    )
+    if not cr.rowcount:
         return []  # empty table
 
     if "{parallel_filter}" not in query:
         sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
         query += sep_kw + "{parallel_filter}"
```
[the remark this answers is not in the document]
aj-fuentes: I had the same impression, moving it.

src/util/pg.py
```diff
@@ -196,27 +196,41 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
-    if min_id is None:
+    cr.execute(
+        format_query(
+            cr,
+            """
+            SELECT min(id), max(id)
+              FROM (SELECT id,
+                           ROW_NUMBER() OVER (ORDER BY id) / %s
+                      FROM {table}
+                     ORDER BY id
```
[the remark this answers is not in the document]
Pirols: My suggestion concerned the table-level order by, not the one in the window:

```SQL
pied_1345804_17.0> SELECT id, (ROW_NUMBER() OVER (ORDER BY id) -1) / 2 FROM UNNEST(ARRAY[1, 10, 2, 3]) AS t(id)
+----+----------+
| id | ?column? |
|----+----------|
| 1  | 0        |
| 2  | 0        |
| 3  | 1        |
| 10 | 1        |
+----+----------+
SELECT 4
Time: 0.015s
pied_1345804_17.0> SELECT id, (ROW_NUMBER() OVER () -1) / 2 FROM UNNEST(ARRAY[1, 10, 2, 3]) AS t(id) ORDER BY id
+----+----------+
| id | ?column? |
|----+----------|
| 1  | 0        |
| 2  | 1        |
| 3  | 1        |
| 10 | 0        |
+----+----------+

```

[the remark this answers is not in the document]
aj-fuentes: If we **don't** remove the order by from the window -- which seems to be wrong according to the [error](https://runbot.odoo.com/runbot/build/59837160) we get in runbot. Then the global order by id is just to get an ordered set of buckets, we could achieve the same with `order by g`, the advantage of `order by id` is that id should be already indexed -- faster order. Let me push back the query with both `order by`s.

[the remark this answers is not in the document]
aj-fuentes: Kept `order by g` for now. In the _main_ query. I just noticed you were referring to the subquery.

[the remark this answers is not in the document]
Pirols: > If we **don't** remove the order by from the window -- which seems to be wrong according to the [error](https://runbot.odoo.com/runbot/build/59837160) we get in runbot. Then the global order by id is just to get an ordered set of buckets, we could achieve the same with `order by g`, the advantage of `order by id` is that id should be already indexed -- faster order. Let me push back the query with both `order by`s.

Yes, my original thought was that the two table-level `ORDER BY`s were redundant to each other. I agree it's best to keep the `BY id` one.

src/util/pg.py
```diff
@@ -196,27 +196,40 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+        ),
+        [bucket_size],
+    )
+    if not cr.rowcount:
+        return []  # empty table
+
+    if cr.rowcount == 1:
+        # Only one bucket, no need to explode the query.
+        # Since the query may only be valid if there is no split, force usage of `prefix` in the query to validate its
+        # correctness and avoid scripts that pass the CI but fail in production.
```
Pirols, line 225:
```suggestion
        # Since the query may only be valid if there is no split, force usage of `prefix` in the query to validate its
        # correctness and avoid scripts that pass the CI but fail in production.
```

I guess I'm lacking some context to understand this comment however I'm under the impression that the single-bucket case would be handled correctly without this if-branch (parallel_filter would be `BETWEEN <min(id)> AND <max(id)>` instead of `id IS NOT NULL`, I don't see a problem with that). Am I wrong?

aj-fuentes: The idea is to ensure we validate the query always with the parallel filter. To avoid it failing later when we _actually_ try to use it.

src/util/pg.py
```diff
@@ -196,27 +196,56 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+                """,
+                table=table,
+            ),
+            [bucket_size],
+        )
+        ids, min_id, max_id = cr.fetchone()
+        if min_id is None:
+            return []  # empty table
+
+    assert min_id == ids[0] and max_id + 1 != ids[-1]  # sanity checks
```
cawo-odoo, line 234:
```suggestion
                         max(id)
                    FROM t
                """,
                table=table,
            ),
            [bucket_size],
        )
        ids, max_id = cr.fetchone()
        if max_id is None:
            return []  # empty table

    min_id = ids[0]
```
I don't think we need to assume bugs in PostgreSQL, here.

aj-fuentes: It's not PG whom I doubt, it's _us_. This will server as a check point if anything done a few lines above is adapted in the future ;)

src/util/pg.py
```diff
@@ -196,27 +196,58 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+    if "{parallel_filter}" not in query:
+        sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
+        query += sep_kw + "{parallel_filter}"
+
+    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
     min_id, max_id = cr.fetchone()
     if min_id is None:
         return []  # empty table
+    count = (max_id + 1 - min_id) // bucket_size
+    if count > 100000:
```
cawo-odoo, line 208: I find this threshold to be problematic. For a small table with an off sequence space this may be good, for bigger tables, it is unclear why this threshold is reached. How about determining the proportion of sequence space (max - min) to the  live tuple count? Something like
```suggestion
    cr.execute(format_query(cr, "SELECT min(id), max(id), count(id) FROM {}", table))
    min_id, max_id, count_id = cr.fetchone()
    if min_id is None:
        return []  # empty table
    if (max_id + 1 - min_id) / count_id > 10:
```

src/util/pg.py
```diff
@@ -196,27 +196,58 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+            return []  # empty table
+    else:
+        ids = list(range(min_id, max_id + 1, bucket_size))
 
-    if "{parallel_filter}" not in query:
-        sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
-        query += sep_kw + "{parallel_filter}"
+    assert min_id == ids[0] and max_id + 1 != ids[-1]  # sanity checks
+    ids.append(max_id + 1)  # ensure last bucket covers whole range
+    # `ids` holds a list of values that will cover all ids in pairwise half-open intervals
```
cawo-odoo, line 238:
```suggestion
    # `ids` holds a list of values marking the interval boundaries for all buckets
```
I think it is confusing to talk of half-open intervals contained in `ids`, here. The half-open intervals are a result of calculations we do on this list, later.

src/util/pg.py
aj-fuentes: We discussed this in DM (with @KangOl)

Empty buckets are fine as long as they run fast. If we do as you propose we would be switching to the slower strategy for queries that otherwise will just run fast. OTOH the _real_ issue here is the MemoryError. Thus average-//-query-size * 100K looks acceptable as a threshold. We can adapt the value in the future if it proves problematic.

The precise bucketing could be slow due to (already tested) high `Heap fetches`. In the end there is no good option for all cases.

src/util/pg.py
aj-fuentes: `holds a list of values` not a `list of half-open intervals` :D

src/util/pg.py
KangOl, line 208: As it is to avoid potential MemoryError, it depends on the machine it runs on.
I would do it like `MAX_WORKERS` and define this threshold behind an env variable.
```suggestion
    if count > int(os.getenv("MAX_BUCKETS", "100000")):
```

src/util/pg.py
```diff
@@ -196,27 +196,58 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+                         max(id)
+                    FROM t
+                """,
+                table=table,
+            ),
+            [bucket_size],
+        )
+        ids, min_id, max_id = cr.fetchone()
+        if min_id is None:
+            return []  # empty table
```
KangOl, line 232: Should never be true (already verified line 205)

src/util/pg.py
aj-fuentes: I increased the value to 150K -> found a case in our logs with 129K that ran without issues (relatively fast). cc: @cawo-odoo

src/util/pg.py
```diff
@@ -196,27 +197,56 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+        query += sep_kw + "{parallel_filter}"
+
+    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
     min_id, max_id = cr.fetchone()
     if min_id is None:
         return []  # empty table
+    count = (max_id + 1 - min_id) // bucket_size
+    if count > MAX_BUCKETS:
+        _logger.getChild("explode_query_range").warning(
+            "More than %s queries generated, switching to a precise bucketing strategy", count
```
KangOl, line 211:
```suggestion
            "High number of queries generated (%s); switching to a precise bucketing strategy", count
```

src/util/pg.py
```diff
@@ -196,27 +197,56 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
+        # correctness and avoid scripts that pass the CI but fail in production.
         parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
         return [query.format(parallel_filter=parallel_filter)]
 
     parallel_filter = "{alias}.id BETWEEN %(lower-bound)s AND %(upper-bound)s".format(alias=alias)
     query = query.replace("%", "%%").format(parallel_filter=parallel_filter)
     return [
-        cr.mogrify(query, {"lower-bound": index, "upper-bound": index + bucket_size - 1}).decode()
-        for index in range(min_id, max_id, bucket_size)
+        cr.mogrify(query, {"lower-bound": low, "upper-bound": upp - 1}).decode() for (low, upp) in zip(ids, ids[1:])
```
KangOl, line 249: Use [itertools.pairwise](https://docs.python.org/3.12/library/itertools.html#itertools.pairwise) instead (we should have a shim in `util.misc` (note: python2 code should use `izip`))
For big buckets, `ids` will be huge, so better not copy it.

KangOl (changes requested):

KangOl:
Can a test be added?

src/util/pg.py
aj-fuentes: Well, as much as I like pairwise:
> New in version 3.10.

But we don't need to overdo it -- the good old C-style iteration should be fine here.

aj-fuentes:
> Can a test be added?

Pushed something... not tested locally :/

src/util/pg.py
```diff
@@ -196,27 +197,57 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
 
-    if ((max_id - min_id + 1) * 0.9) <= bucket_size:
-        # If there is less than `bucket_size` records (with a 10% tolerance), no need to explode the query.
-        # Force usage of `prefix` in the query to validate it correctness.
-        # If we don't the query may only be valid if there is no split. It avoid scripts to pass the CI but fail in production.
+    if (max_id - min_id + 1) <= 1.1 * bucket_size or (len(ids) == 3 and ids[2] - ids[1] <= 0.1 * bucket_size):
+        # If we return one query `parallel_execute` skip spawning new threads. Thus we return only one query if we have
+        # only two buckets and the second would have at most 10% of bucket_size records.
+        # Still, since the query may only be valid if there is no split, we force the usage of `prefix` in the query to
+        # validate its correctness and avoid scripts that pass the CI but fail in production.
```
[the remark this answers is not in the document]
Pirols: That looks much more understandable to me!

src/base/tests/test_util.py
```diff
@@ -485,6 +485,13 @@ def test_explode_mult_filters(self):
             cr.execute(q)
             self.assertTrue(all(x for (x,) in cr.fetchall()))
 
+    def test_precise_bucket(self):
+        cr = self.env.cr
+        pt = self.env["res.partner.title"].create({"name": "x"})
+        cr.execute("UPDATE res_partner_title SET id = 10000000 WHERE id = %s", [pt.id])
+        qs = util.explode_query_range(cr, "SELECT 1", table="res_partner_title", bucket_size=10)
```
KangOl, line 492:
```
❯ rg -c -F res.partner.title addons/l10n_cr/data/l10n_cr_res_partner_title.xml
14
```

```suggestion
        qs = util.explode_query_range(cr, "SELECT 1", table="res_partner_title", bucket_size=100)
```

aj-fuentes: Working on it ;)

src/base/tests/test_util.py
```diff
@@ -485,6 +485,27 @@ def test_explode_mult_filters(self):
...
+        # set one record with very high id
+        tid = self.env["res.partner.title"].create({"name": "x"}).id
+        count += 1
+        cr.execute("UPDATE res_partner_title SET id = 10000000 WHERE id = %s", [tid])
+
+        qs = util.explode_query_range(cr, "SELECT 1", table="res_partner_title", bucket_size=count)
+        self.assertEqual(len(qs), 1)  # one bucket should be enough for all records
+
+        qs = util.explode_query_range(cr, "SELECT 1", table="res_partner_title", bucket_size=count - 1)
+        self.assertEqual(len(qs), 1)  # 10% rule for second bucket, 1 <= 0.1(count - 1) since count >= 11
```
KangOl, line 507: Only true if count is 11.
This is not always the case, like with `l10n_cr`.

aj-fuentes: This is true if count >**=** 11, which is guaranteed by
https://github.com/odoo/upgrade-util/pull/56/files#diff-d21f53fbfb10b7592abbfa97c35cfb969bbf545b9e36a44ed5069b59f438e350R493-R499

aj-fuentes: In other words the record created in line 499 is at _least_ the 11th.

KangOl:
upgradeci retry with always only base l10n_cr

src/base/tests/test_util.py
```diff
@@ -485,6 +485,27 @@ def test_explode_mult_filters(self):
             cr.execute(q)
             self.assertTrue(all(x for (x,) in cr.fetchall()))
 
+    def test_explode_query_range(self):
```
KangOl, line 488:
```suggestion
    @mute_logger("odoo.upgrade.util.pg.explode_query_range"}.
    def test_explode_query_range(self):
```

aj-fuentes: Done, also the last run with l10n_cr was green :) https://upgradeci.odoo.com/upgradeci/change/KktpRCzA6naVmtVyUHtA9Q

src/util/pg.py
```diff
@@ -196,27 +197,57 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
...
         parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
         return [query.format(parallel_filter=parallel_filter)]
 
     parallel_filter = "{alias}.id BETWEEN %(lower-bound)s AND %(upper-bound)s".format(alias=alias)
     query = query.replace("%", "%%").format(parallel_filter=parallel_filter)
     return [
-        cr.mogrify(query, {"lower-bound": index, "upper-bound": index + bucket_size - 1}).decode()
-        for index in range(min_id, max_id, bucket_size)
+        cr.mogrify(query, {"lower-bound": ids[i], "upper-bound": ids[i + 1] - 1}).decode() for i in range(len(ids) - 1)
```
KangOl, line 250: In python2, `range` is not lazy and will return a list. We need to use [xrange](https://docs.python.org/2/library/functions.html#xrange) instead.
I would add the following at the top of the file
```python
try:
    range = xrange
except NameError:
    pass
```

KangOl:
@robodoo r+

KangOl (approved):

PR #58, opened 2024-03-20
[FIX] util/snippets: generate HTML instead of XML for mailing.mailing

bso-odoo:
The `convert_html_content` tool writes XML values if a change happened. This is a problem for the fields of `mailing.mailing` because they strictly expect HTML content.
E.g. if `` becomes ``, the wysiwyg editor displays an empty template instead of the actual content.

This commit avoids this by making the writes to the fields of `mailing.mailing` formatted in HTML.

task-3797854

### DIFF
--- a/src/util/snippets.py
+++ b/src/util/snippets.py
@@ -186,6 +186,12 @@ class HTMLConverter:
     def __init__(self, callback, selector=None):
         self.selector = selector
         self.callback = make_pickleable_callback(callback)
+        self.format = "xml"
+
+    def set_format(self, format):
+        # Using a format string instead of html and etree because self
+        # needs to be pickleable.
+        self.format = format
 
     def has_changed(self, els):
         if self.selector:
@@ -206,8 +212,9 @@ class HTMLConverter:
             parser=utf8_parser,
         )
         has_changed = self.has_changed(els)
+        formatter = etree if self.format == "xml" else html
         new_content = (
-            re.sub(r"(^<wrap>|</wrap>$)", "", etree.tostring(els, encoding="unicode").strip())
+            re.sub(r"(^<wrap>|</wrap>$)", "", formatter.tostring(els, encoding="unicode").strip())
             if has_changed
             else content
         )
@@ -322,6 +329,8 @@ def convert_html_content(
     :param dict kwargs: extra keyword arguments to pass to :func:`convert_html_column`
     """
 
+    if "set_format" in dir(converter_callback):
+        converter_callback.set_format('xml')
     convert_html_columns(
         cr,
         "ir_ui_view",
@@ -332,4 +341,8 @@ def convert_html_content(
     )
 
     for table, columns in html_fields(cr):
+        if "set_format" in dir(converter_callback):
+            # mass_mailing requires HTML
+            converter = "html" if table == "mailing_mailing" else "xml"
+            converter_callback.set_format(converter)
         convert_html_columns(cr, table, columns, converter_callback, where_column=where_column, **kwargs)


bso-odoo:
Related to https://github.com/odoo/upgrade/pull/5808

src/util/snippets.py
```diff
@@ -322,6 +329,8 @@ def convert_html_content(
     :param dict kwargs: extra keyword arguments to pass to :func:`convert_html_column`
     """
 
+    if "set_format" in dir(converter_callback):
```
qsm-odoo, line 332: Do we need this `if` ?

src/util/snippets.py
```diff
@@ -332,4 +341,8 @@ def convert_html_content(
     )
 
     for table, columns in html_fields(cr):
+        if "set_format" in dir(converter_callback):
+            # mass_mailing requires HTML
+            converter = "html" if table == "mailing_mailing" else "xml"
+            converter_callback.set_format(converter)
```
qsm-odoo, line 347: - Should not this be a thing *per column* instead of *per table* ?
- Can't this be somehow guessed by checking the column definition ? Somewhere in our codebase, there must already be that kind of check no? When you save that column or when you try to translate it, don't we have any code guessing that?

> E.g. if `` becomes ``, the wysiwyg
editor displays an empty template instead of the actual content.

Actually, why is that a problem? I don't understand how adding the auto-close marker would make the HTML field invalid.

qsm-odoo (changes requested):

src/util/snippets.py
bso-odoo: Since `upgrade-util` is an open repo, it might be used with a `converter_callback` that is not the expected `HTMLConverter` from this class.

src/util/snippets.py
bso-odoo: Still trying to find out "why" this is a problem, but it is a display problem: if an `` is in the mail `body_arch` and `body_html`, the backend is fine with it, it gets sent as is to the client as the result of `web_read`, but the display shows an empty iframe - and an error occurs if you start dragging a block.

qsm-odoo: Maybe it's related to what the mailing editor does. I think they create an empty iframe then fill it with content. Maybe that content goes through the jQuery parser first or something like that? :shrug:

aj-fuentes: It could be also that it's invalid XML, still valid HTML, and some intermediate parser is too strict... I have no idea what's done in the frontend -- XHTML maybe? :/

bso-odoo: The closing `` tag makes the browser put the HTML content inside `<style id="design-element">` instead of after - thus not actually displaying it.

bso-odoo: > Can't this be somehow guessed by checking the column definition ? Somewhere in our codebase, there must already be that kind of check no? When you save that column or when you try to translate it, don't we have any code guessing that?

Unfortunately there is nothing inside the `ir_model_fields` table that gives this kind of information. And in `upgrade` we cannot even look at the value of the `translate` attribute of the `Html` fields - and if we could, it would not be complete.
Given that we did not notice other issues with XML-mode until now, I think we can only rely on a white list.

We could also go for a more general approach: XML only for `arch_db`, and use HTML for all `Html` fields. But doesn't that sound risky ?

qsm-odoo: > The closing `` tag makes the browser put the HTML content inside `<style id="design-element">` instead of after - thus not actually displaying it. Happens with: `iframeTarget.innerHTML = value;`. E.g.:
> 
> ```js
> el.innerHTML='<div class="o_layout oe_unremovable oe_unmovable bg-200 o_empty_theme" data-name="Mailing"><style id="design-element"/><div class="container o_mail_wrapper o_mail_regular oe_unremovable"><div class="row"><div class="col o_mail_no_options o_mail_wrapper_td bg-white oe_structure o_editable theme_selection_done"><div class="s_text_image o_mail_snippet_general pt32 pb32" data-snippet="s_image_text" data-name="Image - Text">\n        <div class="container">\n            <div class="row align-items-center">\n                <div class="col-lg-6 o_cc px-0">\n                    \n                </div>\n                <div class="col-lg-6 o_cc pt16 pb16">\n                    <h3 class="o_default_snippet_text">Omnichannel sales</h3>\n                    <p style="text-align: justify;" class="o_default_snippet_text">Get your inside sales (CRM) fully integrated with online sales (eCommerce), in-store sales (Point of Sale) and marketplaces like eBay and Amazon.</p>\n                    <div style="text-align: left;">\n                        <a href="#" class="btn btn-link o_default_snippet_text">Read More</a>\n                    </div>\n                </div>\n            </div>\n        </div>\n    </div></div></div></div></div>';
> ```

@bso-odoo Is not that line the problem? `.innerHTML = "<style id="stuff"/>"` is not something valid to do :scream:  Style is not an auto-closing tag, you cannot code that the jQuery way :grimacing: Surprised it's not already full of bugs without this migration issue :open_mouth: 

I think this should be fixed then we can continue just parsing everything as XML? (Unless of course there is more I don't know about between XML/HTML parsing)

qsm-odoo: @bso-odoo In fact, it of course would still be great to also use the right formatter (but still need a fix for the `innerHTML` code). I see we always parse as HTML then re-transform as XML, weird indeed. Can't there be problem in XML too because of that?

Random idea: what about guessing which seems to be the best by:
- parse original content with html, reformat it again as an HTML string
- parse original content with etree, reformat it again as an XML string
- check which one is the closest as the original string => this is the one to use to upgrade the content.

Not sure though, to investigate. But my guess is that:
- when doing that, if you do not manage to guess: using either XML or HTML would be fine (but use XML, IMO less chance to break properly coded functional code)
- when doing that, if you do see one that change the original content more, then it's very probably not the one to use: if it is a mistake, then the original Odoo code that saved that content as it was is properly flawed in the first place.

To see how doing that would impact perf but my guess is that it would be negligible?

bso-odoo: It seems this part of the issue was solved in the meantime: `snippets.py` can now be used through `HTMLConverter` or `QWebConverter`.

PR #59, opened 2024-03-21
[FIX] upgrade report: fix disabled/overwritten views links on saas~17.2

xavieralt:
With path-based routing added to saas~17.2 we need to force the root relative url as the discuss channel url will not be `/web` anymore, but `/odoo/discuss.channel_*/discuss'.

### DIFF
--- a/src/util/report-migration.xml
+++ b/src/util/report-migration.xml
@@ -9,7 +9,6 @@
           'Overridden views': ['The below standard views have been overriden with their original content during the migration because the changes made in them were no longer valid.', '%(xml_id)s'],
           'Disabled views': ['The below custom views have been disabled during the migration because they were no longer valid.', '%(name)s'],
         }"/>
-        <t t-set="view_action_link" t-value="'web?debug=1#view_type=form&amp;model=ir.ui.view&amp;action=%s&amp;id=%%s' % action_view_id"/>
         <t t-set="message_type" t-value="view_message_types[category]"/>
         <li><details t-if="messages[category]">
           <summary>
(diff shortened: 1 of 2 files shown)


src/util/report-migration.xml
```diff
@@ -39,7 +39,7 @@
           <summary>During the upgrade some fields have been removed. The records below have been automatically corrected.</summary>
           <ul>
             <t t-foreach="messages[category]" t-as="message">
-              <li><a t-att-href="'web?debug=1#view_type=form&amp;model=%s&amp;id=%s' % (message[0][0], message[0][1])" t-esc="message[0][2]" target="_blank"/></li>
+              <li><a t-att-href="'/web?debug=1#view_type=form&amp;model=%s&amp;id=%s' % (message[0][0], message[0][1])" t-esc="message[0][2]" target="_blank"/></li>
```
KangOl, line 42: I would instead use the `get_anchor_link_to_record` helper (which then need to be given in the `render_template` call).
```suggestion
              <li><t t-raw="get_anchor_link_to_record(*message[0])"/></li>
```

KangOl:
upgradeci retry with always only account account_asset account_reports analytic base base_automation delivery_sendcloud documents hr_holidays hr_payroll l10n_mx_reports_closing l10n_ph mail mrp payment_paypal sale_subscription website

KangOl:
The `matt` error is not related to this PR.

@robodoo r+

fmdl:
@KangOl when it is push in migration script ?

PR #60, opened 2024-03-25
[FIX] adapt test to `sale_subscription` module

KangOl:
The upgrade of the `sale_subscription` module remove the model `sale.subscription` but keep the table. As this table as a FK on `res.country`, the `test_replace_record_references_batch__uniqueness` test needs to be adapted to ignore this table.

This avoid a `matt` error when upgrading the `sale_subscription` module in version 16.0: https://upgradeci.odoo.com/upgradeci/run/125497


KangOl:
upgradeci retry with always only sale_subscription

KangOl:
@robodoo r+

PR #64, opened 2024-04-03
[FIX] util/report: fix hyper link in migration report

sagu-odoo:
**Step to reproduce:**
1. create db in 14.0 migrate to 15.0 or onwards
2. studio view is disable somehow in migration report tag is coming as text not with hyperlink

**Why this is happening?.**
   earlier t-raw is using for showing html tag
    same as it is but from 15 onwards t-raw is depricated t-out is in use but t-out is escape the html content during rendering and 
    coming like this
```
 &lt;a target="_blank" href="/web?debug=1#view_type=form&amp;amp;model=ir.ui.view&amp;amp;action=28&amp;amp;id=896"&gt;Odoo Studio: mrp.production.form customization&lt;/a&gt;+
```
but it should come like this to behave correctly 
```
 <a target="_blank" href="http://localhost:8069/web?debug=1#view_type=form&amp;model=ir.ui.view&amp;action=28&amp;id=896">Odoo Studio: mrp.production.form customization</a>
```
after this [commit](https://github.com/odoo/upgrade-util/pull/59) t-raw is used and t-raw is replacing [here](https://github.com/odoo/upgrade-util/blob/06085f5ed5c5d604084a5104ae82cd996e464a65/src/util/report.py#L89) as being depricated 

 For fixing this issue unescape the html content in message

### Before fix:
beforei
### After fix:
afteri


KangOl:
This looks wrong.
Actually, `get_anchor_link_to_record` should return a Markup record.

sagu-odoo:
> This looks wrong. Actually, `get_anchor_link_to_record` should return a Markup record.

@KangOl Thanks for the review and guidance. The suggested changes made; could you please have a look?. ;)

KangOl:

KangOl:
@robodoo r+

PR #67, opened 2024-04-09
[FIX] handle the '/' operator when evaluating expression

KangOl:
The `__div__` / `__rdiv__` methods are only supported by python2.


aj-fuentes (approved):

KangOl:

PR #68, opened 2024-04-15
[IMP] spreadsheet: implement a light tokenizer

VincentSchippefilt:
It happens that during migrations we need some kind of tokenized formula in spreadsheet. This helper will tokenize a string in a similar fashion as spreadsheet.

Task: 3761476

### DIFF
--- a/src/spreadsheet/tokenizer.py
+++ b/src/spreadsheet/tokenizer.py
@@ -1,0 +1,248 @@
+import re
+
+"""
+This entire file is a direct translation of the original JavaScript code found in https://github.com/odoo/o-spreadsheet/blob/master/src/formulas/tokenizer.ts.
+"""
+
+
+class CellErrorType:
+    NotAvailable = "#N/A"
+    InvalidReference = "#REF"
+    BadExpression = "#BAD_EXPR"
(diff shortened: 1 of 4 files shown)


src/spreadsheet/tokenizer.py
```diff
@@ -0,0 +1,165 @@
...
+STRING = "STRING"
+BOOLEAN = "BOOLEAN"
+UNKNOWN = "UNKNOWN"
+OPERATOR = "OPERATOR"
+SYMBOL = "SYMBOL"
+FUNCTION = "FUNCTION"
+LEFT_PAREN = "LEFT_PAREN"
+RIGHT_PAREN = "RIGHT_PAREN"
+COMMA = "COMMA"
+DEBUGGER = "DEBUGGER"
```
KangOl, line 24: Define these as [an Enum](https://docs.python.org/3/library/enum.html)?

VincentSchippefilt: this version of the tokenizer didn't work. I converted the complete one of o-spreadsheet.

KangOl (approved):
@robodoo r+

PR #70, opened 2024-04-17
[IMP] util/records: add missing doc details

aj-fuentes:
`delete_unused` also marks as noupdate.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1091,6 +1091,9 @@ def delete_unused(cr, *xmlids, **kwargs):
     This function will remove records pointed by `xmlids` only if they are not referenced
     from any table.
 
+    .. note::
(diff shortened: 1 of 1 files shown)


src/util/records.py
```diff
@@ -1089,7 +1089,7 @@ def delete_unused(cr, *xmlids, **kwargs):
     Remove unused records.
 
     This function will remove records pointed by `xmlids` only if they are not referenced
-    from any table.
+    from any table. The records that cannot be removed are set as `noupdate=True`.
```
KangOl, line 1092:
```suggestion
    from any table.

    .. note:
       The records that cannot be removed are set as `noupdate=True`.
```

aj-fuentes:
@robodoo r+

PR #71, opened 2024-04-22
[FIX] util/pg: always change type with USING for booleans

aj-fuentes:
For the ORM a NULL value in a boolean colum means False. We cannot exclude those values from the type conversion otherwise the assigned value may be wrong.

For example, in this usage:
```py
util.alter_column_type(cr, "table", "col", "varchar", using="CASE WHEN {0} THEN 'x' ELSE 'y' END"
```
we would fail to assign `x`/`y` to the column in null values.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -570,9 +570,12 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
     cr.execute(format_query(cr, "ALTER TABLE {} ADD COLUMN {} {}", table, column, sql.SQL(type)))
 
     using = sql.SQL(format_query(cr, using, tmp_column))
+    where_clause = sql.SQL("")
+    if column_type(cr, table, column) != "bool":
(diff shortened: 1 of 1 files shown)


src/util/pg.py
```diff
@@ -570,9 +570,12 @@ def alter_column_type(cr, table, column, type, using=None, logger=_logger):
     cr.execute(format_query(cr, "ALTER TABLE {} ADD COLUMN {} {}", table, column, sql.SQL(type)))
 
     using = sql.SQL(format_query(cr, using, tmp_column))
+    where_clause = sql.SQL("")
+    if column_type(cr, table, column) == "bool":
```
KangOl, line 574:
```suggestion
    if column_type(cr, table, column) != "bool":
```

aj-fuentes: 🤦🏼‍♂️

KangOl (approved):
@robodoo r+

PR #73, opened 2024-04-23
[IMP] util/pg: remove constraint without table prefix

aj-fuentes:
Constraint by default have the table name as prefix. It's a common error to forget that.

When a table is renamed it's expected that ir_model_constraint may be missing. The ORM may not know about it yet.

See odoo/upgrade#5915

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -642,18 +642,44 @@ def view_exists(cr, view):
     return bool(cr.rowcount)
 
 
-def remove_constraint(cr, table, name, cascade=False):
+def remove_constraint(cr, table, name, cascade=False, warn=True):
+    """
+    Remove a table constraint.
+
+    This function removes the constraint `name` from `table`. It also removes records from
+    `ir_model_constraint` and its xml_ids. It logs not found constraints.
+
+    .. note::
+
+       If there is no constraint `name`, this function will attempt to remove
+       ``{table}_{name}``, the latter is the default name the ORM uses for constraints
+       created from `_sql_constraints`.
+
+    :param str table: table from where to remove the constraint
+    :param str name: name of the constraint to remove
+    :param bool cascade: cascade the constraint removal
+    :param bool warn: use warning level when logging not found constraints, otherwise use
+                      info level
+    :return: whether the constraint was removed
+    :rtype: bool
+    """
     _validate_table(table)
+    log = _logger.warning if warn else _logger.info
     cascade = "CASCADE" if cascade else ""
     cr.execute('ALTER TABLE "{}" DROP CONSTRAINT IF EXISTS "{}" {}'.format(table, name, cascade))
-    # small exception: odoo specific action.
-    # needs to be kept here to avoid resurive imports.
-    # a solution would be to not do it now but adds an `end-` script that remove the invalid entries
-    # from the `ir_model_constraint` table
+    # Exceptionally remove Odoo records, even if we are in PG land on this file. This is somehow
+    # valid because ir.model.constraint are ORM low-level objects that relate directly to table
+    # constraints.
     cr.execute("DELETE FROM ir_model_constraint WHERE name = %s RETURNING id", [name])
     if cr.rowcount:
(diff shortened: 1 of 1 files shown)


src/util/pg.py
```diff
@@ -608,18 +608,45 @@ def view_exists(cr, view):
...
-        cr.execute("DELETE FROM ir_model_data WHERE model = 'ir.model.constraint' AND res_id IN %s", [ids])
+    if not cr.rowcount:
+        if name.startswith(table + "_"):
+            log("%r not found in ir_model_constraint, table=%r", name, table)
+            return False
+        log("%r not found in ir_model_constraint, attempting to remove with table %r prefix", name, table)
+        return remove_constraint(cr, table, "{}_{}".format(table, name), cascade, warn)
+    ids = tuple(c for (c,) in cr.fetchall())
+    cr.execute("DELETE FROM ir_model_data WHERE model = 'ir.model.constraint' AND res_id IN %s", [ids])
+    return True
```
Pirols, line 649:
```suggestion
    if cr.rowcount:
        ids = tuple(c for (c,) in cr.fetchall())
        cr.execute("DELETE FROM ir_model_data WHERE model = 'ir.model.constraint' AND res_id IN %s", [ids])
        return True
    if name.startswith(table + "_"):
        log("%r not found in ir_model_constraint, table=%r", name, table)
        return False
    log("%r not found in ir_model_constraint, attempting to remove with table %r prefix", name, table)
    return remove_constraint(cr, table, "{}_{}".format(table, name), cascade, warn)
```

Pirols: Arguments:

* Avoid changing the blame for the normal case
* No nested `if`s
* `fetchall()` right after query

src/util/pg.py
```diff
@@ -608,18 +608,45 @@ def view_exists(cr, view):
...
+
+    This function removes the constraint `name` from the table named `table`. It also
+    removes records from `ir_model_constraint` and its xmlids. It logs not found
+    constraints.
+
+    .. note::
+
+       If there is no constraint `name`, this function will attempt to remove
+       ``{table}_{name}``, the latter is the default name the ORM uses for constraints
+       created from `_sql_constraints`. This name mismatch will be logged.
```
Pirols, line 623:
```suggestion
       If there is no constraint `name` on `table`, this function will attempt to remove
       ``{table}_{name}``, the latter is the default name the ORM uses for constraints
       created from `_sql_constraints`.
```

I wouldn't add the last ambiguous part to this note. In fact, mismatches are logged iff `name` is not found in the first place.

Pirols (commented):
IMO that's a good change. I commented about a couple of detais

src/util/pg.py
aj-fuentes: Indeed, this code is a remnant of a non-recursive version I had first...

src/util/pg.py
aj-fuentes: I removed it, although it was _not_ ambiguous: name mismatches are always logged, in the form "... not found in ..." ;)

Pirols (approved):
lgtm

KangOl (approved):
@robodoo r+

KangOl:
Hey!
But runbot is red :(

@robodoo r-

KangOl:
upgradeci retry with always only account account_disallowed_expenses account_edi_proxy_client account_intrastat appointment base calendar document documents documents_spreadsheet fleet hr hr_holidays hr_timesheet hr_work_entry l10n_lu_reports l10n_rs mail onboarding project_forecast repair sale sale_project sale_purchase sale_renting sale_subscription sale_timesheet spreadsheet_edition stock survey website_slides website_slides_survey

aj-fuentes:
upgradeci retry

aj-fuentes:
upgradeci retry

aj-fuentes:
upgradeci retry

aj-fuentes:
See https://github.com/odoo/upgrade/pull/5915

KangOl:
@robodoo r+

PR #74, opened 2024-04-24
[FIX] util/domains: add domain field

aj-fuentes:
Since saas~15.3 we have:
* `loyalty.rule` with  `product_domain`
* `loyalty.reward` with `discount_product_domain`

See odoo/odoo@9bb9eaf06726ec6a97c377c1cc2dfa550c1f2be7


KangOl (approved):
@robodoo r+

PR #75, opened 2024-04-25
[REF] base: safe_eval rewrite using AST

ghost:
Context
=======

`safe_eval` is Odoo's sandbox, this mechanism allows users and developpers to write templates, server actions, and more without worrying about the security risks associated with arbitrary code execution.

The current version of the sandbox heavily relies on Python's bytecodes and compile-time verifications. This causes 2 major problems:

1) In every release of Python, its bytecodes are updated or modified.
   Which makes Odoo unusable until the security / framework team updates the
   whitelist of bytecode.

2) Most sandboxing issues we have faced for the last years was due to a
   lack of runtime checks (functions inputs (arguments) and outputs
   (return values)). All most every times those kind of issues were
   fixed with "dirty" hacks such as adding a list of
   "unsafe attribute" or adding a wrapper for modules that are exposing
   unsafe objects (such as the `sys` module)

Goal of the change
==================

During this rewrite we had a few goals:

1) Retain compatibility with the original version:
    * Find a way to keep the old checks (deny dunders, attribute storing
      and deleting)
    * Keep the same exposed API, limiting the amount of code that needs
      to be rewritten as much as possible

2) Add runtime checks to verify that every types passed and returns are
   safe by checking their type. The way that the sandbox does it is by
   using two set of types. One for the types we allow to instanciate
   (the ones that we have absolute trust, most of them are primitive
   types such as `str` and `int`) and the ones that we only allow as
   instance, this means that you CANNOT instanciate them inside of the
   sandbox (for example the sql cursor or the Odoo environement).

3) Eliminate the issues with the `.format` and `.format_map`.
   This is a well known issue within the Python security community, if
   you want more info : https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

--
See odoo/odoo#138611 odoo/enterprise#48919 odoo/upgrade#5653

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -35,7 +35,7 @@ except ImportError:
 from .const import NEARLYWARN
 from .helpers import _dashboard_actions, _validate_model
 from .inherit import for_each_inherit
-from .misc import SelfPrintEvalContext
+from .misc import SelfPrintEvalContext, version_gte
 from .pg import column_exists, get_value_or_en_translation, table_exists
 from .records import edit_view
 
@@ -207,7 +207,11 @@ def _adapt_one_domain(cr, target_model, old, new, model, domain, adapter=None, f
     # pre-check domain
     if isinstance(domain, basestring):
         try:
-            eval_dom = expression.normalize_domain(safe_eval(domain, evaluation_context, nocopy=True))
+            if version_gte("saas~17.2"):
+                eval_dom = expression.normalize_domain(safe_eval(domain, locals_dict=evaluation_context))
+            else:
+                eval_dom = expression.normalize_domain(safe_eval(domain, evaluation_context, nocopy=True))
+
         except Exception as e:
             oops = ustr(e)
             _logger.log(NEARLYWARN, "Cannot evaluate %r domain: %r: %s", model, domain, oops)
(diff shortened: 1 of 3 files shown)


src/util/domains.py
```diff
@@ -207,7 +207,11 @@ def _adapt_one_domain(cr, target_model, old, new, model, domain, adapter=None, f
     # pre-check domain
     if isinstance(domain, basestring):
         try:
-            eval_dom = expression.normalize_domain(safe_eval(domain, evaluation_context, nocopy=True))
+            if version_gte("saas~17.2"):
+                eval_dom = expression.normalize_domain(safe_eval(domain, locals_dict=evaluation_context))
```
KangOl, line 211: Why passing the `evaluation_context` as `locals_dict` in this case?

src/util/fields.py
```diff
@@ -1142,7 +1148,10 @@ def adapt_dict(d):
                     adapt_dict(d[vt])
 
         for _, act in _dashboard_actions(cr, match, *only_models or ()):
-            context = safe_eval(act.get("context", "{}"), eval_context, nocopy=True)
+            if version_gte("saas~17.2"):
+                context = safe_eval(act.get("context", "{}"), eval_context, nocopy=True)
+            else:
+                context = safe_eval(act.get("context", "{}"), eval_context, nocopy=True)
```
KangOl, line 1154: No difference?

src/util/domains.py
joda-odoo: iirc, if you pass it in the globals it will break.

KangOl:
upgradeci retry with always only base hr in 17.0 master

PR #76, opened 2024-04-25
[FIX] util/pg: use the old column to check the type

aj-fuentes:
It's the original column what we need to check, the target column would
have a different type.

See 353c206a3303ccecf9ebedc5c51dd3381bb74ef6

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -431,6 +431,30 @@ class TestIterBrowse(UnitTestCase):
 
 
 class TestPG(UnitTestCase):
+    def test_alter_column_type(self):
+        cr = self.env.cr
+        cr.execute(
+            """
+            ALTER TABLE res_partner_title ADD COLUMN x bool;
+
+            UPDATE res_partner_title
+               SET x = CASE id % 3
+                           WHEN 1 THEN NULL
+                           WHEN 2 THEN True
+                           ELSE False
+                       END
+            """
+        )
+        self.assertEqual(util.column_type(cr, "res_partner_title", "x"), "bool")
+        util.alter_column_type(cr, "res_partner_title", "x", "int", using="CASE {0} WHEN True THEN 2 ELSE 1 END")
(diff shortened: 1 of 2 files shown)


KangOl:

src/base/tests/test_util.py
```diff
@@ -431,6 +431,28 @@ def test_iter_browse_call_twice(self):
...
+
+            UPDATE res_partner_title
+               SET x = CASE id % 3
+                           WHEN 1 THEN NULL
+                           WHEN 2 THEN True
+                           ELSE False
+                       END
+            """
+        )
+        util.alter_column_type(cr, "res_partner_title", "x", "int", using="CASE {0} WHEN True THEN 2 ELSE 1 END")
```
KangOl, line 448:
```suggestion
        util.alter_column_type(cr, "res_partner_title", "x", "int", using="CASE {0} WHEN True THEN 2 ELSE 1 END")
        self.assertEqual(util.column_type(cr, "res_partner_title", "x"), "int")
```

aj-fuentes: done

src/base/tests/test_util.py
```diff
@@ -431,6 +431,30 @@ def test_iter_browse_call_twice(self):
...
+               SET x = CASE id % 3
+                           WHEN 1 THEN NULL
+                           WHEN 2 THEN True
+                           ELSE False
+                       END
+            """
+        )
+        self.assertEqual(util.column_type(cr, "res_partner_title", "x"), "bool")
+        util.alter_column_type(cr, "res_partner_title", "x", "int", using="CASE {0} WHEN True THEN 2 ELSE 1 END")
+        self.assertEqual(util.column_type(cr, "res_partner_title", "x"), "int")
```
KangOl, line 450:
```suggestion
        self.assertEqual(util.column_type(cr, "res_partner_title", "x"), "int4")
```

Sorry for the first bad proposal.

aj-fuentes: done

KangOl:
@robodoo r+

PR #77, opened 2024-04-25
[FIX] base/saas~17.0: do not adapt invalid domains

diagnoza:
Pretty much a copy of https://github.com/odoo/upgrade/pull/5916 since this file was moved to `upgrade-util` since then.

We handle the most common cases and as a last resort we call `convert_domain_leaf()` in order to convert a domain to a valid Python expression. If that fails, we may confidently state that the given domain is invalid, thus we skip its adaptation altogether.


KangOl:
Can you refer to the right version in the commit message?

aj-fuentes (approved):

KangOl:
@robodoo r+

PR #80, opened 2024-05-06
[IMP] util/inconsistencies: allow to further restrict UoMs checks

aj-fuentes:
In some cases we are checking more than what we need to. This leads to blocking more upgrades than we need to.


aj-fuentes:
upgradeci retry with always mrp stock website_sale delivery in versions 13.0 14.0 15.0 16.0 17.0

aj-fuentes:
We'll try a different approach

PR #81, opened 2024-05-06
[IMP] util/records: enforce `cascade` removal for actions

pauloamed:
The implementation of the python inheritance mechanism between the base class `ir.actions.actions` and its child classes (eg. `ir.actions.act_window`) does not allow the creation of foreign keys when `ir.actions.actions` is a M2O field of another model; what leads to the not execution of some constraints, one of them being the `ondelete='cascade'` constraint, which is set in PSQL level.

That said, when a `ir.actions.actions` record is deleted, if it is being referenced as a M2O field by another model (eg. `ir.filters`), records from this second model won't be affected, what leads to undesired behaviour: a MissingError in the UI, indicating that the action was deleted.

Such behaviour of not creating foreign keys and thus constraints is specific to `ir.actions.actions`.
This commit remedies this specific case, removing records with a M2O field to `ir.actions.actions` with `ondelete=cascade` when the action referenced in such field is removed using `upgrade-util`.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -310,6 +310,51 @@ else:
 # fmt:on
 
 
+def remove_action(cr, xml_id=None, action_id=None):
+    assert bool(xml_id) ^ bool(action_id)
+
+    action_model = None
+    if action_id:
+        cr.execute("SELECT type FROM ir_actions WHERE id=%s", [action_id])
+        [action_model] = cr.fetchone()
+    else:
+        module, _, name = xml_id.partition(".")
+        cr.execute("SELECT model, res_id FROM ir_model_data WHERE module=%s AND name=%s", [module, name])
+        action_model, action_id = cr.fetchone()
+        if action_model not in {ihn.model for ihn in for_each_inherit(cr, "ir.actions.actions")}:
+            raise ValueError(
+                "%r should point to a model inheriting from 'ir.actions.actions', not a %r" % (xml_id, action_model)
+            )
+
+    # on_delete value is correct only from version 11, thus can only be used on upgrades to version > 11
+    if not version_gte("12.0"):
+        return remove_records(cr, action_model, [action_id])
+
+    cr.execute(
+        """
+        SELECT name,
+               model,
+               on_delete
+          FROM ir_model_fields
+         WHERE relation = 'ir.actions.actions'
+           AND on_delete IN ( 'cascade', 'set null' )
+           AND ttype = 'many2one'
+        """
+    )
+
+    for column_name, model, on_delete in cr.fetchall():
+        model_table = table_of_model(cr, model)
+        if on_delete == "cascade":
+            query = format_query(cr, "SELECT id FROM {} WHERE {} = %s", model_table, column_name)
+            cr.execute(query, (action_id,))
+            remove_records(cr, model, [id_ for (id_,) in cr.fetchall()])
+        else:
+            query = format_query(cr, "UPDATE {} SET {} = NULL WHERE {} = %s", model_table, column_name, column_name)
+            explode_execute(cr, cr.mogrify(query, [action_id]).decode(), table=model_table)
+
+    return remove_records(cr, action_model, [action_id])
+
+
 def remove_record(cr, name):
     """
     Remove a record and its references corresponding to the given :term:`xml_id <external identifier>`.
@@ -351,6 +396,10 @@ def remove_record(cr, name):
     if model == "res.groups":
         _logger.log(NEARLYWARN, "Removing group %r", name)
         return remove_group(cr, group_id=res_id)
+
+    if model in {ihn.model for ihn in for_each_inherit(cr, "ir.actions.actions")}:
+        _logger.log(NEARLYWARN, "Removing action %r", name)
+        return remove_action(cr, action_id=res_id)
 
     return remove_records(cr, model, [res_id])
 
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1046,6 +1046,37 @@ class TestRecords(UnitTestCase):
         [count] = self.env.cr.fetchone()
         self.assertEqual(count, 1)
 
+    @unittest.skipUnless(util.version_gte("12.0"), "Only work on Odoo >= 12")
+    def test_remove_action_apply_cascade(self):
+        action = self.env["ir.actions.act_url"].create({"name": "act_test", "url": "test.com"})
+        self.env["ir.model.data"].create(
+            {"name": "act_test", "module": "base", "model": "ir.actions.act_url", "res_id": action.id}
+        )
+
+        filter_ = self.env["ir.filters"].create({"name": "filter", "model_id": "res.users", "action_id": action.id})
+
+        util.remove_record(self.env.cr, "base.act_test")
+
+        self.assertIsNone(util.ref(self.env.cr, "base.act_test"))
+        self.assertFalse(action.exists())
+        self.assertFalse(filter_.exists())
+
+    @unittest.skipUnless(util.version_gte("12.0"), "Only work on Odoo >= 12")
+    def test_remove_action_apply_setnull(self):
+        action = self.env["ir.actions.server"].create(
+            {"name": "act_test", "model_id": util.ref(self.env.cr, "base.model_res_users")}
+        )
+
+        user = self.env["res.users"].create({"login": "U1", "name": "U1", "action_id": action.id})
+
+        util.remove_action(self.env.cr, action_id=action.id)
+
+        util.invalidate(user)
+
+        self.assertFalse(action.exists())
+        self.assertTrue(user.exists())
+        self.assertFalse(user.action_id)
+
 
 class TestMisc(UnitTestCase):
     @parametrize(


src/util/records.py
```diff
@@ -309,6 +309,52 @@ def remove_asset(cr, name):
         remove_view(cr, name, silent=True)
 # fmt:on
 
+def get_models_inheriting_actions_actions(cr):
+    # NOTE: it would be nice to have this cached since its static info
+    # + opted for sql querying over hard-coding for better maintainability and handling custom models
+    cr.execute(
+        """
+        SELECT child_model.model
```
aj-fuentes, line 317: Why not performing this check on the query in line 340?

aj-fuentes:
You could move the notes from the code to the PR. Add them as comments here.

src/util/records.py
```diff
@@ -351,10 +397,13 @@ def remove_record(cr, name):
     if model == "res.groups":
         _logger.log(NEARLYWARN, "Removing group %r", name)
         return remove_group(cr, group_id=res_id)
+    
+    if model in get_models_inheriting_actions_actions(cr):
```
aj-fuentes, line 401: We could have a hard-coded list of standard action models. We don't need to take into account all possible models extending the actions base model. We don't that for views for example.

KangOl: We can just use `for_each_inherit`

src/util/records.py
aj-fuentes: I meant that you can perform this check
https://github.com/odoo/upgrade-util/pull/81/files#diff-bdae1b00c18f37288ef023e7125a477b10e973d73dbedecc794bdb729e2daa75R336
directly in the query before, but `for_each_inherit` would also work in terms of providing a pre-computed set of models.

src/util/records.py
aj-fuentes: The SQL query version plus the query itself adds complexity, it's cleaner with `for_each_inherit`, IMO.

src/util/records.py
```diff
@@ -309,6 +309,43 @@ def remove_asset(cr, name):
...
+    if action_id:
+        module, _, name = xml_id.partition(".")
+        cr.execute("SELECT model FROM ir_model_data WHERE module=%s AND name=%s", [module, name])
+        [model] = cr.fetchone()
+        if not model in get_models_inheriting_actions_actions(cr):
+            raise ValueError(
+                "%r should point to a model inheriting from 'ir.actions.actions', not a %r" % (xml_id, model)
+            )
+    
+    cr.execute("SELECT name, model FROM ir_model_fields WHERE relation = 'ir.actions.actions' AND on_delete ='cascade'")
```
KangOl, line 341: You should also search the one linking to the xmlid's model
```suggestion
    cr.execute("SELECT name, model FROM ir_model_fields WHERE relation IN %s AND on_delete ='cascade'", [tuple({model, "ir.actions.actions"})])
```

src/util/records.py
```diff
@@ -309,6 +309,43 @@ def remove_asset(cr, name):
...
+    for column_name, model in cr.fetchall():
+        model_table = table_of_model(cr, model)
+        cr.execute(
+            'DELETE FROM "{}" WHERE "{}" = %s'.format(model_table, column_name),
```
KangOl, line 345: use `util.format_query`

src/util/records.py
KangOl, line 341: You should also handle the fields having `on_delete="set null"`.

KangOl, line 341: Is the `on_delete` column defined in all versions?

KangOl (changes requested):

src/util/records.py
```diff
@@ -309,6 +309,54 @@ def remove_asset(cr, name):
...
+            )
+    
+    cr.execute(
+        """
+        SELECT name,
+               model,
+               on_delete
+          FROM ir_model_fields
+         WHERE relation = 'ir.actions.actions'
+           AND on_delete IN ( 'cascade', 'set null' )
```
KangOl, line 348:
```suggestion
           AND on_delete IN ( 'cascade', 'set null' )
           AND ttype = 'many2one'
```

src/util/records.py
KangOl: However, it wasn't filled correctly until v11.0: https://github.com/odoo/odoo/commit/4945495401e0bac0a7d19d8bf9e70bed02fa091d

KangOl: We could guess the value based on the `NOT NULL` state of the column.

src/util/records.py
```diff
@@ -310,6 +311,69 @@ def remove_asset(cr, name):
...
 
+def remove_action(cr, xml_id=None, action_id=None):
+    assert bool(xml_id) ^ bool(action_id)
+
+    if xml_id:
+        action_id = ref(cr, xml_id)
+
+    action_model = None
+    if action_id:
+        module, _, name = xml_id.partition(".")
```
KangOl, line 322: This cannot work when the xmlid is not provided.

You should get the model from the `type` column of `ir_actions` table.

src/util/records.py
```diff
@@ -310,6 +311,69 @@ def remove_asset(cr, name):
...
+        cr.execute("SELECT model FROM ir_model_data WHERE module=%s AND name=%s", [module, name])
+        [action_model] = cr.fetchone()
+        if action_model not in [ihn.model for ihn in for_each_inherit(cr, "ir.actions.actions")]:
+            raise ValueError(
+                "%r should point to a model inheriting from 'ir.actions.actions', not a %r" % (xml_id, action_model)
+            )
+
+    if version_gte("11.0"):
```
KangOl, line 330: The `if` can be moved into the for loop.
```python
trust_imf = version_gte("12.0")  # as the value is only correct in v11, we cannot use it when upgrading *to* v11
cr.execute("SELECT name, model, on_delete FROM ...")
for col, model, on_delete in cr.fetchall():
    if not trust_imf:
       on_delete = "set null" if column_nullable(cr, table, col) else "cascade"
    if on_delete == "cascade":
       ...
    else:
       ...
```

KangOl: It is in the schema. But the value in not trustable.

src/util/records.py
```diff
@@ -310,6 +311,70 @@ def remove_asset(cr, name):
...
+        )
+
+        for column_name, model, on_delete in cr.fetchall():
+            model_table = table_of_model(cr, model)
+            if on_delete == "cascade":
+                query = format_query(cr, "DELETE FROM {} WHERE {} = %s", model_table, column_name)
+            else:
+                query = format_query(cr, "UPDATE {} SET {} = NULL WHERE {} = %s", model_table, column_name, column_name)
+            cr.execute(query, (action_id,))
+    else:
```
aj-fuentes, line 351: We could do nothing instead.

aj-fuentes: If it hasn't in all these years... we don't need to worry much now. That being said if we some case we can revisit this.

src/util/records.py
KangOl: Yes. In this case, the filtering on `on_delete` should be removed.

src/util/records.py
```diff
@@ -310,6 +310,51 @@ def remove_asset(cr, name):
...
+
+    for column_name, model, on_delete in cr.fetchall():
+        model_table = table_of_model(cr, model)
+        if on_delete == "cascade":
+            query = format_query(cr, "SELECT id FROM {} WHERE {} = %s", model_table, column_name)
+            cr.execute(query, (action_id,))
+            remove_records(cr, model, [id_ for (id_,) in cr.fetchall()])
+        else:
+            query = format_query(cr, "UPDATE {} SET {} = NULL WHERE {} = %s", model_table, column_name, column_name)
+            cr.execute(query, (action_id,))
```
KangOl, line 353:
```suggestion
            explode_execute(cr, cr.mogrify(query, [action_id]).decode(), table=model_table)
```

KangOl:
> Hey I believe we can revert the following PR [odoo/upgrade#5869](https://github.com/odoo/upgrade/pull/5869) _before_ this ~once~ is merged

No, this is still needed for actions not deleted via this new helper.

erl-odoo:
Any news here? I've discovered another issue with filters on an upgrade project, though it's not related to upgrading technically -> install a module, like hr_holidays, save a filter, uninstall the module = the filter is now corrupt because of the missing model in registry:
image

IMO, apart from this util helper, either:
1 - have a `base` script in `upgrade` that fixes the filters on upgrade, removing ones that have missing actions like the example in this PR and were **NOT deleted with this helper**, and **removes also ones with missing model like I describe above**

2 - I fix this in a PR to `odoo` to cleanup the filters on module / model removal

wdyt? @aj-fuentes @KangOl

aj-fuentes:
@erl-odoo The uninstall process of standard Odoo is full of corner cases where data is left in a wrong state, this is yet another example :)

IMO yes you can try to fix the standard process to cleanup those filters in master odoo. During upgrades, removing what seems to be broken filters may be acceptable, but we usually try to avoid removing data (they may work in case the model re-appears after a reinstall). Generally speaking, if it was broken _before_ the upgrade and doesn't fail the upgrade, we don't fix it.

aj-fuentes:
opw-3931090

PR #83, opened 2024-05-10
[FIX] attr_domains2expr: missing parameter and tracestack if view doesn't have inherit_id

duong77476-viindoo:
-This commit add missing 'cr' parameter to util.ref method and fix when getting combined arch for inherit view it might get stace track if that view doesn't have inherit_id, check at
https://github.com/odoo/odoo/blob/86fb725c03171b84287812d0a9d1e00c86044acd/odoo/addons/base/models/ir_ui_view.py#L628

### DIFF
--- a/src/base/17.0.1.3/attr_domains2expr.py
+++ b/src/base/17.0.1.3/attr_domains2expr.py
@@ -58,14 +58,14 @@ def adapt_view(cr, view_xmlid):
     IrUiView = util.env(cr)["ir.ui.view"].with_context(lang=lang)
     ```
     """
-    vid = util.ref(view_xmlid)
+    vid = util.ref(cr, view_xmlid)
     IrUiView = util.env(cr)["ir.ui.view"]
     view = IrUiView.browse(vid)
 
     # disable view to avoid it being applied to the parent arch
     view.active = False
     # get combined arch of the parent view
-    comb_arch = view.inherit_id._get_combined_arch()
+    comb_arch = view.inherit_id._get_combined_arch() if view.inherit_id else None
 
     # update current view arch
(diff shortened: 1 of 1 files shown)


duong77476-viindoo:
Hi @aj-fuentes can you take a look cc @KangOl

aj-fuentes (approved):
LGTM, thanks!

src/base/17.0.1.3/attr_domains2expr.py
```diff
@@ -58,14 +58,14 @@ def adapt_view(cr, view_xmlid):
...
-    vid = util.ref(view_xmlid)
+    vid = util.ref(cr, view_xmlid)
     IrUiView = util.env(cr)["ir.ui.view"]
     view = IrUiView.browse(vid)
 
     # disable view to avoid it being applied to the parent arch
     view.active = False
     # get combined arch of the parent view
-    comb_arch = view.inherit_id._get_combined_arch()
+    comb_arch = view.inherit_id._get_combined_arch() if view.inherit_id else False
```
aj-fuentes, line 68: Actually from the usage -- see [line 145](https://github.com/odoo/upgrade-util/blob/de2b64766edcf38294a840f143ebaabe0a446014/src/base/17.0.1.3/attr_domains2expr.py#L145) -- we should use `None` instead of `False` here.
```suggestion
    comb_arch = view.inherit_id._get_combined_arch() if view.inherit_id else None
```

duong77476-viindoo: Oki dukey Done

aj-fuentes:
@robodoo r+

PR #87, opened 2024-05-17
[IMP] util.edit_view: better default value for `active`

KangOl:
Only active views by default when selected by xmlid. Views edited by id are most of the time mass-edited to adapt css classes or widget options and should not be blindly activated.

Also add tests.


aj-fuentes (approved):
This is really needed.

KangOl:
@robodoo r+

PR #88, opened 2024-05-23
[IMP] util/domains.py: handle `any` operator introduced in Odoo 17

KangOl:
In such domain, the right-hand term is domain that should also be adatped.

See https://github.com/odoo/odoo/commit/5a998694a6f353da05d7f69e4c59b9e7dd139e27


KangOl:

aj-fuentes (approved):
LGTM

KangOl:
@robodoo r+

PR #91, opened 2024-05-28
[IMP] specific: check if model exists

Pirols:
When a custom model conflicts with a standard one we may choose to write a specific fix, using the `rename_custom_model` util to rename it.

However, although unlikely, customers may remove that model in successive upgrade requests, making the call to `rename_model` unnecessary and the extra entry in the migration report unnecessary.


Pirols:
> It may be a good idea to do it for all `rename_custom_` helpers.

Done!

KangOl:
upgradeci skip
@robodoo rebase-ff r+

PR #92, opened 2024-05-29
[MOV] spreadsheet: move spreadsheet utils to util folder

LucasLefevre:
`upgrade-util/src/spreadsheet` clashes with `upgrade/spreadsheet`. Both are merged in `odoo.upgrade.spreadsheet` from the python point of view.

When importing `from odoo.upgrade.util.spreadsheet`, python follows the path and stops at the first package package (first folder with `__init__.py`)

Currently, if you add tests in `upgrade/spreadsheet/tests`, they are ignored (because of `upgrade-util/spreadsheet/__init__.py`). If you remove the `__init__.py` file, it works.

With this commit, we move all spreadsheet utils (more are coming) into the `util` folder, which allows to safely have an `__init__.py` file.

Tests are kept in `upgrade-util/spreadsheet`, but without any `__init__.py` file.

Note that moving those files is a breaking change and it's not retro-compatible. However, those spreadsheet utils have only been merged a few weeks ago. They are most probably not used outside our internal upgrade repo.

See odoo/upgrade#6086


LucasLefevre:
(`upgradeci/matt` is red pre-16.0 because `spreadsheet` module didn't exist.

KangOl:
upgradeci retry with always only spreadsheet* in 16.0 17.0 master

KangOl:
upgradeci retry

KangOl:
upgradeci retry with always only spreadsheet* in 17.0 master

KangOl:
@robodoo r+

PR #85, opened 2024-05-15
[IMP] saas~17.3: remove doall/numbercall from ir_cron

fdamhaut:
- Remove multiple calls and number of calls mechanism.

COM: odoo/odoo#151628
ENT: odoo/enterprise#58157
UP: https://github.com/odoo/upgrade/pull/5786
task-id: 3687442

PR #93, opened 2024-06-03
[FIX] util.get_admin_channel: get the standard channel if exists.

KangOl:
Complement of odoo/upgrade-util#85

See odoo/upgrade#6105


aj-fuentes (approved):

KangOl:

aj-fuentes:
@robodoo r+

PR #95, opened 2024-06-11
[IMP] base version: lower from warning to info

clbr-odoo:
In Odoofin, we don't have a master branch, we always update current stable version (17.0 at the moment) through module migration scripts. This was previously handled by internal at each major version update. We now handle the scripts in the accounting team, thus have an upgrade CI.

The warning modified in this PR seems to be legitimally raised, since base is not changed, it is probably skipped in the update even if we do the upgrade with `-u all`. The warning level is preventing us to merge without admin rights, thus we propose to lower the warning to info level.

Example of build with this warning: https://runbot.odoo.com/runbot/build/63601210

### DIFF
--- a/src/util/inherit.py
+++ b/src/util/inherit.py
@@ -3,7 +3,7 @@ import logging
 import operator
 import os
 
-from .const import ENVIRON
+from .const import ENVIRON, NEARLYWARN
 from .misc import _cached, parse_version, version_gte
 
 _logger = logging.getLogger(__name__)
@@ -63,7 +63,8 @@ def _get_base_version(cr):
         state, version = cr.fetchone()
         if state != "to upgrade":
             major = ".".join(version.split(".")[:2])
-            _logger.warning(
+            _logger.log(
+                NEARLYWARN,
                 "Assuming upgrading from Odoo %s. If it's not the case, specify the environment variable `ODOO_BASE_VERSION`.",
                 major,
             )


clbr-odoo:
@KangOl Hello :wave: 
I already discussed this problem a bit with Alvaro, one of his possible solution was to manually set the environment variable `ODOO_BASE_VERSION` on each migration script, but this is not very handy.
Hopefully this is an acceptable solution :slightly_smiling_face:

src/util/inherit.py
```diff
@@ -63,7 +63,7 @@ def _get_base_version(cr):
         state, version = cr.fetchone()
         if state != "to upgrade":
             major = ".".join(version.split(".")[:2])
-            _logger.warning(
+            _logger.info(
```
aj-fuentes, line 66: We could make almost a warn in CI.
```suggestion
            _logger.log(
                util.NEARLYWARN if util.on_CI() else logging.WARNING,
```

aj-fuentes:
For info: this also blocks SH custom code upgrade -- it was reported already in some cases. I'd say we should keep it in NEARLYWARN always the noise in SH upgrade tool. @KangOl

src/util/inherit.py
```diff
@@ -6,6 +6,8 @@
 from .const import ENVIRON
 from .misc import _cached, parse_version, version_gte
 
+from odoo.upgrade import util
```
KangOl, line 9:
```suggestion
from .const import ENVIRON, NEARLYWARN
from .misc import _cached, parse_version, version_gte
```

src/util/inherit.py
```diff
@@ -63,7 +65,8 @@ def _get_base_version(cr):
         state, version = cr.fetchone()
         if state != "to upgrade":
             major = ".".join(version.split(".")[:2])
-            _logger.warning(
+            _logger.log(
+                util.NEARLYWARN if util.on_CI() else logging.WARNING,
```
KangOl, line 69:
```suggestion
               NEARLYWARN,
```

KangOl (changes requested):

clbr-odoo:
@KangOl Made the changes, thank you for the review !

KangOl:

KangOl (approved):
@robodoo r+

PR #98, opened 2024-06-18
[FIX] util/records: fix `delete_unused` DOCSTRING

Pirols:
`delete_unused`, in spite of its name, deletes records even if they are used, provided they are not part of a foreign key with `restrict` ondelete clause.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1098,10 +1098,13 @@ def __update_record_from_xml(
 
 def delete_unused(cr, *xmlids, **kwargs):
     """
-    Remove unused records.
+    Remove unrestricted records.
 
-    This function will remove records pointed by `xmlids` only if they are not referenced
-    from any table.
+    This function will remove records, pointed by `xmlids`, only if they are NOT referenced
+    from any table with `restrict` or `set` ondelete clauses.
+
+    .. warning::
+       Indirect references are not checked and can cause the util to fail.
 
     .. note::
        The records that cannot be removed are set as `noupdate=True`.


Pirols:
upgradeci skip

UemuS (commented):
We could also mention that it does not handle indirect references.
If table 1 is referenced with ON DELETE CASCADE by table 2 and table 2 is restricted by table 3, the upgrade would crash.
I'm trying to fix this in addition to some other cases, but it's a bit tricky

src/util/records.py
```diff
@@ -1098,10 +1098,10 @@ def add_ref(ref):
 
 def delete_unused(cr, *xmlids, **kwargs):
     """
-    Remove unused records.
+    Remove unrestricted records.
 
-    This function will remove records pointed by `xmlids` only if they are not referenced
-    from any table.
+    This function will remove records pointed by `xmlids` only if they are NOT referenced
+    with `restrict` ondelete clause from any table.
```
jjmaksoud, line 1105: The way I see the definition of unused here, is that records are not referenced from any table or only referenced with `on delete cascade`, or have a reference from the same model that is not `on delete restrict`. 

So the set null is the same as restrict, only cascade is ignored when checking fk on other tables.

diagnoza:
I thought that it's somewhat implied that if there's no direct link in the database, the record won't be removed :p The term "used" is used in the sense that there is a constraint linking a record to another.

Imho what would be more important to mention is what Yazid said.

Pirols:
> We could also mention that it does not handle indirect references. If table 1 is referenced with ON DELETE CASCADE by table 2 and table 2 is restricted by table 3, the upgrade would crash.

That's a good idea, I will add a warning.

> So the set null is the same as restrict, only cascade is ignored when checking fk on other tables.

This is right and I will clarify it in the DOCSTRING. 

However, I think we should consider addressing this: the behavior sounds to be inconsistent when we have a `set` clause vs a `restrict` one on a direct/indirect reference. See the possibilities:

* Direct `restrict` reference -> no removal :white_check_mark: 
* Direct `set` reference -> no removal :white_check_mark: 
* Indirect `restrict` reference -> failure. Not ideal but at least it raises a concern :warning: 
* Indirect `set` reference -> (my untested guess) removes the records (and triggers the indirect ondelete clause). Inconsistent case :x: 

But well if @UemuS manages to solve the indirect issue, this inconsistency will also be gone, just wanted to acknowledge it.

> I thought that it's somewhat implied that if there's no direct link in the database, the record won't be removed :p

Exactly! And that's the problem with the name/DOCSTRING: they don't explain that records will still be removed if their references can cascade.

src/util/records.py
Pirols: https://github.com/odoo/upgrade-util/pull/98#issuecomment-2176405285

Pirols:
upgradeci skip

PR #99, opened 2024-06-19
[IMP] core: odoo.domains

kmagusiak:
Add the new module in util.

odoo/odoo#170009

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -24,11 +24,9 @@ except ImportError:
     unescape = lambda x: x
 
 try:
-    from odoo.osv import expression
     from odoo.tools import exception_to_unicode
     from odoo.tools.safe_eval import safe_eval
 except ImportError:
-    from openerp.osv import expression
     from openerp.tools import exception_to_unicode
     from openerp.tools.safe_eval import safe_eval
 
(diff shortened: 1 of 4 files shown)


KangOl:
Please [refer](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests) to the odoo/enterprise/design-themes PR (if applicable) in the **commit message**.
It's needed by [UpgradeCI](https://github.com/odoo/upgrade/wiki/CI:-upgradeci) to correctly build this PR with the corresponding branches.

If you don't want to have notifications on the related PR each time you force push, simply don't force push and add new [fixup commits](https://git-scm.com/docs/git-commit/2.32.0#Documentation/git-commit.txt---fixupamendrewordltcommitgt) that will be squashed before the r+.

Note that UpgradeCI will also extract the info from *the PR body*. It is however better to have those references in the commit message itself.

<sub><i>This message is a canned response and thus may not be totally accurate.</i></sub>

KangOl:
upgradeci retry with always only product

src/util/domains.py
```diff
@@ -45,6 +43,91 @@
...
+                expected -= 1
+                if len(token) == 3 and token[1] in ('any', 'not any'):
+                    token = (token[0], token[1], normalize_domain(token[2]))
+                else:
+                    token = normalize_leaf(token)
+            else:
+                expected += op_arity.get(token, 0) - 1
+            result.append(token)
+        if expected:
+            raise ValueError(f'Domain {domain} is syntactically not correct.')
```
KangOl, line 84: This file should be syntactically valid in pyhhon2.
```suggestion
            raise ValueError("Domain {!r} is syntactically not correct.".format(domain))
```

KangOl:
upgradeci retry with always only helpdesk

src/util/domains.py
```diff
@@ -45,6 +43,95 @@
...
+    NOT_OPERATOR = _dom.DomainNot.OPERATOR
+    AND_OPERATOR = _dom.DomainAnd.OPERATOR
+    OR_OPERATOR = _dom.DomainOr.OPERATOR
+    DOMAIN_OPERATORS = {NOT_OPERATOR, AND_OPERATOR, OR_OPERATOR}
+
+    # normalization functions are redefined here because they will be deprecated
+    # the Domain factory normalized but also distributes operators during creation of domains
+
+    def normalize_domain(domain):
+        """Return a normalized version of the domain.
```
aj-fuentes, line 61: Why don't we implement this in terms of the new `Domain` factory? 
`return list(_dom.Domain(domain))` would this be enough?
If not, I still believe it's better to use the _new_ implementation, it will bring better compatibility for future versions.

kmagusiak: it is not strictly equivalent and some tests were failing.
Here is an example from an older build: https://runbot.odoo.com/runbot/build/64886163
One of the differences is stated above: `list(Domain(xxx))` is `distribute_not(normalize_domain(xxx))`.

I suggest that we merge as-is, and I can implement a follow-up PR with the simplification.

src/util/domains.py
```diff
@@ -45,6 +43,95 @@
...
+                    result.append(normalize_leaf(token))
+            else:
+                expected += op_arity.get(token, 0) - 1
+                result.append(token)
+        if expected:
+            raise ValueError("Domain {!r} is syntactically not correct.".format(domain))
+        return result
+
+    def normalize_leaf(leaf):
+        if not is_leaf(leaf):
```
aj-fuentes, line 94: If we allow in the linked PR in community to register optimizations that work accross all models we could do here:
`list(domains.Domain("a","<>",1)._optimize(None))`
which I can only seem to get to work with
```
list(domains.Domain("valid_field","<>",1)._optimize(self.env['base']))
```

src/util/domains.py
```diff
@@ -45,6 +43,95 @@
...
+            raise TypeError("Leaf must be a tuple or list of 3 values")
+        left, operator, right = leaf
+        original = operator
+        operator = operator.lower()
+        if operator == "<>":
+            operator = "!="
+        if isinstance(right, bool) and operator in ("in", "not in"):
+            _logger.warning("The domain term '%s' should use the '=' or '!=' operator.", ((left, original, right),))
```
aj-fuentes, line 102: Is this still true?
```py
>>> list(domains.Domain("id","in",True)._optimize(self.env['base']))
[('id', 'in', [True])]
>>> list(domains.Domain("active","in",True)._optimize(self.env['res.users']))
[('active', 'in', [True])]
>>> list(domains.Domain("active","in",False)._optimize(self.env['res.users']))
[(0, '=', 1)]
```
The last output looks odd... I'd expect an error or `[('active','in',[False])]`/`[('active','=',False)]` as output.

If we no longer show any warning then why would we show one during the upgrade. If we now silently replace the leaf by something else, the warnings should be now more like `This is translated to ... since Odoo 19`.

I still think it's better to implement `normalize_leaf` using the new machinery at least for the cases where it doesn't do anything too involved.

src/util/domains.py
aj-fuentes: That failure doesn't make sense 🤯  (irrelevantly of the current implementation) it's a test that should only run when going over 17.3. Why is it trying to upgrade from 17 to master, that's not valid. Only >=18 can be upgraded to master. Checking with runbot team...

src/util/domains.py
kmagusiak: I don't think you should optimize domains during the migration, it does more than simple rewrites.
That operator is replaced automatically for a long time and continues to work for now.

Anyways, same remark as above.

src/util/domains.py
aj-fuentes: > The implementation is a copy of what is done today.

That's what I understood. My point: we cannot keep for the future what's done today. It should be aligned. For merging the task this may be OK, but for the long term whatever we do here should be aligned with what's expected from the implementation in community.

src/util/domains.py
kmagusiak: @aj-fuentes Added to my TODO list

src/util/domains.py
rco-odoo: Well spotted.

src/util/domains.py
rco-odoo: `self.env['base']` is a _particular_ model.
You need a different sentinel value for "any possible model".

aj-fuentes: > You need a different sentinel value for "any possible model".

Indeed, to me it should be `None`. I was just hacking my way to get a feeling how this now works :)

src/util/domains.py
aj-fuentes: Note that the last runbot run upgraded 18->master. Thus the tests that failed before, shouldn't anymore.

kmagusiak:
@robodoo r+

PR #100, opened 2024-06-25
[ADD] util.fields: handle base_import.mapping model/fields renames/removal

pauloamed:
Implement proper fixing and/or removal of base_import.mapping and records on model/fields renames/removal:
- renamed fields: update affected base_import.mapping records
- removed fields: remove affected base_import.mapping records
- renamed models: update affected base_import.mapping records `res_model`
- removed models: remove affected base_import.mapping records

This is a follow-up on 5b944f7cad5512d90938a28b6db93efb9231a5d8

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -40,13 +40,7 @@ except ImportError:
 from .const import ENVIRON
 from .domains import _adapt_one_domain, _replace_path, _valid_path_to, adapt_domains
 from .exceptions import SleepyDeveloperError
-from .helpers import (
-    _dashboard_actions,
-    _remove_export_lines,
-    _validate_model,
-    resolve_model_fields_path,
-    table_of_model,
-)
+from .helpers import _dashboard_actions, _validate_model, resolve_model_fields_path, table_of_model
 from .inherit import for_each_inherit
 from .misc import SelfPrintEvalContext, log_progress, version_gte
 from .orm import env, invalidate
@@ -63,6 +57,7 @@ from .pg import (
     savepoint,
     table_exists,
 )
+from .records import _remove_import_export_paths
 from .report import add_to_migration_reports, get_anchor_link_to_record
 
 # python3 shims
@@ -208,8 +203,7 @@ def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inh
             [(fieldname, fieldname + " desc"), model, r"\y{}\y".format(fieldname)],
         )
 
-    # ir.exports.line
-    _remove_export_lines(cr, model, fieldname)
+    _remove_import_export_paths(cr, model, fieldname)
 
     def adapter(leaf, is_or, negated):
         # replace by TRUE_LEAF, unless negated or in a OR operation but not negated
@@ -969,6 +963,54 @@ def update_field_references(cr, old, new, only_models=None, domain_adapter=None,
     return _update_field_usage_multi(cr, models, old, new, domain_adapter=domain_adapter, skip_inherit=skip_inherit)
 
 
+def _update_impex_renamed_fields_paths(cr, old_field_name, new_field_name, only_models):
+    export_q = cr.mogrify(
+        """
+        SELECT el.id,
+               e.resource,
+               STRING_TO_ARRAY(el.name, '/')
+          FROM ir_exports_line el
+          JOIN ir_exports e
+            ON el.export_id = e.id
+         WHERE el.name ~ %s
+        """,
+        [r"\y{}\y".format(old_field_name)],
+    ).decode()
+    impex_data = [(export_q, "ir_exports_line", "name")]
+    if table_exists(cr, "base_import_mapping"):
+        import_q = cr.mogrify(
+            """
+            SELECT id,
+                   res_model,
+                   STRING_TO_ARRAY(field_name, '/')
+              FROM base_import_mapping
+             WHERE field_name ~ %s
+            """,
+            [r"\y{}\y".format(old_field_name)],
+        ).decode()
+        impex_data.append((import_q, "base_import_mapping", "field_name"))
+
+    for query, table, column in impex_data:
+        cr.execute(query)
+        if not cr.rowcount:
+            continue
+        fixed_paths = {}
+        for record_id, related_model, path in cr.fetchall():
+            new_path = [
+                new_field_name
+                if field.field_name == old_field_name and field.field_model in only_models
+                else field.field_name
+                for field in resolve_model_fields_path(cr, related_model, path)
+            ]
+            if len(new_path) == len(path) and new_path != path:
+                fixed_paths[record_id] = "/".join(new_path)
+        if fixed_paths:
+            cr.execute(
+                format_query(cr, "UPDATE {} SET {} = (%s::jsonb)->>(id::text) WHERE id IN %s", table, column),
+                [Json(fixed_paths), tuple(fixed_paths)],
+            )
+
+
 def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_inherit=()):
     assert models
     only_models = None if models == "*" else tuple(models)
@@ -1079,33 +1121,9 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
         """
         cr.execute(q.format(col_prefix=col_prefix), p)
 
-        # ir.exports.line
+        # ir.exports.line, base_import.mapping # noqa
         if only_models:
-            cr.execute(
-                """
-                SELECT el.id,
-                       e.resource,
-                       STRING_TO_ARRAY(el.name, '/')
-                  FROM ir_exports_line el
-                  JOIN ir_exports e
-                    ON el.export_id = e.id
-                 WHERE el.name ~ %s
-                """,
-                [r"\y{}\y".format(old)],
-            )
-            fixed_lines_paths = {}
-            for line_id, line_model, line_path in cr.fetchall():
-                new_path = [
-                    new if x.field_name == old and x.field_model in only_models else x.field_name
-                    for x in resolve_model_fields_path(cr, line_model, line_path)
-                ]
-                if len(new_path) == len(line_path) and new_path != line_path:
-                    fixed_lines_paths[line_id] = "/".join(new_path)
-            if fixed_lines_paths:
-                cr.execute(
-                    "UPDATE ir_exports_line SET name = (%s::jsonb)->>(id::text) WHERE id IN %s",
-                    [Json(fixed_lines_paths), tuple(fixed_lines_paths)],
-                )
+            _update_impex_renamed_fields_paths(cr, old, new, only_models)
 
         # mail.alias
         if column_exists(cr, "mail_alias", "alias_defaults"):
--- a/src/util/helpers.py
+++ b/src/util/helpers.py
@@ -315,27 +315,3 @@ def resolve_model_fields_path(cr, model, path):
         {"model": model, "path": list(path)},
     )
     return [FieldsPathPart(**row) for row in cr.dictfetchall()]
-
-
-def _remove_export_lines(cr, model, field=None):
-    q = """
-    SELECT el.id,
-           e.resource,
-           STRING_TO_ARRAY(el.name, '/')
-      FROM ir_exports_line el
-      JOIN ir_exports e
-        ON el.export_id = e.id
-    """
-    if field:
-        q = cr.mogrify(q + " WHERE el.name ~ %s ", [r"\y{}\y".format(field)]).decode()
-    cr.execute(q)
-    to_rem = [
-        line_id
-        for line_id, line_model, line_path in cr.fetchall()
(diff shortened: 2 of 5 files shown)


aj-fuentes (commented):
Please do not over document the code. It's clear from the code itself, or the commit message, what it does. Comments tend to go out of sync with code later on, and they produce bigger diffs.

KangOl:
upgradeci retry with always only base_import+hr

src/util/helpers.py
```diff
@@ -339,3 +339,25 @@ def _remove_export_lines(cr, model, field=None):
...
+    to_rem = [
+        mapping_id
+        for mapping_id, mapping_model, mapping_path in cr.fetchall()
+        if any(
+            x.field_model == model and (field is None or x.field_name == field)
+            for x in resolve_model_fields_path(cr, mapping_model, mapping_path)
+        )
+    ]
+    if to_rem:
+        cr.execute("DELETE FROM base_import_mapping WHERE id IN %s", [tuple(to_rem)])
```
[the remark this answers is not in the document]
aj-fuentes: This is done in a different new private function, it could be refactored later. I was referring to the extra lines in the main processing logic in `field.py`

aj-fuentes (commented):
I find it clearer if the inner function just _executes_ a query without knowing about parameters.

It also makes the inner function relocatable --not a closure.

src/util/helpers.py
```diff
@@ -339,3 +339,25 @@ def _remove_export_lines(cr, model, field=None):
     ]
     if to_rem:
         cr.execute("DELETE FROM ir_exports_line WHERE id IN %s", [tuple(to_rem)])
+
+
+def _remove_import_mappings(cr, model, field=None):
```
aj-fuentes, line 344: Following @andreabak suggestion I think you can just merge these two into one `_remove_import_export_lines`

src/util/helpers.py
```diff
@@ -317,8 +317,22 @@ def resolve_model_fields_path(cr, model, path):
...
+    to_rem = [
+        path_id
+        for path_id, related_model, path in cr.fetchall()
+        if any(
+            x.field_model == model and (field is None or x.field_name == field)
+            for x in resolve_model_fields_path(cr, related_model, path)
+        )
+    ]
+    if to_rem:
+        cr.execute(delete_query, [tuple(to_rem)])
```
aj-fuentes, line 331: I still believe you don't need more than one function:
```py
def _remove_import_export_paths(cr, model, field=None):
        export_q = """
            SELECT el.id,
                   e.resource,
                   STRING_TO_ARRAY(el.name, '/')
              FROM ir_exports_line el
              JOIN ir_exports e
                ON el.export_id = e.id
        """
        if field:
            export_q = cr.mogrify(export_q + " WHERE el.name ~ %s ", [r"\y{}\y".format(field)]).decode()

        import_q = """
        SELECT id,
               res_model,
               STRING_TO_ARRAY(field_name, '/')
          FROM base_import_mapping
        """
        if field:
            import_q = cr.mogrify(import_q + " WHERE field_name ~ %s ", [r"\y{}\y".format(field)]).decode()

        for query, table in [(export_q, "ir_exports_line"), (import_q, "base_import_mapping")]):
            cr.execute(query)
            to_rem = [
                path_id
                for path_id, related_model, path in cr.fetchall()
                if any(
                    x.field_model == model and (field is None or x.field_name == field)
                    for x in resolve_model_fields_path(cr, related_model, path)
                )
            ]
            if to_rem:
                cr.execute("DELETE FROM {} WHERE id IN %s".format(table), [tuple(to_rem)])
```

src/util/fields.py
```diff
@@ -1092,20 +1112,27 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
-                if len(new_path) == len(line_path) and new_path != line_path:
-                    fixed_lines_paths[line_id] = "/".join(new_path)
-            if fixed_lines_paths:
-                cr.execute(
-                    "UPDATE ir_exports_line SET name = (%s::jsonb)->>(id::text) WHERE id IN %s",
-                    [Json(fixed_lines_paths), tuple(fixed_lines_paths)],
-                )
+            ).decode()
+
+            update_query = "UPDATE ir_exports_line SET name = (%s::jsonb)->>(id::text) WHERE id IN %s"
```
aj-fuentes, line 1117: Instead of passing the update_query, why not just getting the fixed_paths from the inner function and do the update outside, here directly.

src/util/helpers.py
aj-fuentes: I don't see how we would cleanup exports but not imports.

src/util/fields.py
```diff
@@ -1079,9 +1081,27 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
+                        new
+                        if field.field_name == old_field_name and field.field_model in only_models
+                        else field.field_name
+                        for field in resolve_model_fields_path(cr, related_model, path)
+                    ]
+                    if len(new_path) == len(path) and new_path != path:
+                        fixed_paths[record_id] = "/".join(new_path)
+
+                if fixed_paths:
+                    cr.execute(update_query, [Json(fixed_paths), tuple(fixed_paths)])
```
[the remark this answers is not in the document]
aj-fuentes: > i tried following the same idea here: inner function is ignorant of query parameters (assuming table and field name also as parameters)

You need to pass also `new_field_name` (instead of reusing `new`). Note that this would probably result in shorter code if you pass the table and column to update and build the update query internally. I'll let you decide about which version seems better if you implement and test them locally.

src/util/helpers.py
```diff
@@ -317,25 +317,36 @@ def resolve_model_fields_path(cr, model, path):
...
+    import_q = """
+        SELECT id,
+               res_model,
+               STRING_TO_ARRAY(field_name, '/')
+          FROM base_import_mapping
+        """
+    if field:
+        import_q = cr.mogrify(import_q + " WHERE field_name ~ %s ", [r"\y{}\y".format(field)]).decode()
+
+    for query, table in [(export_q, "ir_exports_line"), (import_q, "base_import_mapping")]:
```
KangOl, line 341:
```suggestion
    for query, impex_model in [(export_q, "ir.exports.line"), (import_q, "base_import.mapping")]:
```

src/util/helpers.py
```diff
@@ -317,25 +317,36 @@ def resolve_model_fields_path(cr, model, path):
...
+        to_rem = [
+            path_id
+            for path_id, related_model, path in cr.fetchall()
+            if any(
+                x.field_model == model and (field is None or x.field_name == field)
+                for x in resolve_model_fields_path(cr, related_model, path)
+            )
+        ]
+        if to_rem:
+            cr.execute("DELETE FROM {} WHERE id IN %s".format(table), [tuple(to_rem)])
```
KangOl, line 352:
```suggestion
            remove_records(cr, impex_model, to_rem)
```

KangOl (changes requested):

src/util/helpers.py
```diff
@@ -317,25 +317,38 @@ def resolve_model_fields_path(cr, model, path):
...
+        to_rem = [
+            path_id
+            for path_id, related_model, path in cr.fetchall()
+            if any(
+                x.field_model == model and (field is None or x.field_name == field)
+                for x in resolve_model_fields_path(cr, related_model, path)
+            )
+        ]
+        if to_rem:
+            from .records import remove_records
```
KangOl, line 352: I'm not a fan of this import in a function body, moreover in a loop.
Maybe this function should be moved to `records.py`

aj-fuentes: I don't see any issue with the new dependency. We can figure out something later if it becomes an issue.

KangOl: > is that an ok consequence?

Yes. I don't see why it can be a problem.

src/util/helpers.py
```diff
@@ -317,25 +317,38 @@ def resolve_model_fields_path(cr, model, path):
...
-    """
+def _remove_import_export_paths(cr, model, field=None):
+    export_q = """
+            SELECT el.id,
+                   e.resource,
+                   STRING_TO_ARRAY(el.name, '/')
+              FROM ir_exports_line el
+              JOIN ir_exports e
+                ON el.export_id = e.id
+        """
```
aj-fuentes, line 328:
```suggestion
        SELECT el.id,
               e.resource,
               STRING_TO_ARRAY(el.name, '/')
          FROM ir_exports_line el
          JOIN ir_exports e
            ON el.export_id = e.id
    """
```

src/util/helpers.py
```diff
@@ -317,25 +317,38 @@ def resolve_model_fields_path(cr, model, path):
...
-    if to_rem:
-        cr.execute("DELETE FROM ir_exports_line WHERE id IN %s", [tuple(to_rem)])
+        export_q = cr.mogrify(export_q + " WHERE el.name ~ %s ", [r"\y{}\y".format(field)]).decode()
+
+    import_q = """
+        SELECT id,
+               res_model,
+               STRING_TO_ARRAY(field_name, '/')
+          FROM base_import_mapping
+        """
```
aj-fuentes, line 337:
```suggestion
    SELECT id,
           res_model,
           STRING_TO_ARRAY(field_name, '/')
      FROM base_import_mapping
    """
```

src/util/fields.py
```diff
@@ -208,8 +208,8 @@ def clean_context(context):
             [(fieldname, fieldname + " desc"), model, r"\y{}\y".format(fieldname)],
         )
 
-    # ir.exports.line
-    _remove_export_lines(cr, model, fieldname)
+    # for ir.exports.line and base_import.mapping we have to take care of "nested" references in fields "paths"
```
aj-fuentes, line 211: We can get rid of this comment, the code in the function would speak by itself.
```suggestion
```

src/util/fields.py
```diff
@@ -1079,8 +1078,23 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
         """
         cr.execute(q.format(col_prefix=col_prefix), p)
 
-        # ir.exports.line
+        # ir.exports.line, base_import.mapping # noqa
         if only_models:
+
+            def fix_paths(potential_broken_paths, old_field_name, new_field_name):
```
aj-fuentes, line 1084: Note: now that we do this more than once, you can move all one in this if block to a `_update_impex_fields_paths` to be in line with the other new function ;)

Untested! just a copy-pasta example:
```py
def _update_impex_fields_paths(cr, model, old, new):
    export_q = cr.mogrify(
        """
        SELECT el.id,
               e.resource,
               STRING_TO_ARRAY(el.name, '/')
          FROM ir_exports_line el
          JOIN ir_exports e
            ON el.export_id = e.id
         WHERE el.name ~ %s
        """,
        [r"\y{}\y".format(old)],
    ).decode()
    impex_data = [("ir_exports_line", "name", export_q)]
    if table_exists(cr, "base_import_mapping"):
        import_q = cr.mogrivy(
            """
            SELECT id,
                   res_model,
                   STRING_TO_ARRAY(field_name, '/')
              FROM base_import_mapping
             WHERE field_name ~ %s
            """,
            [r"\y{}\y".format(old)],
        ).decode()
        impex_data.append(("base_import_mapping", "field_name", import_q))
    
    for table, column, query in impex_data:
        cr.execute(query)
        if not cr.rowcount:
            continue
        fixed_paths = {}
        for record_id, related_model, path in cr.fetachall():
            new_path = [
                new
                if field.field_name == old and field.field_model in only_models
                else field.field_name
                for field in resolve_model_fields_path(cr, related_model, path)
            ]
            if len(new_path) == len(path) and new_path != path:
                fixed_paths[record_id] = "/".join(new_path)
        if fixed_paths:
            cr.execute(
                format_query(cr, "UPDATE {} SET {} = (%s::jsonb)->>(id::text) WHERE id IN %s", table, column),
                [Json(fixed_paths), tuple(fixed_paths)],
            )
```

aj-fuentes: You can keep it inline, but this entry function `rename_field` is getting very long already, since we have a similar function for removal (although it's true that the latter is used in multiple places) we could have it extracted into a new funciton with a relevant name and the code here would look neater.

src/util/models.py
```diff
@@ -11,7 +11,12 @@
 
 from .const import ENVIRON
 from .fields import IMD_FIELD_PATTERN, remove_field
-from .helpers import _ir_values_value, _remove_export_lines, _validate_model, model_of_table, table_of_model
+from .helpers import (
+    _ir_values_value,
+    _validate_model,
+    model_of_table,
+    table_of_model,
```
KangOl, line 18: Could be kept in one line
(remove the trailing comma to force `ruff` to reformat it on one line)
```suggestion
    table_of_model
```

aj-fuentes (approved):
LGTM

@KangOl ?

KangOl (approved):
@robodoo r+

PR #101, opened 2024-06-25
[FIX] util/helpers: do not override docs in Python2

aj-fuentes:
We cannot write to `__doc__` attributes of `namedtuple`s in Python2
```
Python 2.7.18 (default, Oct 15 2023, 16:43:11)
[GCC 11.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import namedtuple
>>> x = namedtuple("X", "x")
>>> x.__doc__ = "x"
AttributeError: attribute '__doc__' of 'type' objects is not writable
```

### DIFF
--- a/src/util/helpers.py
+++ b/src/util/helpers.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 import os
+import sys
 from collections import namedtuple
 
 import lxml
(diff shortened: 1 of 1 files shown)


KangOl (approved):
@robodoo r+

KangOl:
Nice picture of python2 and python3 happy together...

PR #103, opened 2024-07-01
[IMP] util/orm: do not tweak res.users.lang cache validation

aj-fuentes:
`res.users.lang` is a widely used field in the ORM that it's better to not mess with for performance reasons. If a DB has invalid codes the ORM would deal with them graciously.


KangOl:

KangOl (approved):
@robodoo r+

PR #107, opened 2024-07-05
[FIX] util.fields: error when table of model base_import.mapping doesn't exist

luisg123v:
In 1d9e20e5, a new feature was implemented to consider the model `base_import.mapping` when removing fields. However, it was not considering if the table doesn't exist, which could occur when migrating from a version <= 11.0. That causes the following error in such cases:

    relation "base_import_mapping" does not exist

This commit fixes the above issue by ensuring the table actually exists before accessing it.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -495,18 +495,21 @@ def _remove_import_export_paths(cr, model, field=None):
     else:
         export_q += " WHERE el.name IS NOT NULL"
 
-    import_q = """
-        SELECT id,
-               res_model,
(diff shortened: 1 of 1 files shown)


luisg123v:
Hi @pauloamed, @KangOl,

Could you review this fix, please?

CC @desdelinux @claurelb @moylop260

desdelinux:
@aj-fuentes @KangOl @pauloamed @andreabak 

Could you please review this?

KangOl:
Thank you for the fix!

@robodoo r+ priority alone cancel=staging

PR #110, opened 2024-07-09
[FIX] tests/test_util: filters in `sort` must be a JSON array

aj-fuentes:
Since odoo/odoo@5b95abb8 we enforce `sort` column to be a stringified
JSON array of field names.


KangOl (approved):
@robodoo r+

PR #112, opened 2024-07-11
[FIX] attr_domains2expr: Remove redundant xpaths

sagu-odoo:
if empty attrs is present which useless itself like this

**org**
```
<xpath expr='..' position='attributes'>
	<attribute name='attrs'/>
</xpath>
```

**before fix**
```
<xpath expr='..' position='attributes'>
</xpath>
```
image

this is just snippet of view their can be more customization.Yes it is customization issue but if we are removing attrs empty node. So, this can be handled by removing that xpath also .So at least studio view or any view won't disable due to this while validating view ``check_xml``. 

**traceback**
```
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 2443, in <dictcomp>
    v_type: self.get_view(
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 2628, in get_view
    result = dict(self._get_view_cache(view_id, view_type, **options))
  File "<decorator-gen-51>", line 2, in _get_view_cache
  File "/home/odoo/odoo/odoo/17.0/odoo/tools/cache.py", line 104, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 2584, in _get_view_cache
    arch, view = self._get_view(view_id, view_type, **options)
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 2531, in _get_view
    arch = view._get_combined_arch()
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 950, in _get_combined_arch
    arch = root.with_prefetch(tree_views._prefetch_ids)._combine(hierarchy)
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 891, in _combine
    combined_arch = view.apply_inheritance_specs(combined_arch, arch)
  File "/home/odoo/odoo/enterprise/17.0/web_studio/models/ir_ui_view.py", line 603, in apply_inheritance_specs
    return self._apply_studio_specs(source, specs_tree)
  File "/home/odoo/odoo/enterprise/17.0/web_studio/models/ir_ui_view.py", line 586, in _apply_studio_specs
    source = super(View, self).apply_inheritance_specs(source, spec)
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 828, in apply_inheritance_specs
    self._raise_view_error(str(e), specs_tree)
  File "/home/odoo/odoo/odoo/17.0/odoo/addons/base/models/ir_ui_view.py", line 705, in _raise_view_error
    raise err from from_exception
ValueError: Element '<xpath expr="//form[1]/sheet[1]/group[1]/group[2]/field[@name=&#39;tag_ids&#39;]">' cannot be located in parent view

```
**after fix** 
it will also remove because this redundant and does not make any sense present in view.

### DIFF
--- a/src/base/17.0.1.3/attr_domains2expr.py
+++ b/src/base/17.0.1.3/attr_domains2expr.py
@@ -396,6 +396,8 @@ def fix_attrs(cr, model, arch, comb_arch):
             new_elem = etree.Element("attribute", name=mod)
             new_elem.text = fake_elem.get(mod)
             parent.append(new_elem)
+        if len(parent) == 0 and parent.getparent() is not None:
+            parent.getparent().remove(parent)
 
     return success
 


src/base/17.0.1.3/attr_domains2expr.py
```diff
@@ -377,15 +377,21 @@ def fix_attrs(cr, model, arch, comb_arch):
     # </field>
     # becomes the fake element
     # <field name=... position=replace attrs={invisible: xxx} invisible=yyy readonly=zzz/>`
+    empty_parent = False  # for check parent is empty or not
```
aj-fuentes, line 380:
```suggestion
```

sagu-odoo:
Hello @aj-fuentes , suggested changes done could you please have a look?. thanks ;)

aj-fuentes:
Isn't this a very specific case? If the view didn't make sense originally why should we patch this?

sagu-odoo:
Yes, you are right. Still, attrs will remove from node shouldn't be better to remove node also otherwise, it can disable the view if the xpath is not compatible. That is why I made this fix.

sagu-odoo:
I have added traceback to the commit and bit detail.thanks

aj-fuentes:
I still believe this is not a problem here, the original view seems incorrect. For example if there was something else in that xpath the view would still fail, isn't it?
Do you have more cases where this fix could actually avoid an error?

aj-fuentes (approved):

KangOl (approved):
@robodoo r+

PR #114, opened 2024-07-14
[FIX] util.indirect_references: add new, special case for ir_act_window

diagnoza:
`ir_act_window` is a special case where it's possible to have `res_model` set and `res_id` NULL
at the same time, contrary to the usual where `res_id` points to the table of `res_model`,
thus it usually makes sense to have both set. Because of this we need two IR entries that will
cover both cases.
    
An example of this special case can be an action that was created automatically for a button
created through Studio. Steps to reproduce:

1. Create a database in 15.0 and install `web_studio` and `sale_subscription`.

2. Using Studio, create a button in form view of `res.partner` that as a related field takes any field from `sale.subscription`. Studio will automatically create an action for this button with `res_model = 'sale.subscription'` and `res_id = NULL`.

3. Upgrade to 16.0. When the merge of `sale.subscription` into `sale.order` happens, the action created for the button will be removed from `ir_act_window` because it's still referencing `sale.subscription` instead of `sale.order`. That in turn is caused by `replace_record_references_batch` not replacing references for `ir_act_window` records that don't have `res_id` set.

### DIFF
--- a/src/util/indirect_references.py
+++ b/src/util/indirect_references.py
@@ -29,6 +29,7 @@ def indirect_references(cr, bound_only=False):
         IR("ir_cron", "model", None, set_unknown=True),
         IR("ir_act_report_xml", "model", None, set_unknown=True),
         IR("ir_act_window", "res_model", "res_id"),
+        IR("ir_act_window", "res_model", None),
         IR("ir_act_window", "src_model", None),
         IR("ir_act_server", "wkf_model_name", None),
         IR("ir_act_server", "crud_model_name", None),


aj-fuentes:
This patch looks wrong. This function is replacing _records_, the replace of model names is done [elsewhere](https://github.com/odoo/upgrade-util/blob/8d25062c14dd6c5fd59f300b60d8ec26cb44aa4a/src/util/models.py#L436-L438)

I'd suggest adding a query just for `ir_act_window` because this seems to be a especial case: it has `res_id` field but it can be `NULL`. If we are willing to accept the tables as holding correct information we could check if the `ir.res_id` column is nullable:
```py
if ir.res_model and (not ir.res_id or is_nullable(ir.res_id)) and ir.table not in ignores:
```

@KangOl wdyt?

diagnoza:
> This patch looks wrong. This function is replacing _records_, the replace of model names is done [elsewhere](https://github.com/odoo/upgrade-util/blob/8d25062c14dd6c5fd59f300b60d8ec26cb44aa4a/src/util/models.py#L436-L438)
> 
> I'd suggest adding a query just for `ir_act_window` because this seems to be a especial case: it has `res_id` field but it can be `NULL`. If we are willing to accept the tables as holding correct information we could check if the `ir.res_id` column is nullable:
> 
> ```python
> if ir.res_model and (not ir.res_id or is_nullable(ir.res_id)) and ir.table not in ignores:
> ```
> 
> @KangOl wdyt?

Yes, I think it's a valid point.

KangOl:
I don't think it's correct.
As Alvaro said, this function is for replacing records.
This case should be handled by `merge_models`.

If the `res_id` being empty on `ir_act_window` is a valid case, then, the correct patch is
```diff
diff --git src/util/indirect_references.py src/util/indirect_references.py
index 7780366..4bd9ac8 100644
--- src/util/indirect_references.py
+++ src/util/indirect_references.py
@@ -29,6 +29,7 @@ def indirect_references(cr, bound_only=False):
         IR("ir_cron", "model", None, set_unknown=True),
         IR("ir_act_report_xml", "model", None, set_unknown=True),
         IR("ir_act_window", "res_model", "res_id"),
+        IR("ir_act_window", "res_model", None),
         IR("ir_act_window", "src_model", None),
         IR("ir_act_server", "wkf_model_name", None),
         IR("ir_act_server", "crud_model_name", None),
```

aj-fuentes:
> If the res_id being empty on ir_act_window is a valid case, then, the correct patch is

I think we need a hybrid approach. Because in some cases we can correctly replace records in batch using res_id, while in others (null res_id) we can only replace the model. With that patch we will replacing models always. Hence why I believe we need a special case in `merge_models` to update the rows with null `res_id`.

diagnoza:
> > If the res_id being empty on ir_act_window is a valid case, then, the correct patch is
> 
> I think we need a hybrid approach. Because in some cases we can correctly replace records in batch using res_id, while in others (null res_id) we can only replace the model. With that patch we will replacing models always. Hence why I believe we need a special case in `merge_models` to update the rows with null `res_id`.

Isn't that was Christophe's patch would do? See that his suggestion is to have two entries for `ir_act_window`, one with `res_id` and one without:

```
IR("ir_act_window", "res_model", "res_id"),
IR("ir_act_window", "res_model", None),
```

That would result in replacing records where possible when `IR("ir_act_window", "res_model", "res_id")` is processed and then replacing just the model due to subsequently running `IR("ir_act_window", "res_model", None)`.

aj-fuentes:
Ah, yes, I missread the patch, it's adding one line, not changing the current one. Then yes it would work IMO.

diagnoza:
@KangOl @aj-fuentes it's done, tested and works.

KangOl:
Can you please change the commit message to explain this special case?

diagnoza:
> Can you please change the commit message to explain this special case?

Done.

KangOl:
upgradeci retry with only sale_subscription

KangOl (approved):
@robodoo r+

PR #115, opened 2024-07-18
[ADD] specific: rename_custom_field

Pirols:
Joining the `rename_custom_{model,module,table,column}` family.

```
> rg 'rename_field.*custom"' src/upgrade-specific/scripts -tpy | wc -l
85
```

Recent example: https://github.com/odoo/upgrade-specific/pull/16967

### DIFF
--- a/src/util/specific.py
+++ b/src/util/specific.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 
+from .fields import rename_field
 from .helpers import _validate_table
 from .misc import _cached
 from .models import rename_model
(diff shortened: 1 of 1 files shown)


KangOl:

src/util/specific.py
```diff
@@ -62,6 +63,21 @@ def rename_custom_model(cr, model_name, new_model_name, custom_module=None, repo
     )
 
 
+def rename_custom_field(cr, old_field_name, new_field_name, model, report_details=""):
```
UemuS, line 66: The entire family would want `model` as the first arg :p

UemuS (approved):

jjmaksoud:
I have the impression we are discouraged from renaming fields in specific fixes, renaming columns is the way to go unless it doesnt work for some reason

Pirols:
> I have the impression we are discouraged from renaming fields in specific fixes, renaming columns is the way to go unless it doesnt work for some reason

I think it's not much different from renaming a model: if an old custom field at some point clashes with a newly introduced standard field we need to rename or remove it. Renaming the column only wouldn't cut, would it?

aj-fuentes:
I'm not in favor of this. Renaming fields is rarely the way to go for specific fixes. When it is, namely when there is a client's support ticket, it can be done with the existing tools.

aj-fuentes:
> Renaming the column only wouldn't cut, would it?

It does in most cases. If it doesn't then it needs further analysis. But renaming a field is a too big operation that should be avoided.

aj-fuentes:
>  think it's not much different from renaming a model

In some cases we can rename a table and it works. When not then we need to actually rename the model. In most situations where models clash the existing model is a backport of the to-be-added model, or something compatible. Since models have higher chances to be incompatible than single fields, it makes more sense to fully rename a model. But I think renaming a field can be avoided more easily.

Pirols:
IRL discussion summary: While not having this unfortunately won't prevent anyone from renaming fields in specific scripts at least it won't reinforce/validate it more.

PR #117, opened 2024-07-22
[FIX] tests/test_util: use new cursor for commit in tests

aj-fuentes:
Since https://github.com/odoo/odoo/commit/1241fac8 we can not longer commit test cursors. Some
tests here actually test flows using `commit`. To avoid the issue we use
an extra cursor.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -652,8 +652,13 @@ class TestPG(UnitTestCase):
         result = cr.fetchone()[0]
         self.assertEqual(result, expected)
 
+    def _get_cr(self):
+        cr = self.registry.cursor()
+        self.addCleanup(cr.close)
+        return cr
+
     def test_explode_mult_filters(self):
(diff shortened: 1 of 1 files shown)


aj-fuentes:
upgradeci retry with always base in versions 8.0 9.0 10.0 11.0 12.0 saas~12.3 13.0 14.0 15.0 16.0 17.0 saas~17.5

src/base/tests/test_util.py
```diff
@@ -652,8 +652,11 @@ def test_pg_text2html(self, value, expected):
         result = cr.fetchone()[0]
         self.assertEqual(result, expected)
 
+    def _get_cr(self):
+        return self.registry.cursor()
```
KangOl, line 656:
```suggestion
        cr = self.registry.cursor()
        self.addCleanup(cr.close)
        return cr
```

KangOl (approved):
@robodoo r+ priority

PR #119, opened 2024-07-24
[IMP] util/records: Fix edit_view docstring

elre-odoo:
[`edit_view`](https://github.com/odoo/upgrade-util/blob/c2cd2b7645d8d6a2f023f7a3ae2d75f8411ebea4/src/util/records.py#L169-L280) yields an xmlid view according to the following logic:
> If the view has `noupdate=False`, we assume that it is a standard view that'll be updated by the ORM later on, so by default we don't yield it when calling `edit_view`. We may force yield such a view when setting 'skip_if_not_noupdate` to False.

This behaviour was initially described correctly in the docstring, but the values of `noupdate` got mixed up in the new docstring in aaa1f0fee6870075e25cb5e6744e4c589bb32b46 .
This commit reverts the values in the docstring to the correct ones.


KangOl:
🤦 

@robodoo r+

PR #120, opened 2024-07-25
[FIX] rename_model: don't rename table to itself

KangOl:
When renaming a model, it may happen that the table stay the same when underscores are replaced by dots.


KangOl:
upgradeci retry with only mail

KangOl:

aj-fuentes (approved):

KangOl:
@robodoo r+

PR #124, opened 2024-08-08
[IMP] core: jsonb company dependent field

HydrionBurst:
support jsonb company dependent field

See https://github.com/odoo/odoo/pull/175627
See https://github.com/odoo/enterprise/pull/67925
See https://github.com/odoo/upgrade/pull/6366

Task: 3954610

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -506,7 +506,8 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
             # (before saas~11.2, where pattern changed)
             name = "%s_%s" % (name, fid)
             cr.execute("UPDATE ir_model_data SET name=%s WHERE model='ir.model.fields' AND res_id=%s", [name, fid])
-        cr.execute("UPDATE ir_property SET name=%s WHERE fields_id=%s", [new, fid])
+        if table_exists(cr, "ir_property"):
+            cr.execute("UPDATE ir_property SET name=%s WHERE fields_id=%s", [new, fid])
     # Updates custom field relation name to match the renamed standard field during upgrade.
     cr.execute(
         """
@@ -652,7 +653,66 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
         convert_field_to_html(cr, inh.model, field, skip_inherit=skip_inherit)
 
 
-def convert_field_to_property(
+def _convert_field_to_company_dependent(
+    cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
+):
+    _validate_model(model)
+    if target_model:
+        _validate_model(target_model)
+
+    type2field = {"char", "float", "boolean", "integer", "text", "many2one", "date", "datetime", "selection", "html"}
+    assert type in type2field
+
+    table = table_of_model(cr, model)
+
+    # update "ir_model_fields"."company_dependent" as an identifier for indirect reference
+    cr.execute(
+        """
+        UPDATE ir_model_fields
+           SET company_dependent = TRUE
+         WHERE model = %s
+           AND name = %s
+     RETURNING id
+        """,
+        (model, field),
+    )
+    if not cr.rowcount:
+        # no ir_model_fields, no column
+        remove_column(cr, table, field, cascade=True)
+        return
+    [field_id] = cr.fetchone()
+
+    if default_value is None:  # noqa: SIM108
+        where_condition = "{0} IS NOT NULL"
+    else:
+        where_condition = cr.mogrify("{0} != %s", [default_value]).decode()
+    where_condition += format_query(cr, " AND {} IS NOT NULL", company_field)
+
+    using = format_query(cr, "jsonb_build_object({}, {{0}})", company_field)
+    alter_column_type(cr, table, field, "jsonb", using=using, where=where_condition)
+
+    # delete all old default
+    cr.execute("DELETE FROM ir_default WHERE field_id = %s", [field_id])
+
+    # add fallback
+    if default_value:
+        cr.execute(
+            "INSERT INTO ir_default(field_id, json_value) VALUES (%s, %s) RETURNING id",
+            (field_id, json.dumps(default_value)),
+        )
+        [default_id] = cr.fetchone()
+        if default_value_ref:
+            module, _, xid = default_value_ref.partition(".")
+            cr.execute(
+                """
+                INSERT INTO ir_model_data(module, name, model, res_id, noupdate)
+                     VALUES (%s, %s, 'ir.default', %s, True)
+                """,
+                [module, xid, default_id],
+            )
+
+
+def _convert_field_to_property(
     cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
 ):
     """
@@ -777,7 +837,13 @@ def convert_field_to_property(
 
 
 # alias with a name related to the new API to declare property fields (company_dependent=True attribute)
-make_field_company_dependent = convert_field_to_property
+if version_gte("saas~17.5"):
+    make_field_company_dependent = _convert_field_to_company_dependent
+else:
+    make_field_company_dependent = _convert_field_to_property
+
+# retro-compatibility
+convert_field_to_property = make_field_company_dependent
 
 
 def convert_binary_field_to_attachment(cr, model, field, encoded=True, name_field=None):
--- a/src/util/indirect_references.py
+++ b/src/util/indirect_references.py
@@ -5,7 +5,11 @@ from .helpers import model_of_table, table_of_model
 from .pg import column_exists, table_exists
 
 
-class IndirectReference(collections.namedtuple("IndirectReference", "table res_model res_id res_model_id set_unknown")):
+class IndirectReference(
+    collections.namedtuple(
+        "IndirectReference", "table res_model res_id res_model_id set_unknown company_dependent_comodel"
+    )
+):
     def model_filter(self, prefix="", placeholder="%s"):
         if prefix and prefix[-1] != ".":
             prefix += "."
@@ -109,12 +113,25 @@ def indirect_references(cr, bound_only=False):
 
         yield ir
 
+    if column_exists(cr, "ir_model_fields", "company_dependent"):
+        cr.execute(
+            """
+            SELECT model, name, relation
+              FROM ir_model_fields
+             WHERE company_dependent IS TRUE
+               AND ttype = 'many2one'
+            """,
+        )
+        for table_name, column_name, comodel_name in cr.fetchall():
+            yield IR(table_name, None, column_name, company_dependent_comodel=comodel_name)
+
     # XXX Once we will get the model field of `many2one_reference` fields in the database, we should get them also
     # (and filter the one already hardcoded)
 
 
 def generate_indirect_reference_cleaning_queries(cr, ir):
     """Yield queries to clean an `IndirectReference`."""
+    assert not ir.company_dependent_comodel  # not supported for now
     if ir.res_model:
         query = """
             SELECT {ir.res_model}
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -22,6 +22,7 @@ from .pg import (
     column_updatable,
     explode_execute,
     explode_query_range,
+    format_query,
     get_m2m_tables,
     get_value_or_en_translation,
     parallel_execute,
@@ -76,6 +77,34 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
 
     # remove references
     for ir in indirect_references(cr):
+        if ir.company_dependent_comodel:
+            if ir.company_dependent_comodel == "ir.model":
+                # clean on delete set null
+                cr.execute("SELECT id FROM ir_model WHERE name = %s", [model])
+                [mod_id] = cr.fetchone() or [None]
+                if mod_id:
+                    cr.execute(
+                        format_query(
+                            """
+                            UPDATE {table}
+                               SET {field} = (
+                                   SELECT jsonb_object_agg(
+                                          key,
+                                          CASE
+                                              WHEN value::int4 = %s THEN NULL
+                                              ELSE value::int4
+                                          END)
+                                     FROM jsonb_each_text({field})
+                                   )
+                             WHERE {field} IS NOT NULL
+                               AND {field} @? %s
+                            """,
+                            table=ir.table,
+                            field=ir.res_id,
+                        ),
+                        [mod_id, "$.* ? (@ == {})".format(mod_id)],
+                    )
+            continue
         if ir.table in ("ir_model", "ir_model_fields", "ir_model_data"):
             continue
         ref_model = model_of_table(cr, ir.table)
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -410,8 +410,34 @@ def remove_records(cr, model, ids):
     table = table_of_model(cr, model)
     cr.execute('DELETE FROM "{}" WHERE id IN %s'.format(table), [ids])
     for ir in indirect_references(cr, bound_only=True):
-        query = 'DELETE FROM "{}" WHERE {} AND "{}" IN %s'.format(ir.table, ir.model_filter(), ir.res_id)
-        cr.execute(query, [model, ids])
+        if not ir.company_dependent_comodel:
+            query = 'DELETE FROM "{}" WHERE {} AND "{}" IN %s'.format(ir.table, ir.model_filter(), ir.res_id)
+            cr.execute(query, [model, ids])
+        elif ir.company_dependent_comodel == model:
+            query = cr.mogrify(
+                format_query(
+                    cr,
+                    """
+                    UPDATE {table}
+                    SET {column} = (
+                        SELECT jsonb_object_agg(
+                            key,
+                            CASE
+                                WHEN value::int4 IN %s THEN NULL
+                                ELSE value::int4
+                            END)
+                        FROM jsonb_each_text({column})
+                    )
+                    WHERE {column} IS NOT NULL
+                    AND {column} @? %s
+                    AND {{parallel_filter}}
+                    """,
+                    table=ir.table,
+                    column=ir.res_id,
+                ),
+                [ids, "$.* ? ({})".format(" || ".join(map("@ == {}".format, ids)))],
+            ).decode()
+            explode_execute(cr, query, table=ir.table)
     _rm_refs(cr, model, ids)
 
     if model == "res.groups":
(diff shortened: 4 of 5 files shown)


src/util/orm.py
```diff
@@ -221,7 +221,7 @@ def no_selection_cache_validation(f=None):
         return f
     old_convert = ofields.Selection.convert_to_cache
 
-    def _convert(self, value, record, validate=True):
+    def _convert(self, value, record, validate=True, **kwargs):
         if self.model_name == "res.users" and self.name == "lang":
             return old_convert(self, value, record, validate=validate)
         return old_convert(self, value, record, validate=False)
```
KangOl, line 227: Pass-through kwargs
```suggestion
            return old_convert(self, value, record, validate=validate, **kwargs)
        return old_convert(self, value, record, validate=False, **kwargs)
```

KangOl (changes requested):
Please [refer](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests) to the odoo/enterprise/design-themes PR (if applicable) in the **commit message**.
It's needed by [UpgradeCI](https://github.com/odoo/upgrade/wiki/CI:-upgradeci) to correctly build this PR with the corresponding branches.

If you don't want to have notifications on the related PR each time you force push, simply don't force push and add new [fixup commits](https://git-scm.com/docs/git-commit/2.32.0#Documentation/git-commit.txt---fixupamendrewordltcommitgt) that will be squashed before the r+.

Note that UpgradeCI will also extract the info from *the PR body*. It is however better to have those references in the commit message itself.

<sub><i>This message is a canned response and thus may not be totally accurate.</i></sub>

src/util/fields.py
```diff
@@ -631,6 +632,71 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+        where_condition = f"{_field} IS NOT NULL"
+    else:
+        where_condition = f"{_field} != %s"
+        params.append(default_value)
+
+    cr.execute(f"ALTER TABLE {table} RENAME COLUMN {field} TO {_field}")
+
+    if is_field_anonymized(cr, model, field):
+        # TODO: cwg not sure what to do
+        pass
```
KangOl, line 665: Nothing. You can just ignore that.
```suggestion
```

src/util/fields.py
```diff
@@ -631,6 +632,71 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+    cr.execute(f"ALTER TABLE {table} ADD COLUMN {field} jsonb")
+    cr.execute(
+        f"""
+        UPDATE {table}
+           SET {field} = jsonb_build_object({company_field}, {_field})
+         WHERE {where_condition}
+        """,
+        params,
+    )
+    cr.execute(f"ALTER TABLE {table} DROP COLUMN {_field}")
```
KangOl, line 676: Use `util.alter_column_type`

src/util/fields.py
```diff
@@ -756,8 +822,10 @@ def convert_field_to_property(
 
 
 # alias with a name related to the new API to declare property fields (company_dependent=True attribute)
-make_field_company_dependent = convert_field_to_property
-
+if version_gte("saas~17.5"):
+    make_field_company_dependent = convert_field_to_company_dependent
+else:
+    make_field_company_dependent = convert_field_to_property
```
KangOl, line 828: I don't like the fact that you have the two *public* methods available.

I propose this design instead.
```python

def _convert_field_to_property(
    cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
):
    # existing `convert_field_to_property` function
    ...
def _convert_field_to_company_dependent_jsonb(
     cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
):
     # new impl
     ...

def make_field_company_dependent(
    cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
):
    if table_exists(cr, "ir_property"):
        return _convert_field_to_property(...)
    else:
        return _convert_field_to_company_dependent_jsonb(...)

# retro-compatibility alias
convert_field_to_property = make_field_company_dependent
```

KangOl (changes requested):

src/util/fields.py
HydrionBurst: I cannot directly use `alter_column_type`
1. the logic ```remove the existing linked `ir_model_fields_selection` recods in case it was a selection field``` shouldn't be executed, since changing a selection field to company dependent field shouldn't remove all its selection options
2. I need a parameter `where` to avoid migrate data which is the same as the default value

I think the current `alter_column_type` is a good and simple logic abstraction, it would be better to not modify it and just copy the similar code to `convert_field_to_company_dependent`

What do you think?

KangOl: 1. This is not a problem. The `ir_model_fields_selection` records are not required. They will be recreated by the ORM when reflecting the field. You can also modify this function to skip this step conditionally.
2. it could be included in the `using` clause by wrapping the `jsonb_build_object` around a `CASE WHEN`.

HydrionBurst: Another problem is the SQL requires variable for `default`
I guess `explode_execute` doesn't support params?

KangOl: You just call `cr.mogrify` to inject params.

src/util/records.py
```diff
@@ -410,8 +410,28 @@ def remove_records(cr, model, ids):
...
+                        SELECT jsonb_object_agg(
+                            key,
+                            CASE
+                                WHEN (value->>0)::int4 in %s THEN NULL
+                                ELSE value
+                            END)
+                        FROM jsonb_each({ir.res_id})
+                    )
+                    WHERE {ir.res_id} IS NOT NULL
+                    AND {ir.res_id} @? %s
```
aj-fuentes, line 428: Can't we remove the key completely instead of setting null for the sake of simplicity?
nit: formatting
```suggestion
                    UPDATE {ir.table}
                       SET {ir.res_id} = (
                           SELECT jsonb_object_agg(key, value)
                             FROM jsonb_each({ir.res_id})
                            WHERE (value->>0)::int4 NOT IN %s
                           )
                     WHERE {ir.res_id} IS NOT NULL
                       AND {ir.res_id} @? %s
```

src/util/models.py
```diff
@@ -76,6 +76,38 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+            if ir.company_dependent_comodel != 'ir.model':
+                # we don't handle company dependent here
+                # the field should be field_remove when upgrade ir.company_dependent_comodel
+                # since the referenced model here is removed
+                continue
+            # clean on delete set null
+            cr.execute("SELECT id FROM ir_model WHERE name = %s", [model])
+            [mod_id] = cr.fetchone() or [None]
+            if mod_id:
+                cr.exeucte(
```
HydrionBurst, line 89: like remove_record, I think the below code shouldn't be executed since I cannot image the usage of
`model_id = fields.Many2one('ir.model', company_dependent=True)`
Should we remove it and just raise a warning?

src/util/indirect_references.py
```diff
@@ -109,12 +110,75 @@ def indirect_references(cr, bound_only=False):
...
         yield ir
 
+    if version_gte("saas~17.5"):
+        # jsonb company dependent fields
+        # todo: share the table with table_of_model
+        exceptions = dict(
+            line.split()
+            for line in splitlines(
+                """
+            ir.actions.actions          ir_actions
```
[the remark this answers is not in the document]
aj-fuentes: What if select all jsonb columns of stored m2o fields? You can still hardcode here the exceptions for which the model doesn't match the one from the field with the usual replace (. -> _).

I'm not worried about performance, this is not a common operation and in the context of an upgrade it won't be noticeable. I'm worried about correctness. We should update custom code fields, for which we do not have the code ever during the upgrade.

In the end the most bulletproof approach would be to have a `company_dependent` column in ir_model_fields, but that's a hard decision. But given that now we have many more jsonb columns (translated fields) I'd may be worth to have a way to differentiate them. 

Finally, is it possible to have a translated company dependent field? How does this work in such case?

src/util/models.py
aj-fuentes: `exeucte` -> `execute`

> I think the below code shouldn't be executed since I cannot image the usage of...

If it was allowed, we would get a case most probably. Thus I'd suggest you keep it.

src/util/records.py
aj-fuentes: Now with the details you gave about ir_defaults and how defaults work for the former ir_property, it's clear that we need to do as you do here. Thanks.

src/util/indirect_references.py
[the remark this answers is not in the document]
HydrionBurst: > What if we select all jsonb columns of stored m2o fields?
reduce the joined lines to improve the performance? If so do you think I should UPDATE ir_model_fields store=True in `_convert_field_to_company_dependent` in case the store column is really inconsistent with the ORM?

> Finally, is it possible to have a translated company dependent field? How does this work in such case?
Don't warry it is not supported and will raise warning in ORM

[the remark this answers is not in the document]
aj-fuentes: > reduce the joined lines to improve the performance? 

So, I just re-read the code and you are doing exactly what I asked for, my bad. The only difference is the check for `store=True`. I think it would be more consistent, if the field is not stored we can ignore the column. It may be some leftover that we better not update.

>  If so do you think I should UPDATE ir_model_fields store=True in _convert_field_to_company_dependent in case the store column is really inconsistent with the ORM?

I think you can, yes, and log an error if the column is not stored. @KangOl wdyt?

> Don't warry it is not supported and will raise warning in ORM

Here is what worries me: the only way to unambiguously differentiate right now what's the purpose of a jsonb column is via code once the fields are _loaded_. That poses problems for upgrades because we never have custom code. Thus having some metadata in ir_model_fields that clearly discriminates about what's the purpose of the column may be worth it. One hypothetical example was translated company dependent fields. Anyway I guess we can think about that once the case comes. Just please have this in mind for the (future) patches in the ORM :)

src/util/fields.py
```diff
@@ -777,8 +830,10 @@ def convert_field_to_property(
 
 
 # alias with a name related to the new API to declare property fields (company_dependent=True attribute)
-make_field_company_dependent = convert_field_to_property
-
+if version_gte("saas~17.5"):
+    make_field_company_dependent = _convert_field_to_company_dependent
+else:
+    make_field_company_dependent = _convert_field_to_property
```
KangOl, line 836: Keep the old name available for retro-compatibility
```suggestion
if version_gte("saas~17.5"):
    make_field_company_dependent = _convert_field_to_company_dependent
else:
    make_field_company_dependent = _convert_field_to_property
    
# retro-compatibility
convert_field_to_property = make_field_company_dependent
```

src/util/indirect_references.py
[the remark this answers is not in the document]
KangOl: Yes, we could force the `store` column to `true` with a warning if it wasn't.

src/util/records.py
```diff
@@ -1440,6 +1460,11 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
     for ir in indirect_references(cr, bound_only=True):
         if ir.table in ignores:
             continue
+        if ir.company_dependent_comodel:
+            if ir.company_dependent_comodel == 'model_src':
```
KangOl, line 1464:
```suggestion
            if ir.company_dependent_comodel == model_src:
```

src/util/records.py
```diff
@@ -410,8 +410,28 @@ def remove_records(cr, model, ids):
...
+                        SELECT jsonb_object_agg(
+                            key,
+                            CASE
+                                WHEN value::int4 in %s THEN NULL
+                                ELSE value::int4
+                            END)
+                        FROM jsonb_each_text({ir.res_id})
+                    )
+                    WHERE {ir.res_id} IS NOT NULL
+                    AND {ir.res_id} @? %s
```
KangOl, line 428: Use `format_query`.

src/util/records.py
```diff
@@ -1440,6 +1465,11 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
     for ir in indirect_references(cr, bound_only=True):
         if ir.table in ignores:
             continue
+        if ir.company_dependent_comodel:
+            if ir.company_dependent_comodel == model_src:
+                assert model_src == model_dst
+                raise Exception("Company dependent merge should be handled but don't support it for now")
```
aj-fuentes, line 1471: What's wrong with updating the values? Isn't it just replacing one id by another?
Something like:
```sql
                     UPDATE {table}
                        SET {column} = (
                                SELECT jsonb_object_agg(key, %s::jsonb->value)
                                  FROM jsonb_each_text({column})
                            )
                      WHERE {column} @? %s
```
with parameters `[Json(id_mapping), "$.* ? ({})".format(" || ".join(map("@ == {}".format, tuple(id_mapping)))})]`

Maybe you have some other constraints in mind, could you please elaborate?

src/util/fields.py
```diff
@@ -652,7 +653,69 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+           SET company_dependent = TRUE
+         WHERE model = %s
+           AND name = %s
+     RETURNING id
+        """,
+        (model, field)
+    )
+    if not cr.rowcount:
+        # no ir_model_fields, no column
+        remove_column(cr, table, field, cascade=True)
```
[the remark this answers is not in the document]
aj-fuentes: OK, I guess it's there of a reason.

src/util/fields.py
```diff
@@ -652,7 +653,66 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+        return
+    [field_id] = cr.fetchone()
+
+    if default_value is None:
+        where_condition = "{0} IS NOT NULL"
+    else:
+        where_condition = "{0} != " + cr.mogrify("%s", [default_value]).decode()
```
KangOl, line 688:
```suggestion
        where_condition =  cr.mogrify("{0} != %s", [default_value]).decode()
```

src/util/fields.py
```diff
@@ -652,7 +653,66 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+    where_condition += format_query(cr, " AND {} IS NOT NULL", company_field)
+
+    using = format_query(cr, "jsonb_build_object({}, {{0}})", company_field)
+    alter_column_type(cr, table, field, "jsonb", using=using, where=where_condition)
+
+    # delete all old default
+    cr.execute("DELETE FROM ir_default WHERE field_id = %s", [field_id,])
```
KangOl, line 695: Avoid `ruff` to reformat this on multiple lines.
```suggestion
    cr.execute("DELETE FROM ir_default WHERE field_id = %s", [field_id])
```

src/util/fields.py
```diff
@@ -777,8 +837,13 @@ def convert_field_to_property(
...
 
 # alias with a name related to the new API to declare property fields (company_dependent=True attribute)
-make_field_company_dependent = convert_field_to_property
+if version_gte("saas~17.5"):
+    make_field_company_dependent = _convert_field_to_company_dependent
+else:
+    make_field_company_dependent = _convert_field_to_property
 
+# retro-compatibility
+convert_field_to_property = make_field_company_dependent
```
KangOl, line 846: Extra line to make `ruff` happy.
```suggestion
convert_field_to_property = make_field_company_dependent

```

src/util/indirect_references.py
```diff
@@ -109,12 +110,27 @@ def indirect_references(cr, bound_only=False):
 
         yield ir
 
+    # use column_exists instead of version_gte("saas~17.5") since for
+    # base/0.0.0/pre-clean-transients.py, the column hasn't been added
```
KangOl, line 114: Useless comment.
```suggestion
```

src/util/indirect_references.py
```diff
@@ -2,10 +2,11 @@
 import collections
 
 from .helpers import model_of_table, table_of_model
+from .misc import version_gte, splitlines
```
KangOl, line 5: Useless import
```suggestion
```

src/util/models.py
```diff
@@ -30,7 +30,7 @@
 )
 
 # avoid namespace clash
-from .pg import rename_table as pg_rename_table
+from .pg import rename_table as pg_rename_table, format_query
```
KangOl, line 33: Import `format_query` in the previous statement.

src/util/models.py
```diff
@@ -76,6 +76,34 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+                                          END)
+                                     FROM jsonb_each_text({field})
+                                   )
+                             WHERE {field} IS NOT NULL
+                               AND {field} @? %s
+                            """,
+                            table=ir.table,
+                            field=ir.res_id
+                        ),
+                        [ids, "$.* ? (@ == {})".format(mod_id)]
```
KangOl, line 104: `ids` is not defined.
```suggestion
                        [mod_id, "$.* ? (@ == {})".format(mod_id)]
```

src/util/models.py
```diff
@@ -76,6 +76,34 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+                if mod_id:
+                    cr.execute(
+                        format_query(
+                            """
+                            UPDATE {table}
+                               SET {field} = (
+                                   SELECT jsonb_object_agg(
+                                          key,
+                                          CASE
+                                              WHEN value::int4 in %s THEN NULL
```
KangOl, line 93:
```suggestion
                                              WHEN value::int4 = %s THEN NULL
```

src/util/records.py
```diff
@@ -410,8 +410,33 @@ def remove_records(cr, model, ids):
...
+                cr.execute(
+                    cr.format_query(
+                        cr,
+                        """
+                        UPDATE {table}
+                        SET {column} = (
+                            SELECT jsonb_object_agg(
+                                key,
+                                CASE
+                                    WHEN value::int4 in %s THEN NULL
```
KangOl, line 424:
```suggestion
                                    WHEN value::int4 IN %s THEN NULL
```

src/util/records.py
```diff
@@ -410,8 +410,33 @@ def remove_records(cr, model, ids):
     table = table_of_model(cr, model)
     cr.execute('DELETE FROM "{}" WHERE id IN %s'.format(table), [ids])
     for ir in indirect_references(cr, bound_only=True):
-        query = 'DELETE FROM "{}" WHERE {} AND "{}" IN %s'.format(ir.table, ir.model_filter(), ir.res_id)
-        cr.execute(query, [model, ids])
+        if ir.company_dependent_comodel:
+            if ir.company_dependent_comodel == model:
```
KangOl, line 414: We could win an indentation level by changing the condition.
This will also make the code clearer.
```python
if not ir.company_dependent_comodel:
    query = "DELETE FROM..."
    cr.execute(...)
elif ir.company_dependent_comodel == model:
    query = format_query(...)
    cr.execute(...)
```

src/util/records.py
```diff
@@ -410,8 +410,33 @@ def remove_records(cr, model, ids):
...
+                cr.execute(
+                    cr.format_query(
```
KangOl, line 416:
```suggestion
                    format_query(
```

src/util/records.py
```diff
@@ -410,8 +410,33 @@ def remove_records(cr, model, ids):
     table = table_of_model(cr, model)
     cr.execute('DELETE FROM "{}" WHERE id IN %s'.format(table), [ids])
     for ir in indirect_references(cr, bound_only=True):
-        query = 'DELETE FROM "{}" WHERE {} AND "{}" IN %s'.format(ir.table, ir.model_filter(), ir.res_id)
-        cr.execute(query, [model, ids])
+        if ir.company_dependent_comodel:
+            if ir.company_dependent_comodel == model:
+                cr.execute(
```
KangOl, line 415: We could even explode this UPDATE query.
//cc @aj-fuentes

src/util/records.py
```diff
@@ -1440,6 +1465,22 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
     for ir in indirect_references(cr, bound_only=True):
         if ir.table in ignores:
             continue
+        if ir.company_dependent_comodel:
+            if ir.company_dependent_comodel == model_src:
+                assert model_src == model_dst
+                cr.execute(
```
KangOl, line 1471: Ditto. The query could b exploded.

KangOl (changes requested):

src/util/records.py
aj-fuentes: Indeed it must be exploded.

aj-fuentes (approved):

KangOl (approved):
@robodoo r+

PR #127, opened 2024-08-14
[ADD] util.invert_boolean_field

KangOl:
New helper for renaming and inverting the meaning of a boolean field.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -48,6 +48,7 @@ from .pg import (
     alter_column_type,
     column_exists,
     column_type,
+    explode_execute,
     explode_query_range,
     format_query,
     get_value_or_en_translation,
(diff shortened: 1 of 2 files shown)


KangOl:

KangOl:
@robodoo r+

malb-odoo:
Still not merged 😬

src/util/fields.py
```diff
@@ -580,6 +581,26 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
         rename_field(cr, inh.model, old, new, update_references=update_references, skip_inherit=skip_inherit)
 
 
+def invert_boolean_field(cr, model, old, new, skip_inherit=()):
+    """Rename a boolean field and invert its value."""
```
aj-fuentes, line 585: Could have longer docs https://www.odoo.com/documentation/17.0/developer/reference/upgrades/upgrade_utils.html#odoo.upgrade.util.fields.invert_boolean_field

PR #128, opened 2024-08-19
[FIX] util.get_admin_channel

KangOl:
Search for old XMLID. Avoid creating a second channel.

Oversight of afa3e94bb04e600babf56f987c048d7a42067719


KangOl:

KangOl:
@robodoo r+

PR #129, opened 2024-08-23
[ADD] util.inconsistencies.break_recursive_loops

KangOl:
Helper to break recursive loops (what a surprise!) in foreign keys on the same table, such as parent/child relations (menus, categories...) or logical dependencies (depending tasks...)

### DIFF
--- a/src/util/inconsistencies.py
+++ b/src/util/inconsistencies.py
@@ -5,11 +5,12 @@ from textwrap import dedent
 
 from psycopg2.extensions import quote_ident
 from psycopg2.extras import Json
+from psycopg2.sql import SQL
 
 from .helpers import _validate_model, table_of_model
 from .misc import chunks, str2bool
-from .pg import get_value_or_en_translation
-from .report import add_to_migration_reports
+from .pg import format_query, get_value_or_en_translation, target_of
+from .report import add_to_migration_reports, get_anchor_link_to_record, html_escape
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


aj-fuentes (approved):

src/util/inconsistencies.py
```diff
@@ -24,6 +23,99 @@
...
+          WHERE {field} IS NOT NULL
+       GROUP BY {field}
+          UNION ALL
+         SELECT child.{field} || curr.path AS path,
+                child.{field} = any(curr.path) AS cycle
+           FROM __loop AS curr
+           JOIN {table} AS child
+             ON child.id = curr.path[1]
+          WHERE child.{field} IS NOT NULL
+            AND NOT curr.cycle
```
aj-fuentes, line 51: Simpler?
```suggestion
         SELECT child.id || curr.path AS path,
                child.id = any(curr.path) AS cycle
           FROM __loop AS curr
           JOIN {table} AS child
             ON child.{field} = curr.path[1]
```

KangOl: The `AND NOT curr.cycle` is needed to actually stop the loop (because of the `UNION ALL`)

Same for the `child.{field} IS NOT NULL`, it skips the last hop in the tree.

KangOl:
@robodoo rebase-ff r+

src/util/inconsistencies.py
aj-fuentes: Yep, I removed too much in the where...

aj-fuentes: > Same for the child.{field} IS NOT NULL, it skips the last hop in the tree.

Since path has non-null values, `child.{field} = path[1]` should avoid the need to check non null field... but anyway it doesn't add much...

KangOl:
@robodoo r-

src/util/inconsistencies.py
aj-fuentes: I think your version is fine. You pick all parents with at least one child to start, then in the recursive step you take all parents of any record already in the table. Seems correct. My suggestion was in the other direction: to add children of what we already have. Both are equivalent I think.

aj-fuentes: ofc the removing the check for cycle was wrong in my suggestion

KangOl:
@robodoo r+

PR #131, opened 2024-08-28
[FIX] Preserve all translated fields on v16 migration

yajo:
Currently, the upgrade process preserves translations only for Odoo CE+EE fields. However, databases usually have more modules (or even custom fields) and those translations get lost when upgrading to Odoo 16.

With this script, all translated fields will inherit their translations in all languages. Since it is executed in the `pre` phase of `base`, it should be able to still find the `ir_translation` table.

@moduon MT-7120


aj-fuentes:
> However, databases usually have more modules (or even custom fields) and those translations get lost when upgrading to Odoo 16.

The table is renamed to _ir_translation and kept always. What translations are you trying to keep? Only entries with `type='code'` are cleaned before renaming the table.

NomadsComputerllc:
Merge to upgrade

those Errors are killing me

NomadsComputerllc (approved):
Thanks for the updates

NomadsComputerllc (approved):
Thanks

yajo:
After a migration, you can see an upgrade report like this:

image

It specifies that migrations _may be wrong_. However, in reality, _all those_ are wrong because only the base language (usually en_US) is migrated.

With this script, all langs are always migrated in all fields (declared by Odoo or not). Still, something may be wrong; but at least, most problems get auto-fixed now.

aj-fuentes:
Yes, you need to use [`_get_translation_upgrade_queries`](https://github.com/odoo/odoo/blob/e01df041b9428cf14cb150bdb0803c6f5ff2d80f/odoo/tools/translate.py#L1552) on your side (via an upgrade script) when updating the custom modules (with custom code available) on your upgraded DB.

That's the same mechanism we use in the standard upgrade script. We cannot do this for custom fields because we don't know if a field has a special `translate=` method. This option comes via code and we do not have the custom code. There are two standard options `xml_translate` and `html_translate`, but there could be even a custom translate.

What you could do is something like (beware: not tested!)
```py
    for model in your_custom_models:      
        if model._auto and not model._abstract:
            for field in model._fields.values():
                if field.store and field.translate:
                    migrate, cleanup = _get_translation_upgrade_queries(cr, field)
                    # run the queries in migrate, and cleanup
```

Side node: when `_get_translation_upgrade_queries` was introduced this repo didn't exist thus the way to share the queries with the community was via the main odoo repo.

yajo:
IIUC there shouldn't be an incompatibility between this PR and your suggestion, right? 🤔 

This PR only improves the default value for those fields. If there's more magic involved, then you can still do that and get to the appropriate fix.

aj-fuentes:
@yajo 
> IIUC there shouldn't be an incompatibility between this PR and your suggestion, right? 🤔

This change is risky because it will interfere with all the battle tested machinery of the translations in 15->16 upgrade. Including PO loading, noupdate flags, etc. Also for specific scripts of some clients.

> This PR only improves the default value for those fields. If there's more magic involved, then you can still do that and get to the appropriate fix.

If you want to contribute an improvement I think it would be better to add something that could be used directly in `end-` scripts for custom code (in [specific.py](https://github.com/odoo/upgrade-util/blob/master/src/util/specific.py)). That way the community could have a simpler alternative than figuring out how to use `_get_translation_upgrade_queries`. We could even add a reference to it in the upgrade report (like in the example picture you sent before).

@KangOl could you give some insight here please?

yajo:
I changed it according to your suggestion. Could you please review again?

src/util/specific.py
```diff
@@ -137,3 +138,66 @@ def reset_cowed_views(cr, xmlid, key=None):
         [key, module, name],
     )
     return set(sum(cr.fetchall(), ()))
+
+
+def translation2jsonb_unofficial_fields(cr):
```
aj-fuentes, line 143: I'd suggest you restrict the fields to be upgraded. Or maybe allow also to restrict by models. It would avoid unintentional changes and be faster if you can callers pinpoint the fields that need a fix (maybe from the info in the upgrade report). 

Also assert that the running DB is at least in 16.0.

Something like:
```suggestion
def convert_custom_translated_fields(cr, models=None, fields=None):
    assert util.version_gte("16.0")
    assert bool(models) ^ bool(fields)
    if models:
        # fix all fields per model that look like they weren't yet updated
    else:
        # fix specific fields = [(model_name, field_name), ...]
```
You could identify if a field looks like it wasn't updated by inspecting its column. If the jsonb dict only has en_US and lacks some langs for which there are translations in _ir_translation then it's most probably something custom. Note that I still believe it's really better to restrict by specific fields instead of per model. I also understand that for huge custom models (with many fields) this may be cumbersome, hence the alternative to restrict by model.

FYI on custom fields we _only_ run `util.convert_field_to_translatable(cr, model_name, field)`.

src/util/specific.py
```diff
@@ -137,3 +138,66 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+        (
+            migrate_queries,
+            cleanup_queries,
+        ) = odoo.tools.translate._get_translation_upgrade_queries(cr, field)
+        all_migrate_queries.extend(migrate_queries)
+        all_cleanup_queries.extend(cleanup_queries)
+    for query in all_migrate_queries:
+        cr.execute(query)
+    for query in all_cleanup_queries:
+        cr.execute(query)
```
aj-fuentes, line 203: You can run the queries in parallel. Note that you cannot run migrate queries from different fields in parallel -- we know it causes issues ;)
```suggestion
        util.parallel_execute(cr, migrate_queries)
        all_cleanup_queries.extend(cleanup_queries)
    
    util.parallel_execute(cr, all_cleanup_queries)
```

src/util/specific.py
```diff
@@ -137,3 +138,66 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+        try:
+            model = env[ir_field.model_id.model]
+            field = model._fields[ir_field.name]
+        except KeyError:
+            _logger.warning(
+                "Field %s in model %s is not loaded; skipping",
+                ir_field.name,
+                ir_field.model_id.model,
+            )
+            continue
```
aj-fuentes, line 185: This suggestion is very close to how we do in our upgrade script. With possible alternatives to narrow down what to update.
```suggestion
    for Model in env.registry.values():
        if Model._auto and not Model._abstract: # perhaps: and Model._name in models:
            for field in Model._fields.values():
                if field.store and field.translate and not field.manual:                    # perhaps: and (Model._name, field.name) in fields
                   
```

src/util/specific.py
```diff
@@ -137,3 +138,66 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+    It will find all those fields and convert translations from the
+    ``_ir_translation`` table into the new JSONB format.
+
+    Any fields not loaded will be skipped.
+    """
+    cr.execute(
+        """
+        SELECT ARRAY_AGG(name)
+        FROM ir_module_module
+        WHERE author LIKE 'Odoo S.A.'
```
aj-fuentes, line 159: This is brittle. It's better if you fully restrict the fields you need to upgrade.

aj-fuentes (commented):
The idea looks correct, thanks!

I left some comments. In summary, IMO the important part of the suggestions is to allow to narrow down what needs to be updated as parameter(s) to the function.

KangOl:

src/util/specific.py
aj-fuentes: > Maintaining a hardcoded list of modules, models or fields to translate? No, thanks. Everything is already in the database. At least, everything we care about (user-generated translations).

If you can be sure you can rely on the info the DB to pick the data then this is fine. In our experience relying only on the DB is error prone, and fails for many (common) edge cases (like wrong data in the DB). All out internal upgrade scripts are specific about modules or fields updates. An upgrade script is something you write once for your module and should be stable between major versions. I don't see any problem in having a full list of models and fields you need to update. It would be exceptional to have hundreds of **translated** fields in a module, and even in such case having a script that you write once is not a problem.

I understand that what you are trying to add here is a "general" function to update all fields. My point is that it's better to add the upgrade script to each module in the right folder. It's each module's responsibility to update their data. In short, I think `translation2jsonb` is fine, `translation2jsonb_all_missing` is a bit problematic.

Note: I didn't check the code yet, this is a reply to your comment only.

yajo: It depends on your use case.

You guys massively migrate thousands of DBs per day, with no control over the input but full control over the migration scope (Odoo CE+EE+design-themes+industry addons; DB schema frozen at launch day). I understand it makes sense for you indeed.

We are more in the retail side. We do have control over the input, so if a translation is wrong there, we can just change it there and migrate again. But we have limited control over migration scope (a new customer can hire us, with an old Odoo installation with a big amount of custom modules; some can be communitary and still get updates 10 years later).

So, I could remove `translation2jsonb_all_missing` from the PR, but at the end most out-of-Odoo people that will use upgrade-util and `translation2jsonb` will write something similar on their own. You choose, anyway.

FWIW, [OpenUpgrade does maintain a list of excluded fields](https://github.com/OCA/OpenUpgrade/blob/2e9d96f1288596f17d5af3fad84b86eca6dc3821/openupgrade_scripts/scripts/base/16.0.1.3/end-migration.py#L9-L39). That's IMHO better suited for downstream scripts.

aj-fuentes: This is a nice perspective. We are all discovering the best approach for this repo. Thanks for the details. For us it's a bit hard to foresee the community usage and so we will try to gain insights like this one.

For the moment I think this patch looks good. I may have some nitpick comments (still review in progress) but it looks good enough.

aj-fuentes (commented):
Some nitpicks

yajo:
All attended.

aj-fuentes:
Could you please rebase your fork's branch in the latest master of this repo? The CIs are all red due to this.

aj-fuentes:
upgradeci retry

yajo:
Rebased

src/util/specific.py
```diff
@@ -137,3 +139,81 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+    into the new JSONB format.
+
+    :param odoo.fields.Field fields: Fields to convert.
+    :return: Converted fields.
+    """
+    if not version_gte("16.0"):
+        raise UpgradeError("JSONB translations are only available from Odoo 16")
+    all_cleanup_queries = []
+    for field in fields:
+        _logger.info("Migrating translations for field %s in model %s", field.name, field.model_name)
```
KangOl, line 211: Won't it be a little bit too verbose to log each field?
What about only log all fields all at once?
```suggestion
    _logger.info("Migrating translations for fields %r, fields)
    for field in fields:
```

src/util/specific.py
```diff
@@ -137,3 +139,81 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+
+    It will find all unofficial fields and convert translations from the
+    ``_ir_translation`` table into the new JSONB format.
+
+    Modules must be available in the environment to be loaded.
+
+    :return: Converted fields.
+    """
+    if not version_gte("16.0"):
+        raise UpgradeError("JSONB translations are only available from Odoo 16")
```
KangOl, line 159: Handle the case where the `_ir_translation` has been removed (once all fields have been converted, we can drop this table).
```suggestion
    if not version_gte("16.0"):
        raise UpgradeError("JSONB translations are only available from Odoo 16")
    if not table_exists(cr, "_ir_translation"):
        _logger.getChild("translation2jsonb_all_missing").info("Missing `_ir_translation` table; skip.")
        return
```

src/util/specific.py
```diff
@@ -1,11 +1,13 @@
...
 from .helpers import _validate_table
-from .misc import _cached
+from .misc import _cached, version_gte
 from .models import rename_model
-from .modules import rename_module
-from .pg import column_exists, rename_table, table_exists
+from .modules import odoo, rename_module
+from .orm import env
+from .pg import column_exists, format_query, parallel_execute, rename_table, table_exists
 from .report import add_to_migration_reports
```
KangOl, line 11: Import the `_get_translation_upgrade_queries` function directly from `odoo` module.
```suggestion
from .modules import rename_module
from .orm import env
from .pg import column_exists, format_query, parallel_execute, rename_table, table_exists
from .report import add_to_migration_reports

try:
    from odoo.tools.translate import _get_translation_upgrade_queries
except ImportError:
    if version_gte("16.0"):
        raise
```

src/util/specific.py
```diff
@@ -137,3 +139,81 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+    """
+    if not version_gte("16.0"):
+        raise UpgradeError("JSONB translations are only available from Odoo 16")
+    all_cleanup_queries = []
+    for field in fields:
+        _logger.info("Migrating translations for field %s in model %s", field.name, field.model_name)
+        (
+            migrate_queries,
+            cleanup_queries,
+        ) = odoo.tools.translate._get_translation_upgrade_queries(cr, field)
```
KangOl, line 215:
```suggestion
        ) = _get_translation_upgrade_queries(cr, field)
```

KangOl (changes requested):

src/util/specific.py
yajo: I don't think it's a good idea. This function is only needed for 16.0+, so raising an `ImportError` in lower versions that import this file would be harmful.

With the current implementation, if some of those older versions happen to use the new `translation2jsonb` function, only then they'd fail. That's better IMHO.

src/util/specific.py
yajo: Not applying as explained above.

src/util/specific.py
KangOl: It's still the case. It only reraises the `ImportError` if the version is *older* than 16.

And this is exactly what I want to catch. If the `_get_translation_upgrade_queries` is removed[^1] from odoo (like it nearly happen in [I can't find back the odoo PR]), I want the CI to directly fail (the new functions are not covered by the CI).


[^1]: or renamed or moved.

yajo: Yikes, true!

src/util/specific.py
```diff
@@ -156,3 +164,84 @@ def reset_cowed_views(cr, xmlid, key=None):
...
+    """
+    if not version_gte("16.0"):
+        raise UpgradeError("JSONB translations are only available from Odoo 16")
+    all_cleanup_queries = []
+    _logger.info("Migrating translations for fields %r", fields)
+    for field in fields:
+        (
+            migrate_queries,
+            cleanup_queries,
+        ) = _get_translation_upgrade_queries(cr, field)
```
KangOl, line 243: Strange formatting
```suggestion
        migrate_queries, cleanup_queries = _get_translation_upgrade_queries(cr, field)
```

yajo:
How's this going?

yajo:
Hello there! Is there intention to merge this? Thanks!

aj-fuentes (approved):
@robodoo r+

Thanks.

PR #132, opened 2024-09-02
[IMP] core: deprecate odoo.registry()

kmagusiak:
odoo/odoo#178784
odoo/enterprise#69269
task-4069446


KangOl:
upgradeci retry with alway only hr

KangOl (approved):
@robodoo r+

PR #134, opened 2024-09-10
[FIX] adapt code and test to renaming of <tree> to <list> in Odoo 18

KangOl:
See odoo/odoo#159909


KangOl:

PR #135, opened 2024-09-11
[IMP] remove and uninstall modules and themes recursively

yajo:
Current implementation won't uninstall/remove dependencies. This can easily lead to a broken database, and is an unexpected and important divergence compared to the normal ORM uninstallation or removal process.

@moduon MT-7110

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -98,16 +98,44 @@ def module_installed(cr, module):
     return modules_installed(cr, module)
 
 
-def uninstall_module(cr, module):
+def module_dependencies(cr, module):
+    """Get dependencies of given module.
+
+    :param str module: name of the module
+    :return: list names of the dependencies
+    :rtype: list(str)
+    """
+    cr.execute(
+        """
+        SELECT m.name
+        FROM ir_module_module m
+        INNER JOIN ir_module_module_dependency d ON d.module_id = m.id
+        WHERE d.name = %s
+        """,
+        [module],
+    )
+    return [name for (name,) in cr.fetchall()]
+
+
+def uninstall_module(cr, module, with_dependencies=False):
     """
     Uninstall and remove all records owned by a module.
 
     :param str module: name of the module to uninstall
+    :param bool with_dependencies: whether to also remove dependencies of the module
+    :return: set of uninstalled module names
+    :rtype: set(str)
     """
+    result = set()
     cr.execute("SELECT id FROM ir_module_module WHERE name=%s", (module,))
     (mod_id,) = cr.fetchone() or [None]
     if not mod_id:
-        return
+        return result
+
+    if with_dependencies:
+        dependencies = module_dependencies(cr, module)
+        for dep in dependencies:
+            result.union(uninstall_module(cr, dep, with_dependencies=with_dependencies))
 
     # delete constraints only owned by this module
     cr.execute(
@@ -219,14 +247,18 @@ def uninstall_module(cr, module):
     if table_exists(cr, "ir_translation"):
         cr.execute("DELETE FROM ir_translation WHERE module=%s", [module])
     cr.execute("UPDATE ir_module_module SET state='uninstalled' WHERE name=%s", (module,))
+    return result | {module}
 
 
-def uninstall_theme(cr, theme, base_theme=None):
+def uninstall_theme(cr, theme, base_theme=None, with_dependencies=False):
     """
     Uninstall a theme module and remove it from websites.
 
     :param str theme: name of the theme module to uninstall
     :param str or None base_theme: if not `None`, unload first this base theme
+    :param bool with_dependencies: whether to also remove dependencies of the theme
+    :return: set of uninstalled module names
+    :rtype: set(str)
 
     .. warning::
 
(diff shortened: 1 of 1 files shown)


aj-fuentes:
This option is not as useful as it looks --I'd say it is dangerous. In many cases when we remove a module all modules that depend on it need to be adapted, instead of removed. Some modules are even custom for which we do not have the code readily available. I think the best approach is to deal with this on a case by case basis, a loop over pre-computed modules --as you propose here-- in _client_ code is better IMO.

In this case the deviation of the ORM behavior is intentional: we want to remove a module without removing the modules that depend on it.

For the ultimate opinion we need to summon @KangOl

src/util/modules.py
```diff
@@ -148,17 +148,50 @@ def module_installed(cr, module):
...
+            """
+            SELECT ARRAY_AGG(DISTINCT m.name ORDER BY m.name)
+            FROM ir_module_module m
+            INNER JOIN ir_module_module_dependency d ON d.module_id = m.id
+            WHERE d.name = ANY(%s)
+            """,
+            (found + [module],),
+        )
+        new = cr.fetchone()[0]
+        if new == found:
```
EmilioPascual:
```suggestion
        if not new or new == found:
```
If module hasn't dependencies, then `new = None` and you have an error like that:
```
Error: unsupported operand type(s) for +: 'NoneType' and 'list'
```

yajo: Thanks, it should be fixed now.

I also optimized it by issuing a single SQL query and renamed the method because `module_dependencies` seemed like it would return the dependencies of the module, and not the dependent modules. Now the method is called `dependent_modules`.

src/util/modules.py
```diff
@@ -148,17 +148,54 @@ def module_installed(cr, module):
...
+        WITH RECURSIVE dependents AS (
+            SELECT m.name
+            FROM ir_module_module m
+            JOIN ir_module_module_dependency d ON d.module_id = m.id
+            WHERE d.name = %s
+            UNION
+            SELECT m2.name
+            FROM ir_module_module m2
+            JOIN ir_module_module_dependency d2 ON d2.module_id = m2.id
+            JOIN dependents dep ON d2.name = dep.name
```
aj-fuentes: The format we expect in this repo:
```suggestion
            SELECT m.name
              FROM ir_module_module m
              JOIN ir_module_module_dependency d 
                ON d.module_id = m.id
             WHERE d.name = %s
            
             UNION
             
            SELECT m2.name
              FROM ir_module_module m2
              JOIN ir_module_module_dependency d2
                ON d2.module_id = m2.id
              JOIN dependents dep
                ON d2.name = dep.name
```

src/util/modules.py
```diff
@@ -148,17 +148,54 @@ def module_installed(cr, module):
...
+        """,
+        (module,),
+    )
+    result = cr.fetchone()[0]
+    return result or []
+
+
 @_warn_usage_outside_base
-def uninstall_module(cr, module):
+def uninstall_module(cr, module, with_dependents=False):
```
aj-fuentes: Keep this function and the rest of this file untouched.

src/util/modules.py
```diff
@@ -148,17 +148,54 @@ def module_installed(cr, module):
...
+def dependent_modules(cr, module):
+    """Get all modules that depend on the given module.
+
+    :param str module: name of the module
+    :return: list of names of the dependent modules
+    :rtype: list(str)
+    """
+    cr.execute(
+        """
+        WITH RECURSIVE dependents AS (
```
aj-fuentes, line 161: You should filter the returned modules by their state --use `util.INSTALLED_MODULE_STATES`-- it makes little sense to return uninstalled modules in our context. The docstring should be adapted to reflect this.

src/util/modules.py
```diff
@@ -148,6 +148,42 @@ def module_installed(cr, module):
     return modules_installed(cr, module)
 
 
+def dependent_modules(cr, modules, states=INSTALLED_MODULE_STATES):
```
aj-fuentes, line 151:
```suggestion
def dependent_modules(cr, modules):
```

src/util/modules.py
```diff
@@ -148,6 +148,42 @@ def module_installed(cr, module):
...
+    """Return all modules that depend (directly or indirectly) on the given module(s).
+
+    :param tuple(str) modules: name or tuple of names of the module(s)
+    :param tuple(str) states: return modules only in these states (default: INSTALLED_MODULE_STATES)
```
aj-fuentes, line 155:
```suggestion
```

src/util/modules.py
```diff
@@ -148,6 +148,42 @@ def module_installed(cr, module):
...
+              FROM ir_module_module m2
+              JOIN ir_module_module_dependency d2
+                ON d2.module_id = m2.id
+              JOIN dependents dep
+                ON d2.name = dep.name
+             WHERE m2.state IN %(states)s
+        )
+        SELECT ARRAY_AGG(DISTINCT name ORDER BY name) FROM dependents
+        """,
+        {"modules": modules, "states": states},
```
aj-fuentes, line 181:
```suggestion
        {"modules": modules, "states": INSTALLED_MODULE_STATES},
```

src/util/modules.py
```diff
@@ -148,6 +148,42 @@ def module_installed(cr, module):
     return modules_installed(cr, module)
 
 
+def dependent_modules(cr, modules, states=INSTALLED_MODULE_STATES):
+    """Return all modules that depend (directly or indirectly) on the given module(s).
```
aj-fuentes, line 152:
```suggestion
    """Return all installed modules that depend (directly or indirectly) on the given module(s).
```

PR #136, opened 2024-09-13
[FIX] util/orm: Disable only custom module record rule

sagu-odoo:
db have record rule which are created from interface or either standard record rule external_id is missing got disable during neturlize(run_sanitizer) this  can block the upgrade process due to access error but actually with that  record rule  database is running fine but during upgrade after neutrlize upgrade process is blocked.So, better to disable only custom module record rule.

```
 count
-------
     4
(1 row)

```
after query result
```
 count
-------
     0
(1 row)
```
error message
```
Due to security restrictions, you are not allowed to modify 'Journal Entry' (account.move) records.

Records: Draft Invoice INV/2024/00001 (id=25)
User: Mitchell Admin (id=2)

This restriction is due to the following rules:
- Point Of Sale Account move

Contact your administrator to request access if necessary.
```

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -612,7 +612,7 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
         WITH custom_rules AS (
                 SELECT r.id
                   FROM ir_rule r
-             LEFT JOIN ir_model_data d ON d.model = 'ir.rule' AND d.res_id = r.id
+                  JOIN ir_model_data d ON d.model = 'ir.rule' AND d.res_id = r.id
                  WHERE COALESCE(d.module, '') NOT IN %s
                    AND r.active
                  )


KangOl:

KangOl:
No, this is wrong. You only disable rules that have an XMLID.

Can you please link to the failing request?

sagu-odoo:
[request link](https://upgrade.odoo.com/web#id=2037504&cids=1&menu_id=107&action=150&model=upgrade.request&view_type=form) is no more blocking wrote the specific fix for this actually
issue was that client have 3 record rule which don't have xmlid and while connecting with upgrade start  that record rule become disable and error is coming from our side  and in their production environment it was working fine. After many back and forth I figured out due to the query record rule was disabled  and in the end testcases also failing.

sagu-odoo:
Hello @KangOl , could you please review this pr ?. 

cc @aj-fuentes

aj-fuentes:
In principle I think the idea is correct. We should be careful though since switching this could potentially make current DONE requests to fail. Can we analyze more cases to ensure we take the risk for something with enough impact?

sagu-odoo:
@aj-fuentes  thanks for quick review.

Issues is not with only blocking request also with as we connect with upgrade start record rule created from ui becomes  disable and acess error raised weather client customization is appropriate according to their business flow and here is no error this can create a chaos that is also the main reason to propose this fix. 

> Can we analyze more cases to ensure we take the risk for something with enough impact?

As I have checked their will be case. if client override the standard record rule by  like domain ``[(1, '=', 1)]`` or any other domain disabling that can cause the issue and as we  draw line for custom module. so we don't need deactivate the custom record rule create from ui 

once again thanks >,<

KangOl:
This is wrong. We deactivate those rules because they may refer to fields defined by custom modules (even if they don't have a xmlid). We must deactivate all custom rules.

The failures when loading UI are unrelated as we reactivate those custom rules at the end of this contextmanager.

aj-fuentes:
> Issues is not with only blocking request also with as we connect with upgrade start record rule created from ui becomes disable and acess error raised weather client customization is appropriate according to their business flow and here is no error this can create a chaos that is also the main reason to propose this fix.

I understood the issue. My point is that making extra requests fail for something happening in _one_ request is not a good approach. Starting a DB can fail for many reasons. I don't agree that it's a "chaos". If a DB doesn't start then the solution is to do what you did: investigate and fully understand why it doesn't start. Note that your patch could equally make currently starting DBs to stop starting, if now we stop disabling a rule for example that was allowing the DB to start. Thus whatever we do, some DBs will still not start and the issue would have to be debugged anyway.

sagu-odoo:
Thanks for your valuable input @aj-fuentes and @KangOl . your points are strong and understandable. True issue should be debugged anyway.

aj-fuentes:
One clarification:

> The failures when loading UI are unrelated as we reactivate those custom rules at the end of this contextmanager.

We don't reactivate the rules when we start the DBs in test mode (`rollback=False`)
https://github.com/odoo/upgrade-platform/blob/7b3d09b2f53273663170849523564b6de0978fce/upgrade-docker/addons/runsanitizer/__init__.py#L32
Run as a hook https://github.com/odoo/upgrade-platform/blob/7b3d09b2f53273663170849523564b6de0978fce/upgrade-docker/addons/runsanitizer/__init__.py#L36
https://github.com/odoo/upgrade-platform/blob/7b3d09b2f53273663170849523564b6de0978fce/upgrade-docker/main.py#L517
 
---

I still agree that those rules must be disabled to avoid more common errors. The case put forward here as justification for the patch looks like a specific case that can be solved by debugging the issue. I understand that it may cause failures in some tests, but in very specific cases (heavily customized rules). In which cases it's always better to check what's going on, even if it may be a bit hard. Moreover the case put as an example was already solved with a specific script. Thus it proves the point.

PR #137, opened 2024-09-13
[IMP] snippets: do not return unneeded data from mp worker

cawo-odoo:
This PR is opened containing 2 commits, of which only one is to be chosen by preference (mine is the 2nd :smile:).

**[IMP] snippets: do not return unneeded data from mp worker**

In `convert_html_columns()`, we select 100MiB worth of DB tuples and pass them to a ProcessPoolExecutor together with a converter callable. So far, the converter returns all tuples, changed or unchanged together with the information if it has changed something. All this is returned through IPC to the parent process. In the parent process, the caller only acts on the changed tuples, though, the rest is ignored. In any scenario I've seen, only a small proportion of the input tuples is actually changed, meaning that a large proportion is returned through IPC unnecessarily.

What makes it worse is that processing of the converted results in the parent process is often slower than the conversion, leading to two effects:
1) The results of all workers sit in the parent process's memory, possibly leading to MemoryError (upg-2021031)
2) The parallel processing is being serialized on the feedback, defeating a large part of the intended performance gains

This commit fixes (1) by only returning `None` for unchanged tuples. An improvement for (2) follows in the next commit.

**[IMP] snippets: move all work from parent to mp workers**

In `convert_html_columns()`, we select 100MiB worth of DB tuples and pass them to a ProcessPoolExecutor together with a converter callable. So far, the converter returns all tuples, changed or unchanged together with the information if it has changed something. All this is returned through IPC to the parent process. In the parent process, the caller only acts on the changed tuples, though, the rest is ignored. In any scenario I've seen, only a small proportion of the input tuples is actually changed, meaning that a large proportion is returned through IPC unnecessarily.

What makes it worse is that processing of the converted results in the parent process is often slower than the conversion, leading to two effects:
1) The results of all workers sit in the parent process's memory, possibly leading to MemoryError (upg-2021031)
2) The parallel processing is being serialized on the feedback, defeating a large part of the intended performance gains

To improve this, this commit
- moves all work into the workers, meaning not just the conversion filter, but also the DB query as well as the DB updates.
  - by doing so reduces the amount of data passed by IPC to just the query texts
  - by doing so distributes the data held in memory to all worker processes
- reduces the chunk size by one order of magnitude, which means
  - a lot less memory used at a time
  - a lot better distribution of "to-be-changed" rows when these rows are clustered in the table

All in all, in my test case, this
- reduces maximum process size in memory to 300MiB for all processes compared to formerly >2GiB (and MemoryError) in the parent process
- reduces runtime from 17 minutes to less than 2 minutes


cawo-odoo:
ok, I stopped the *matt*ness. If anyone wants to review ... :-)

src/base/tests/test_util.py
```diff
@@ -1436,6 +1437,8 @@ def not_doing_anything_converter(el):
 
 class TestHTMLFormat(UnitTestCase):
     def testsnip(self):
+        cr = self.registry.cursor()
```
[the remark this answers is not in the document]
cawo-odoo: >  Is it because now you commit the cursor?

Yes!

> IIRC we have other examples in testing utils were we instead remove the commit restriction from test cursors.

True, but this approach only works with threads, not with forked processes, I think. Also, the approach I chose is the more common approach looking at tests in the `odoo` repo. It also feels cleaner to me, personally.

src/util/snippets.py
```diff
@@ -243,31 +249,39 @@ def _dumps(self, node):
...
-                if has_changed:
-                    for lang, value in content.items():
-                        _, new_content[lang] = converter_callback(value)
-                new_content = Json(new_content)
-            else:
-                has_changed, new_content = converter_callback(content)
-            changes[column] = new_content
-            if has_changed:
-                changes["id"] = res_id
-        return changes
```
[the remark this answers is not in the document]
cawo-odoo: I did, but I couldn't bring myself to make it procedural, so it's probably still a bit more diff than you intended.

src/util/snippets.py
```diff
@@ -243,16 +249,26 @@ def _dumps(self, node):
...
+    def __call__(self, query):
+        with db_connect(self.dbname).cursor() as cr:
+            cr.execute(query)
+            for self.row in cr.fetchall():
+                self._convert_row()
+                if "id" in self.row:
+                    cr.execute(self.update_query, self.row)
 
-    def __call__(self, row):
+    def _convert_row(self):
```
aj-fuentes, line 266:
```suggestion
    def _convert_row(self, row):
```

src/util/snippets.py
```diff
@@ -243,16 +249,26 @@ def _dumps(self, node):
...
         converters = self.converters
         columns = self.converters.keys()
         converter_callback = self.callback
-        res_id, *contents = row
-        changes = {}
+        res_id, *contents = self.row
+        self.row = {}
```
aj-fuentes, line 271:
```suggestion
        changes = {}
        res_id, *contents = row
```

src/util/snippets.py
```diff
@@ -264,10 +280,9 @@ def __call__(self, row):
                 new_content = Json(new_content)
             else:
                 has_changed, new_content = converter_callback(content)
-            changes[column] = new_content
+            self.row[column] = new_content
```
aj-fuentes, line 283:
```suggestion
            changes[column] = new_content
```

src/util/snippets.py
```diff
@@ -264,10 +280,9 @@ def __call__(self, row):
...
             if has_changed:
-                changes["id"] = res_id
-        return changes
+                self.row["id"] = res_id
```
aj-fuentes, line 285:
```suggestion
                changes["id"] = res_id
        return changes if "id" in changes else None
```

src/util/snippets.py
```diff
@@ -243,16 +249,26 @@ def _dumps(self, node):
...
+        self.dbname = dbname
+        self.update_query = update_query
+
+    def __call__(self, query):
+        with db_connect(self.dbname).cursor() as cr:
+            cr.execute(query)
+            for self.row in cr.fetchall():
+                self._convert_row()
+                if "id" in self.row:
+                    cr.execute(self.update_query, self.row)
```
aj-fuentes, line 264: I'd say the idea is to pass the row and return the changes, if any.
```suggestion
            for changes in filter(None, map(self._convert_row, cr.fetchall())):
                cr.execute(self.update_query, changes)
```

aj-fuentes (commented):
Some suggestion about the row processing method. Untested!

Regarding the test I'm OK with it, but you could put a comment to somehow clarify the extra steps in the test.

src/util/snippets.py
cawo-odoo: > I'd say the idea is to pass the row and return the changes, if any.

That's what I meant by "procedural". Passing data back and forth through arguments and return values between two methods of a class feels just icky.

aj-fuentes: > That's what I meant by "procedural". Passing data back and forth through arguments and return values between two methods of a class feels just icky.

You can declare the method `@staticmethod` or `@classmethod` if that feels better. The possibility to override the processing is an advantage of keeping it a method. The latter is not necessary right now. To me the best value here is to clearly separate what is actually processing _one_ row, vs the underlying machinery to fetch/send data from/to the DB. The simpler diff is just a nice thing.

cawo-odoo:
all comments resolved

src/util/snippets.py
```diff
@@ -11,6 +12,11 @@
 from psycopg2.extensions import quote_ident
 from psycopg2.extras import Json
 
+try:
+    from odoo.sql_db import db_connect
+except ImportError:
+    from openerp.sql_db import db_connect
```
aj-fuentes, line 18: You can fully ignore openerp cases, this script won't work in old code anyway.
```suggestion
    pass
```

aj-fuentes (approved):
cc: @KangOl

src/util/snippets.py
```diff
@@ -243,11 +249,19 @@ def _dumps(self, node):
...
         self.converters = converters
         self.callback = callback
+        self.dbname = dbname
+        self.update_query = update_query
+
+    def __call__(self, query):
+        with db_connect(self.dbname).cursor() as cr:
+            cr.execute(query)
+            for changes in filter(None, map(self._convert_row, cr.fetchall())):
+                cr.execute(self.update_query, changes)
```
aj-fuentes, line 262: We need to ensure we close the connection. Also I think we are missing a commit for the cursor 🤔 (maybe there is some sort of autocommit logic I'm missing here).
```suggestion
        conn = db_connect(self.dbname)
        updated = False
        with conn, conn.cursor() as cr:
            cr.execute(query)
            for changes in filter(None, map(self._convert_row, cr.fetchall())):
                updated = True
                cr.execute(self.update_query, changes)
            if updated:
                cr.commit()
        conn.close()
```

cawo-odoo: The contextmanager commits and closes the cursor.
https://github.com/odoo/odoo/blob/17.0/odoo/sql_db.py#L164-L172
I do not want to close the connection, here. It will be added to the pool and re-used upon each call to `__call__()`, i.e. each further query assigned to the worker process. The connection socket is destroyed along with the filehandle along with the child process.
I would have liked to do that explicitly, but the `ProcessPoolExecutor` is limited in that it has an api hook/callback only for initialization code,  not for destruction code.

cawo-odoo:
𓃠 ?

cawo-odoo:
🐈 ?

aj-fuentes (approved):

cawo-odoo:
@nseinlet concerning our earlier coffee-machine meeting: wdyt about the fixup just pushed?

cawo-odoo:
> @nseinlet concerning our earlier coffee-machine meeting: wdyt about the fixup just pushed?

Today I improved the comments about the backwards compatibility, rebased, squashed the fixups. Do I get a review? And/or a :black_cat: ?

src/base/tests/test_util.py
```diff
@@ -1956,33 +1956,37 @@ def not_doing_anything_converter(el):
...
-            snippets.html_converter(
-                not_doing_anything_converter, selector="//*[hasclass('fake_class_not_doing_anything')]"
-            ),
-        )
-        util.invalidate(view_id)
-        res = self.env["ir.ui.view"].search_read([("id", "=", view_id.id)], ["arch_db"])
+        # util.convert_html_columns() commits the cursor, use a new transaction to not mess up the test_cr
+        with self.registry.cursor() as cr:
+            env = api.Environment(cr, SUPERUSER_ID, {})
+            view_arch = """
```
aj-fuentes: Move this out of the context manager.

src/base/tests/test_util.py
```diff
@@ -1956,33 +1956,37 @@ def not_doing_anything_converter(el):
...
+            """
+            view_id = env["ir.ui.view"].create(
+                {
+                    "name": "not_for_anything",
+                    "type": "qweb",
+                    "mode": "primary",
+                    "key": "test.htmlconvert",
+                    "arch_db": view_arch,
+                }
+            )
```
aj-fuentes: You can keep part of this also out of the context manager
```suggestion
            view_id = env["ir.ui.view"].create({"arch_db": view_arch, **vals})
```
then outside:
```py
            vals = {
                "name": "not_for_anything",
                "type": "qweb",
                "mode": "primary",
                "key": "test.htmlconvert",
            }            
```

src/util/snippets.py
```diff
@@ -305,17 +323,25 @@ def convert_html_columns(cr, table, columns, converter_callback, where_column="I
...
+        for future in log_progress(
+            concurrent.futures.as_completed(futures),
+            logger=_logger,
+            qualifier=f"{table} updates",
+            size=len(split_queries),
+            estimate=False,
+            log_hundred_percent=True,
+        ):
+            # just for raising any worker exception
+            future.result()
```
aj-fuentes, line 326: Do we need this? Is `map` plus consuming via `deque` enough?

src/util/snippets.py
```diff
@@ -243,11 +248,24 @@ def _dumps(self, node):
...
+    def __init__(self, converters, callback, dbname=None, update_query=None):
         self.converters = converters
         self.callback = callback
+        self.dbname = dbname
+        self.update_query = update_query
+
+    def __call__(self, query):
+        # backwards compatibility: caller passes rows in the "query" argument and expects us to return converted rows
+        if not (self.dbname and self.update_query and isinstance(query, str)):
+            return self._convert_row(query)
```
aj-fuentes: We can make this simpler: if dbname we know is the new impl.
```suggestion
    def __call__(self, row_or_query):
        # backwards compatibility: caller passes rows in the "query" argument and expects us to return converted rows
        if not self.dbname:
            return self._convert_row(query)
```
I find this merged implementation unnecessary. The two interfaces cannot be used transparently by the same client code because we **need** the extra parameters in the new version.

In this case I think the good old OOP may be cleaner. For example, rename this class to `ConvertorNG` and code it for the query variant. Then make `Convertor` a subclass that keeps the old behavior.
```py
class ConvertorNG:
    def __init__(self, converters, callback, dbname, update_query):
       ...
    def __call__(self, query):
       # filter rows if they do not have id, instead of by None
    def _convert_row(self, row):
       # do not return None if no id, use original impl


class Convertor(ConvertorNG):
    """Deprecated kept for compatibility use ConvertorNG"""
    def __init__(self, converters, callback):
        super().__init__(converters, callback, None, None)
    def __call__(self, row):
        return self._convert_row(self, row)
```

cawo-odoo: I took the suggestion to make it simpler. Sub-classing may be perfectly clean, but it is also a lot of additional loc. Do we want that?

src/util/snippets.py
cawo-odoo: Do we *need* to re-do an approved PR to apply some new trick we found in the half-year it was lying around just to save **1** loc? Because that's essentially the only achievement we would get. Functionally, the two versions are pretty much the same.

aj-fuentes: The PR isn't merged yet... but still it's fine for me.
And yes in one year we can understand better what we do here.

src/util/snippets.py
aj-fuentes: > I took the suggestion to make it simpler. Sub-classing may be perfectly clean, but it is also a lot of additional loc. Do we want that?

TL;DR: I'm OK with both variants.

It's not about LOCs, it's about the clarity in the implementation. **If** somebody out there is using the old interface we don't break them. _We_ will always use the new one including this PR. IMO it isn't worth to have a double approach for us internally if we would never use the other alternative. I'm still OK with this variant of the code, but having to understand and read the "compatibility code" next time we need to debug the code is unnecessary if we can just move it out of the way.

IMO having a double use parameter `query_or_row` only makes sense if **we** will use both variants. Since we are not, I'd suggest a new class keeping the old one for compatibility. Note that we do that already in many other places, just that as function aliases, here we are modifying a callable class hence the subclass suggestion.

That being said. Do not change it just for me. Let's ping @KangOl for a final review :)

src/util/snippets.py
```diff
@@ -219,11 +224,24 @@ def _dumps(self, node):
...
 
 class Convertor:
-    def __init__(self, converters, callback):
+    def __init__(self, converters, callback, dbname=None, update_query=None):
         self.converters = converters
         self.callback = callback
-
-    def __call__(self, row):
+        self.dbname = dbname
+        self.update_query = update_query
```
aj-fuentes, line 232: I'd add this check here to avoid wrong usage of the `Convertor` class.
Note that in principle passing only dbname could be allowed even by CI because the error would happen only when running the update query. There are many ways in which we could end not updating anything.
Anyway, this can be merged. This is a nitipick
```suggestion
        self.update_query = update_query
        # when db_name is set update_query must be set also
        assert self.dbname is None or self.update_query is not None
```

aj-fuentes (approved):
We can move on. This looks good. A tiny nit.

src/util/snippets.py
cawo-odoo: Suggestion applied, made it an XOR though.

cawo-odoo:
Thanks for the latest reviews, they made me look at this again and I realised this was still missing a safety-measure to avoid the possibility of multiple child processes using the same DB connection and I added it.

src/util/snippets.py
```diff
@@ -219,11 +224,30 @@ def _dumps(self, node):
...
+        self.pool_reset = False
+
+    def __call__(self, row_or_query):
+        # backwards compatibility: caller passes rows and expects us to return them converted
+        if not self.dbname:
+            return self._convert_row(row_or_query)
+        # improved interface: caller passes a query for us to fetch input rows, convert and update them
+        if not self.pool_reset:
+            sql_db._Pool = None  # children cannot borrow from copies of the same pool, it will cause protocol error
+            self.pool_reset = True
```
KangOl: Damn! 
That's really low level.

I would prefer to have this "re-init" step be done by the `initializer` argument of the `ProcessPoolExecutor`.

cawo-odoo: I would, too, but:
https://github.com/odoo/upgrade-util/pull/320/files#diff-e83e04eeae2ce53657ba28fd8de2126f5f3ef6422a73066af36dcf9191de1278R347

EDIT: Hum, OK, `html_converter` is first used in scripts for 16.0, meaning python 3.8+, so I guess we could use `initializer` for this one.

cawo-odoo: So either we disable TestHTMLFormat for older versions, or we do not use `initializer`.

IOW: shall I squash or drop the latest "fixup!" commit?

Your call @KangOl :-)

PR #139, opened 2024-09-19
[FIX] util/fields: improve report to be more precise

bda-odoo:
So far we've been matching very broadly, now we only look for actual
references of the field, eg 'type',"type",_type,type. 
An example of what shouldn't be matched in tickets is that we have
to improve the report to be more precise and ignore them.

opw-4126696

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -19,6 +19,11 @@ from psycopg2 import sql
 from psycopg2.extras import Json
 
 try:
+    from odoo import modules
+except ImportError:
+    from openerp import modules
+
+try:
     from odoo import release
     from odoo.osv import expression
     from odoo.tools.misc import mute_logger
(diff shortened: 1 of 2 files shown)


aj-fuentes:
A regexp will never be a full solution to checking code, that's why we put "**may** need" in the report. For example in this case the problem is that the server action has this in [its code](https://github.com/odoo/odoo/blob/7e2bccef15ecde2a77a40d452ea168892daff5b0/addons/product/views/product_template_views.xml#L178-L183)
```py
action = {
    'name': 'Pricelist Report',
    'type': 'ir.actions.client',
    'tag': 'generate_pricelist_report',
    'context': ctx,
}
```
The `'type'` is a false match for this record, but it could be also a _good_ match, imagine a code like
```py
record.write = {'type': 'type_val'}
```
where `record` is of the same same model as the field that is being updated. Thus you cannot ignore those cases, we prefer to be sure we output all that needs to be checked, even if some checks are not necessary.

If any, what you could do is be more explicit in the report and say something like " ... if the server action is a standard one and you didn't edit it, you may ignore them".

bda-odoo:
@aj-fuentes Thank you for the valuable suggestion. I have applied your recommendations, and the report has been updated accordingly.

**Here is a visual capture.**
main

aj-fuentes:
LGTM but I'll let @KangOl decide on this ;)

KangOl:

KangOl:
> As suggested by @bouvyd we can match the server actions by ir_model_data check if their module is a standard one **and** the action is noupdate=True. If `False` we output in the logs. Untested:
> 
> ```sql
>         SELECT a.id, a.{name}
>           FROM ir_act_server a
>           JOIN ir_model_data d
>             ON d.res_id = a.id
>            AND d.model = 'ir.act.server' -- not sure about the model
>          WHERE a.state = 'code'
>            AND (a.code ~ %(old_pattern)s
>                 {col_prefix} OR a.condition ~ %(old)s
>                )
>            AND (  d.module NOT IN %s  -- non standard action
>                OR d.noupdate) -- standard but updated during the upgrade
> ```
> 
> passing a tuple with all standard modules `modules.get_modules()`.

Needs a `LEFT JOIN` to match server actions without an xmlid.

bda-odoo:
Hello @aj-fuentes 

Thank you for your valuable input on this PR. The suggested changes have been implemented.

**Here is a visual capture.**
image13

bda-odoo:

aj-fuentes:
It isn't perfect yet, apparently the `AND d.noupdate` is making us show the [`Base: Activate Private ...`](https://github.com/odoo/upgrade/blob/15e404059189d80afe33eb54d1a320403bc22188/migrations/base/saas~16.4.1.3/pre-20-private-address.py#L129) action that is a standard one. one option is to make that action noupdate=False. @KangOl wdyt?

src/util/fields.py
```diff
@@ -1149,17 +1154,23 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
     col_prefix = ""
     if not column_exists(cr, "ir_act_server", "condition"):
         col_prefix = "--"  # sql comment the line
-
+    standard_modules = modules.get_modules()
```
KangOl, line 1157:
```suggestion
    standard_modules = modules.get_modules() | {"__upgrade__"}
```

bda-odoo: @KangOl Thank you for your valuable time. The suggested changes have been implemented.

KangOl (approved):
@robodoo rebase-ff r+

PR #140, opened 2024-09-24
[IMP] util/modules: force install of new modules

jjmaksoud:
The method `force_install_module` only works when the module already exists in the database and changes its state immediately. For new modules this does not work unless the module is explicitly added, or we use the auto discovery methods instead.
A check `_assert_modules_exists` is also added to the force installation.

As a solution, the check for version is added to see if we can delay the force installation to `post-base` when the module is auto discovered, and the implementation is moved to `_force_install_module`. The calls for `force_install_module` are replaced in the following methods where the module is expected to exist:
- `merge_module`
- `new_module_dep`
- `trigger_auto_install`


KangOl:

src/util/modules.py
```diff
@@ -486,8 +486,19 @@ def force_install_module(cr, module, if_installed=None):
     :param str module: name of the module to install
     :param list(str) or None if_installed: only force the install when these modules are
                                            already installed
-    :return str: the *original* state of the module
```
KangOl, line 489: Even if *we* don't use it, that's unfortunate that the API change.

src/util/modules.py
```diff
@@ -486,8 +486,19 @@ def force_install_module(cr, module, if_installed=None):
...
     """
+    if version_gte("saas~14.5"):
+        # We must delay until the modules actually exists. They are added by the auto discovery process.
+        if not if_installed or modules_installed(cr, *if_installed):
+            ENVIRON["__modules_auto_discovery_force_installs"].add(module)
+        return None
+    return _force_install_module(cr, module, if_installed)
```
KangOl, line 495:
```suggestion
    else:
        _force_install_module(cr, module, if_installed)
```

KangOl (changes requested):

src/util/modules.py
jjmaksoud: are you saying we should not have this change?
if that's the case then adding the assertion `_assert_modules_exists` to the existing implementation alone will force us to use the auto_discovery api for new modules

src/util/modules.py
```diff
@@ -486,8 +486,19 @@ def force_install_module(cr, module, if_installed=None):
     :param str module: name of the module to install
     :param list(str) or None if_installed: only force the install when these modules are
                                            already installed
-    :return str: the *original* state of the module
     """
+    if version_gte("saas~14.5"):
+        # We must delay until the modules actually exists. They are added by the auto discovery process.
+        if not if_installed or modules_installed(cr, *if_installed):
+            ENVIRON["__modules_auto_discovery_force_installs"].add(module)
```
KangOl, line 493: I would also the auto discovery fail when asked to force install an unknown module (if not already done)

src/util/modules.py
```diff
@@ -882,6 +893,7 @@ def _trigger_auto_discovery(cr):
     # Use accumulated values for the auto_install and force_upgrade modules.
 
     force_installs = ENVIRON["__modules_auto_discovery_force_installs"]
+    _assert_modules_exists(cr, *force_installs)
```
KangOl, line 896: Not needed (and done too early).
This is done by the call to `_force_install_module` line 933.
```suggestion
```

src/util/modules.py
```diff
@@ -486,8 +486,19 @@ def force_install_module(cr, module, if_installed=None):
...
+        if not if_installed or modules_installed(cr, *if_installed):
+            ENVIRON["__modules_auto_discovery_force_installs"].add(module)
+    else:
+        _force_install_module(cr, module, if_installed)
+
+
+def _force_install_module(cr, module, if_installed=None):
+    # Low level implementation
+    # Needs the module to exist in the database
+    _assert_modules_exists(cr, module)
```
KangOl, line 501:
```suggestion
    _assert_modules_exists(cr, module, *if_installed or ())
```

src/util/modules.py
jjmaksoud: the condition on line 932 only installs modules that are in the graph and in the force_install list, so we will ignore any unknown modules

KangOl: True.
The call should actually be done after the processing of the graph.
We could also verify the keys of `ENVIRON["__modules_auto_discovery_force_upgrades"]`

src/util/modules.py
```diff
@@ -917,8 +929,11 @@ def _trigger_auto_discovery(cr):
             _set_module_countries(cr, module, countries)
             module_auto_install(cr, module, auto_install)
 
-        if module in force_installs:
-            force_install_module(cr, module)
+    if force_installs:
+        _assert_modules_exists(cr, *force_installs)
+
```
KangOl, line 934: Useless. The call to `_force_install_module` will do it.
```suggestion
```

KangOl (changes requested):
I would also add a call to `_assert_modules_exists` into the function `_force_upgrade_of_fresh_module`.

KangOl (approved):
@robodoo r+

KangOl:
@robodoo r-

On hold until some retro-compatibility questions are sorted out.

jjmaksoud:
well I also forgot about it ¯\\_(ツ)\_/¯

PR #141, opened 2024-09-24
[FIX] util.indirect_references: use table_name instead of model_name

Pirols:
Otherwise errors (such as the following) will ensue.

```
2024-09-24 12:11:34,164 2158 ERROR matt-purchase odoo.upgrade.base.tests.test_util: ERROR: TestRecords.test_rename_xmlid
  File "/upgrade-util/src/base/tests/test_util.py", line 1030, in test_rename_xmlid
    res = util.rename_xmlid(cr, "base.TX1", "base.TX2", on_collision="merge")
  File "/upgrade-util/src/util/records.py", line 751, in rename_xmlid
    replace_record_references(cr, (model, old_id), (model, new_id), replace_xmlid=False)
  File "/upgrade-util/src/util/records.py", line 1308, in replace_record_references
    return replace_record_references_batch(cr, {old[1]: new[1]}, old[0], new[0], replace_xmlid)
  File "/upgrade-util/src/util/records.py", line 1490, in replace_record_references_batch
    explode_execute(cr, query, table=ir.table)
  File "/upgrade-util/src/util/pg.py", line 352, in explode_execute
    explode_query_range(cr, query, table, alias=alias, bucket_size=bucket_size),
  File "/upgrade-util/src/util/pg.py", line 259, in explode_query_range
    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
  File "/tmp/tmpinhjowz1/odoo/pr/153463/odoo/sql_db.py", line 354, in execute
psycopg2.errors.UndefinedTable: relation "res.partner" does not exist
LINE 1: SELECT min(id), max(id) FROM "res.partner"
```

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager
 from operator import itemgetter
 
 import lxml
+from psycopg2 import sql
 from psycopg2.extras import Json, execute_values
 
 try:
@@ -414,28 +415,33 @@ def remove_records(cr, model, ids):
             query = 'DELETE FROM "{}" WHERE {} AND "{}" IN %s'.format(ir.table, ir.model_filter(), ir.res_id)
             cr.execute(query, [model, ids])
         elif ir.company_dependent_comodel == model:
+            json_path = cr.mogrify(
+                "$.* ? ({})".format(" || ".join(["@ == %s"] * len(ids))),
+                ids,
+            ).decode()
+
             query = cr.mogrify(
                 format_query(
                     cr,
                     """
                     UPDATE {table}
-                    SET {column} = (
-                        SELECT jsonb_object_agg(
-                            key,
-                            CASE
-                                WHEN value::int4 IN %s THEN NULL
-                                ELSE value::int4
-                            END)
-                        FROM jsonb_each_text({column})
-                    )
-                    WHERE {column} IS NOT NULL
-                    AND {column} @? %s
-                    AND {{parallel_filter}}
+                       SET {column} = (
+                            SELECT jsonb_object_agg(
+                                key,
+                                CASE
+                                    WHEN value::int4 IN %s THEN NULL
+                                    ELSE value::int4
+                                END)
+                              FROM jsonb_each_text({column})
+                           )
+                     WHERE {column} IS NOT NULL
+                       AND {column} @? {json_path}
                     """,
                     table=ir.table,
                     column=ir.res_id,
+                    json_path=sql.Literal(json_path),
                 ),
-                [ids, "$.* ? ({})".format(" || ".join(map("@ == {}".format, ids)))],
+                [ids],
             ).decode()
             explode_execute(cr, query, table=ir.table)
     _rm_refs(cr, model, ids)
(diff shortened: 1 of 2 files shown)


KangOl:
upgradeci retry with always only purchase

aj-fuentes (approved):
LGTM

KangOl:
There are still some weird errors: https://upgradeci.odoo.com/upgradeci/run/165581

Pirols:
> There are still some weird errors: https://upgradeci.odoo.com/upgradeci/run/165581

Yes, I'm debugging those.

aj-fuentes:
From DM discussion, these lines https://github.com/odoo/upgrade-util/blob/a402a8db925963992f5abb39f7d5a2ed893706b3/src/util/records.py#L1472-L1489

Should be something like:
```
                query = cr.mogrify(
                    format_query(
                        cr,
                        """
                        UPDATE {table}
                           SET {column} = (
                                   SELECT jsonb_object_agg(key, COALESCE(
                                                                   jsonb_object(%s::text, %s::text)->>value)::int,
                                                                   value::int
                                                                 ))
                                   FROM jsonb_each_text({column})
                               )
                         WHERE {column} IS NOT NULL
                           AND {column} @? %s
                           AND {{parallel_filter}}
                        """,
                        table=ir.table,
                        column=ir.res_id,
                    ),
                    [list(id_mapping.keys()), list(id_mapping.values(), "$.* ? ({})".format(" || ".join(map("@ == {}".format, id_mapping)))],
                ).decode()
```
The key idea: pass the keys and values from id_mapping, to be used as jsonb in the query. Avoiding in the process explicit jsonb dicts in the form `{"key1": value1, ...}` which causes issues for the query formatter.

src/util/records.py
```diff
@@ -1475,7 +1475,7 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
                         """
                         UPDATE {table}
                            SET {column} = (
-                                   SELECT jsonb_object_agg(key, COALESCE((%s::jsonb->>value)::int, value::int))
+                                   SELECT jsonb_object(ARRAY_AGG(key), %s)
```
aj-fuentes, line 1478:
```suggestion
                                   SELECT jsonb_object_agg(key, (jsonb_object(%s::text[], %s::text[])->>value)::int, value::int)
```

KangOl: Yes, this version avoids having literal `{` in the query.

Let's try it over my patch.

src/util/records.py
```diff
@@ -1469,23 +1475,28 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
...
+                ).decode()
+
                 query = cr.mogrify(
                     format_query(
                         cr,
                         """
                         UPDATE {table}
                            SET {column} = (
-                                   SELECT jsonb_object_agg(key, COALESCE((%s::jsonb->>value)::int, value::int))
+                                   SELECT jsonb_object_agg(key, COALESCE(((jsonb_object(%s::text[]))->>value)::int, value::int))
```
Pirols, line 1489:
```suggestion
                                   SELECT jsonb_object_agg(key, COALESCE(((jsonb_object(%s::text[]))->>value), value)::int)
```

Given that `NULL` can also be cast to `int`, we can do the casting directly on the output of `COALESCE`.

PR #142, opened 2024-09-26
[IMP] util.explode_query_range: replace problematic `parallel_filter`…

Pirols:
… formatting

The usage of `str.format` to inject the parallel filter used to explode queries is not robust to the presence of other curly braces. Examples:

1. `JSON` strings (typically to leverage their mapping capabilities):

see 79f3d7184285e39b04a6c032aec0e05de6565fe8, where a query had to be modified to accomodate that.

2. Hardcoded sets of curly braces:

```python
>>> "UPDATE t SET c = '{usage as literal characters}' WHERE {parallel_filter}".format(parallel_filter="…")
KeyError: 'usage as literal characters'
```

Which can be (unelegantly) solved adding even more braces, leveraging one side-effect of `.format`:

```python
>>> "UPDATE t SET c = '{{usage as literal characters}}' WHERE {parallel_filter}".format(parallel_filter="…")
"UPDATE t SET c = '{usage as literal characters}' WHERE …"
```

3. Hardcoded curly braces (AFAICT no way to solve this):

```python
>>> "UPDATE t SET c = 'this is an open curly brace = {' WHERE {parallel_filter}".format(parallel_filter="…")
ValueError: unexpected '{' in field name
```

```python
>>> "UPDATE t SET c = 'this is a close brace = }' WHERE {parallel_filter}".format(parallel_filter="…")
ValueError: Single '}' encountered in format string
```


src/util/pg.py
```diff
@@ -311,7 +312,7 @@ def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_log
     Execute a query in parallel.
 
     The query is split by buckets of ids, then processed in parallel by workers. If the
-    query does not include the special `{parallel_filter}` value, it is added to the last
+    query does not include the special `PARALLEL_FILTER` value, it is added to the last
```
aj-fuentes, line 315: This is documentation, it should stay as before.

src/util/pg.py
```diff
@@ -297,10 +298,10 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, prefix=
         # Still, since the query may only be valid if there is no split, we force the usage of `prefix` in the query to
         # validate its correctness and avoid scripts that pass the CI but fail in production.
         parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
-        return [query.format(parallel_filter=parallel_filter)]
+        return [query.replace(PARALLEL_FILTER, parallel_filter)]
```
aj-fuentes, line 301: If we go forward with this I would suggest you keep the previous behavior and replace `{{` -> `{` and `}}` -> `}`. That should avoid the need to adapt already existing queries.
Maybe:
```py
re.sub(r"(?<!\{)\{parallel_filter\}(?!\})", parallel_filter, query).replace("{{", "{").replace("}}", "}")
```

Pirols: > `re.sub(r"(?<!\{)\{parallel_filter\}(?!\})", parallel_filter, query)`

I had the exact same idea, however its only advantage wrt the regular `replace` is that it would allow queries to have an hard-coded instance of the literal `'{parallel_filter}'`. Not that it costs us much, but do we really need that? I think that reserving one special placeholder for that util is fair.

> .replace("{{", "{").replace("}}", "}")

Out of caution we can think of doing that. However, IIMO the upside of not having to introduce this now-totally-artificial, odd and legacy-code-duplicating side-effect are worth considering:

1. It would be less surprising for people writing upgrade scripts (especially first timers).
2. In the long term it would spare us some challenging reviews/bugfixes due to complex setups (e.g. `util.explode_execute(cr, cr.mogrify(util.format_query(cr, f"")))`, each of which would introduce an extra set of curly braces) with dynamically baked-in values. Not to mention that python 3.12 is upon us and will introduce even more f-string capabilities to deal with.
3. People would be able to do `UPDATE t SET c = '{'` to assign `"{"` just fine, but would have to do `UPDATE t SET c = '{{{'` to assign `"{{"`? Not to mention that we would need to engineer something more clever (a regex) to resolve higher-order issues:

```python
>>> "{{{{".replace("{{", "{")
'{{'
```

In summary, IMO it would be wiser to invest towards preventing issues with old queries and "forget" about this, than to try to keep it together, preserving this oddity forever.

aj-fuentes: > its only advantage wrt the regular replace is that it would allow queries to have an hard-coded instance of the literal '{parallel_filter}'. Not that it costs us much, but do we really need that?

No, there is another advantage: you don't need to adapt the existing queries. Do you?

All the extra issues you mentioned are already present with the `format` version. What I aim to here is to not break existing code. You can put a warning about why we do the replace. Something like "this used to be done with str.format" :/

aj-fuentes: For the record
```
>>> "{{{{".replace("{{", "{")
'{{'
>>> "{{{{".format()
'{{'
```
Nothing is broken here for existing code.

Pirols: > No, there is another advantage: you don't need to adapt the existing queries. Do you?

Well yes, but I couldn't find any query where `"{{parallel_filter}}"` was passed to `explode_query_range`. After all, why would one have wanted to? The only reason I can think of is the one I mentioned: to hardcode the literal placeholder in the query. And, unsurprisingly, - at least to the extent of my check - that has not happened in queries in our repo, yet.

> All the extra issues you mentioned are already present with the `format` version. 

You're right (well except for point 3: right now single hardcoded curly braces fail the format, instead of moving on) but that doesn't make them any less valid: those are issues we'd get rid of.

> What I aim to here is to not break existing code. 

Yeah, I understand that. Personally I checked all files (41 to be exact) matched by this: `rg -tpy "\{\{" src/upgrade/migrations`. Of course, there are many other ways one could get a pair of curly braces into a string (e.g. `"{"*2`), not sure how many (notice that only 4 queries needed adjustments after my search) we can expect of those, though.

> For the record
> 
> ```
> >>> "{{{{".replace("{{", "{")
> '{{'
> >>> "{{{{".format()
> '{{'
> ```
> 
> Nothing is broken here for existing code.

Indeed, my bad:

```python
>>> "{} {{{{}}}}".format("hi")
'hi {{}}'
```

aj-fuentes: > Well yes, but I couldn't find any query where "{{parallel_filter}}"

This is a public repo. You will never find all usages out there.

What specifically you want to avoid? If you want to do a replace for `{parallel_filter}` I don't see any downside in ensuring you do not replace `{{parallel_filter}}`. Can you put a specific reason why do you prefer a bare replace over the regexp one? Besides the "now you have two problems" reason OFC.

Pirols: My main point is that if we try to preserve the curly braces reduction side-effect we improve this util, but only marginally. What happens to 
`util.explode_execute(cr, "UPDATE t SET c = '{"hi": {"hello": 3}}::jsonb'`, table="t")`? We remove one of the closing curly brace and and break the jsonb.

The age old way to solve this is by adding a flag `format_side_effect=True` to decide which behaviour to run in new instances and default to the old one.

aj-fuentes: Well, I really don't see how we can keep compatibility with naive replace (even if regexp). Maybe a new flag `use_replace=False` to use replace when we know the query could be problematic will be better.

src/util/pg.py
```diff
@@ -336,6 +347,8 @@ def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_log
     :param str table: name of the *main* table of the query, used to split the processing
     :param str alias: alias used for the main table in the query
     :param int bucket_size: size of the buckets of ids to split the processing
+    :param bool format: Whether to use `.format` (instead of `.replace`) to inject the parallel filter.
```
aj-fuentes, line 350: Check if
```suggestion
    :param bool format: whether to use `str.format` (instead of `str.replace`) to replace the parallel filter,
```

src/util/pg.py
```diff
@@ -336,6 +347,8 @@ def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_log
...
+                        Setting it to `False` can prevent issues with hard-coded curly braces.
```
aj-fuentes, line 351:
```suggestion
                        setting it to `False` can prevent issues with curly braces
```

aj-fuentes (commented):
nits

src/util/pg.py
```diff
@@ -374,6 +385,8 @@ def explode_execute(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET_SIZ
     :param str table: name of the *main* table of the query, used to split the processing
     :param str alias: alias used for the main table in the query
     :param int bucket_size: size of the buckets of ids to split the processing
+    :param bool format: whether to use `.format` (instead of `.replace`) to replace the parallel filter,
```
aj-fuentes: order here as well

Pirols:

PR #143, opened 2024-10-01
[IMP] *: Using class name corresponding to odoo models

Gorash:
If the class does not have the correct name, it is possible to force the
model name using `_name`.
In order to minimize changes for migration scripts, the solution of
forcing the `_name` is made instead of correcting the class names.


kmagusiak (approved):

rco-odoo:
@robodoo r+

PR #147, opened 2024-10-08
[IMP] spreadsheet: add `iter_commands`

KangOl:
The logic to iterate the commands of spreadsheets is reused in many sripts. We collect it here into one util that handles the update if anything changed.

### DIFF
--- a/src/util/spreadsheet/misc.py
+++ b/src/util/spreadsheet/misc.py
@@ -1,0 +1,41 @@
+import json
+
+
+def iter_commands(cr, like_all=(), like_any=()):
+    if not (bool(like_all) ^ bool(like_any)):
+        raise ValueError("Please specify `like_all` or `like_any`, not both")
+    cr.execute(
+        """
+        SELECT id,
+               commands
+          FROM spreadsheet_revision
+         WHERE commands LIKE {}(%s::text[])
+        """.format("ALL" if like_all else "ANY"),
+        [list(like_all or like_any)],
+    )
+    for revision_id, data in cr.fetchall():
+        data_loaded = json.loads(data)
+        if "commands" not in data_loaded:
+            continue
+        data_old = json.dumps(data_loaded, sort_keys=True)
+
+        changed = yield data_loaded["commands"]
+        if changed is None:
+            changed = data_old != json.dumps(data_loaded, sort_keys=True)
+
+        if changed:
+            cr.execute(
+                "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s", [json.dumps(data_loaded), revision_id]
+            )
+
+
+def process_commands(cr, callback, *args, **kwargs):
+    gen = iter_commands(cr, *args, **kwargs)
+    try:
+        cmd = next(gen)
+        while True:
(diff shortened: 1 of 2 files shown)


src/util/spreadsheet/misc.py
```diff
@@ -0,0 +1,18 @@
...
+    cr.execute("SELECT id, commands FROM spreadsheet_revision WHERE id IN %s", [tuple(revision_ids)])
+    for revision_id, data in cr.fetchall():
+        data_loaded = json.loads(data)
+        data_old = json.dumps(data_loaded)
+
+        if "commands" in data_loaded:
+            yield data_loaded["commands"]
+
+        data_new = json.dumps(data_loaded)
+        if data_new != data_old:
```
KangOl, line 16: Not totally true, the keys can be in another order.

aj-fuentes: Added `sort_keys=True`. Still saving the original order to keep consistency with current scripts. It may also minimize the diff in the revision when seen from the users' perspective.

src/util/spreadsheet/misc.py
```diff
@@ -0,0 +1,21 @@
+import json
+
+
+def iter_commands(cr, revision_ids):
```
KangOl, line 4: From the current scripts that will use this function, we could change the signature to take the search patterns instead of the revision ids.

However, we will need to handle the `and` and `or` operators.
Proposal:
```suggestion
def iter_commands(cr, like_all=(), like_any=()):
    if like_all and like_any:
        raise ValueError("Please specify `like_all` or `like_any`, not both")
    where = "true"
    if like_all:
        where = cr.mogrify("commands LIKE ALL(%s)", [list(like_all)])
    elif like_any:
        where = cr.mogrify("commands LIKE ANY(%s)", [list(like_any)])
        
    cr.execute("SELECT id, commands FROM spreadsheet_revision WHERE " + where)
    for revision_id, data in cr.fetchall():
        ...
```

src/util/spreadsheet/misc.py
```diff
@@ -0,0 +1,41 @@
...
+    if not (bool(like_all) ^ bool(like_any)):
+        raise ValueError("Please specify `like_all` or `like_any`, not both")
+    cr.execute(
+        """
+        SELECT id,
+               commands
+          FROM spreadsheet_revision
+         WHERE commands LIKE {}(%s::text[])
+        """.format("ALL" if like_all else "ANY"),
+        [list(like_all or like_any)],
```
KangOl, line 14: The version that I proposed allowed to iterate over all the revisions by not passing `like_all` nor `like_any`.

aj-fuentes: > The version that I proposed allowed to iterate over all the revisions by not passing like_all nor like_any.

And that can be easily added if needed... yagni?

The current implementation is to avoid accidentally iterating over all, given that no example needs that currently.

KangOl: That's a valid explanation.

KangOl:
@robodoo r+

PR #148, opened 2024-10-09
[IMP] ease group removal in module uninstall

nseinlet:
During module uninstall, we should remove some groups, which
is prohibited when a rule is linked to it (ON DELETE RESTRICT).

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -143,7 +143,12 @@ def uninstall_module(cr, module):
     )
 
     # delete data
-    model_ids, field_ids, menu_ids, server_action_ids = [], [], [], []
+    model_ids, field_ids, menu_ids = [], [], []
+
+    # some models' data needs to be remove before or after others.
+    firsts = ["ir.rule"]
+    lasts = ["ir.actions.server"]
+
     cr.execute(
         """
             SELECT model, res_id
@@ -156,9 +161,11 @@ def uninstall_module(cr, module):
                                   AND module != d.module)
                AND module = %s
                AND model != 'ir.module.module'
-          ORDER BY id DESC
+          ORDER BY array_position(%s::text[], model::text) NULLS LAST,
+                   array_position(%s::text[], model::text) NULLS FIRST,
+                   id DESC
     """,
-        [module],
+        [module, firsts, lasts],
     )
     to_group = []
     for model, res_id in cr.fetchall():
@@ -168,8 +175,6 @@ def uninstall_module(cr, module):
             field_ids.append(res_id)
         elif model == "ir.ui.menu":
             menu_ids.append(res_id)
-        elif model == "ir.actions.server":
-            server_action_ids.append(res_id)
         else:
             to_group.append((model, res_id))
 
@@ -179,8 +184,6 @@ def uninstall_module(cr, module):
                 remove_view(cr, view_id=res_id, silent=True)
         else:
             remove_records(cr, model, [it[1] for it in group])
-    if server_action_ids:
-        remove_records(cr, "ir.actions.server", server_action_ids)
 
     if menu_ids:
         remove_menus(cr, menu_ids)


src/util/modules.py
```diff
@@ -177,6 +177,11 @@ def uninstall_module(cr, module):
         if model == "ir.ui.view":
             for _, res_id in group:
                 remove_view(cr, view_id=res_id, silent=True)
+        elif model == "res.groups":
+            group_ids = [it[1] for it in group]
+            if group_ids:
+                cr.execute("DELETE FROM rule_group_rel WHERE group_id in %s", group_ids)
```
aj-fuentes, line 183: Don't we risk here transforming a group-restricted rule into a _global_ rule?

I think this may be fine since we are removing the module owner of the group, and hopefully of the rules... It may be safer if we can verify that the module is indeed owner of all rules linked to this group.

Is this a common case or something specific to a module?

KangOl: You're right. What we need to do instead is to remove the rules before the groups.
We already have the issues with the server actions, hence the special case.

Actually, we could determine the order of deletion with a dependency graph based on the `ON DELETE` constraints. But I don't mind hardcoding model names in the `ORDER BY` clause of the query.

nseinlet: Only if the rule is linked to only one group.

KangOl: That why removing the rules first will keep the `on delete restrict` in application.

aj-fuentes: > What we need to do instead is to remove the rules before the groups.

If rules and groups are all from the removed module we are good. But in the case `module1.rule_x` being linked to `module2.group1,module2.group2` and we remove `module2`, then we should "do something" for `rule_x` of `module1`. Maybe disable it?

KangOl: In this case, we should just explicitly remove the rule and let it be recreated by the ORM.

aj-fuentes (approved):

src/util/modules.py
```diff
@@ -156,9 +161,9 @@ def uninstall_module(cr, module):
                                   AND module != d.module)
                AND module = %s
                AND model != 'ir.module.module'
-          ORDER BY id DESC
+          ORDER BY array_position(%s, model) NULLS LAST, array_position(%s, model) NULLS FIRST, id DESC
```
aj-fuentes, line 164: We at least need the `model::text` but it helps to be explicit about the array type also...
```suggestion
          ORDER BY array_position(%s::text[], model::text) NULLS LAST, array_position(%s::text[], model::text) NULLS FIRST, id DESC
```

KangOl: Are you sure it is required? It never failed during my tests.

aj-fuentes: Well it fails in matt:
```
  File "/upgrade-util/src/util/modules.py", line 152, in uninstall_module
    cr.execute(
  File "/tmp/tmpmn5jisaa/odoo/master/odoo/sql_db.py", line 354, in execute
psycopg2.errors.UndefinedFunction: function array_position(text[], character varying) does not exist
LINE 12:           ORDER BY array_position(ARRAY['ir.rule'], model) N...
                            ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

aj-fuentes: It looks like it works from PG14 up :/ (PG version strikes again)
```
psql (16.4 (Ubuntu 16.4-1.pgdg20.04+1), server 14.13 (Ubuntu 14.13-1.pgdg20.04+1))
Type "help" for help.

afu_test=> select ARRAY_POSITION(array['base'], 'base'::varchar);
 array_position
----------------
              1
```
while PG13:
```
psql (16.4 (Ubuntu 16.4-1.pgdg20.04+1), server 13.16 (Ubuntu 13.16-1.pgdg20.04+1))
Type "help" for help.

afu_test=> select ARRAY_POSITION(array['base'], 'base'::varchar);
ERROR:  function array_position(text[], character varying) does not exist
LINE 1: select ARRAY_POSITION(array['base'], 'base'::varchar);
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

KangOl: If only I had look at the failed CI... 🤦

src/util/modules.py
KangOl: We just get the first case. I don't mind handling them with specific scripts until it become recurrent.

KangOl:
upgradeci retry with always only delivery_barcode barcodes_mobile account_consolidation in 16.0 17.0 18.0

KangOl (approved):
@robodoo r+

PR #150, opened 2024-10-10
[ADD] util.records.remove_records_batch

Pirols:
Helper to remove records in batches meant to dispatch to the right removal method, like `remove_record` would, while also taking advantage of the batch removals capabilities offered by some of those methods (e.g. `remove_menus` and `remove_records`).

Inspired by https://github.com/odoo/upgrade/pull/6595/files#r1794993193

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -23,7 +23,7 @@ except ImportError:
     from openerp.tools.misc import file_open
 
 from .const import NEARLYWARN
-from .exceptions import MigrationError
+from .exceptions import MigrationError, SleepyDeveloperError
 from .helpers import (
     _get_theme_models,
     _ir_values_value,
@@ -372,6 +372,49 @@ def remove_record(cr, name):
         return remove_group(cr, group_id=res_id)
 
     return remove_records(cr, model, [res_id])
+
+
+def remove_records_batch(cr, module, names):
+    """
+    Remove records and their references in batches.
+
+    The records to remove are those corresponding to the
+    :term:`xml_id <external identifier>` formed by the given `module` and one of
+    the provided `names`.
+
+    :param str module: name of the module, whose records must be removed
+    :param list(str) names: identifiers of the records to be removed
+    """
+    if isinstance(names, str):
+        names = [names]
+    elif isinstance(names, tuple):
+        names = list(names)
+
+    if not isinstance(names, list):
+        raise SleepyDeveloperError("The provided value for `names` is not in a supported format: {}".format(names))
+
+    cr.execute(
+        """
+        SELECT ARRAY_AGG(res_id), model
+          FROM ir_model_data
+         WHERE module = %s
+           AND name = ANY(%s)
+         GROUP BY model
+        """,
+        [module, names],
+    )
+
+    for ids, model in cr.fetchall():
+        if model == "ir.ui.view":
+            for id in ids:
+                remove_view(cr, view_id=id)
+        elif model == "ir.ui.menu":
+            remove_menus(cr, ids)
+        elif model == "res.groups":
+            for id in ids:
+                remove_group(cr, group_id=id)
+        else:
+            remove_records(cr, model, ids)
 
 
 def remove_records(cr, model, ids):


diagnoza:
Why not extend the existing `remove_records()` method to allow passing of xmlids in addition to ids? At this point there's so many different methods for records removal that it's getting confusing which ones should be used. Imho we shouldn't be adding yet another one.

aj-fuentes:
Do we really need this? I'd be more inclined to believe so with stronger cases but the linked example removes a few records, all of the same type.

Pirols:
> Why not extend the existing `remove_records()` method to allow passing of xmlids in addition to ids? At this point there's so many different methods for records removal that it's getting confusing which ones should be used. Imho we shouldn't be adding yet another one.

I think[^1] that if we really wanted to do that we'd be better off changing the `remove_record` method, instead. However, without a versioning system this is really not doable and definitely not worth the gain introduced by this new method.

[^1]: Reasons being that `remove_records` is used both as a public API and as an internal helper (it's called by pretty much all the other removal helpers - including `remove_module`).

> Do we really need this? 

No, it's a nice to have. It's just a better implementation of `remove_record`.

> I'd be more inclined to believe so with stronger cases but the linked example removes a few records, all of the same type.

For a quick search: [this](https://github.com/odoo/upgrade/blob/61866bc3eb911eb661f540634ed4d76822ee80ad/migrations/account/saas~13.1.1.1/pre-migrate.py#L53-L86) is a bit chaotic and could have been a single call to the new util. [This](https://github.com/odoo/upgrade/blob/61866bc3eb911eb661f540634ed4d76822ee80ad/migrations/theme_common/saas~13.5.1.0/pre-migrate.py#L6-L85) is close to a hundred records being removed from the same module: ~100 calls to `remove_records` vs 1. 

Now, generally speaking for every version bump there is only ever going to be so many records removed from a given module, so I'd be surprised to find a case where this would be **really** necessary (e.g. >1000 calls to `remove_record`).

diagnoza:
> I think[1](https://github.com/odoo/upgrade-util/pull/150#user-content-fn-1-7e01ce6ed775025ac7827b4e80670d58) that if we really wanted to do that we'd be better off changing the remove_record method, instead. However, without a versioning system this is really not doable and definitely not worth the gain introduced by this new method.

It's only about adding an additional parameter for xmlids and an `if` statement to cover this case, I don't see how this wouldn't be doable tbh.

> For a quick search: [this](https://github.com/odoo/upgrade/blob/61866bc3eb911eb661f540634ed4d76822ee80ad/migrations/account/saas~13.1.1.1/pre-migrate.py#L53-L86) is a bit chaotic and could have been a single call to the new util. [This](https://github.com/odoo/upgrade/blob/61866bc3eb911eb661f540634ed4d76822ee80ad/migrations/theme_common/saas~13.5.1.0/pre-migrate.py#L6-L85) is close to a hundred records being removed from the same module: ~100 calls to remove_records vs 1.

This new util does the same calls, because really it's just a wrapper around existing methods. Or even worse in fact - because it's a wrapper around a wrapper. I wouldn't envy anyone who would have to move up the stack of all those methods to figure out what's being done.

We already have a method for multiple records removal. If anything, we should tweak that instead of creating a new one imo.

Pirols:
> It's only about adding an additional parameter for xmlids and an `if` statement to cover this case, I don't see how this wouldn't be doable tbh.

It's not just a matter of feasibility: do we want to have a method named `remove_record` that removes multiple records at once?

> This new util does the same calls, because really it's just a wrapper around existing methods. 

I'm afraid this is just not the case: the new util does 1 call of `remove_records` per model as opposed to 1 per record. It performs the same number of calls for two special models (groups and views) only.

> Or even worse in fact - because it's a wrapper around a wrapper. I wouldn't envy anyone who would have to move up the stack of all those methods to figure out what's being done.

It's just a batch-enabled implementation of `remove_record`.

> We already have a method for multiple records removal. If anything, we should tweak that instead of creating a new one imo.

That method is not documented and never recommended in our reviews because it requires manual inspection of the model of the records being removed. Generally speaking, it feels (to me, at least) as a semi-public api, mostly intended to be used as a helper for the other public APIs (all the removal methods). Also, it takes res_ids as input, instead of xmlids, requiring the user to run manual queries (or `util.ref` in a loop, which defeats the purpose), which further shows how it's not meant to be called directly in scripts. IMO, the only thing that should be tweaked about that helper is its name to `_remove_records()`.

diagnoza:
> do we want to have a method named remove_record that removes multiple records at once?

The point I was making is about extending `remove_records` to allow xmlids, not `remove_record`.

> I'm afraid this is just not the case: the new util does 1 call of remove_records per model as opposed to 1 per record. It performs the same number of calls for two special models (groups and views) only.

Yes, it depends on the model. For views and groups it's the same number. In order to fix it, we would have to create two more methods - `remove_views()` and `remove_groups()` :D

> It's just a batch-enabled implementation of remove_record.

Sorry, I may be missing something. Isn't `remove_records` supposed to do this already? And in case of groups it may actually be better, because it does remove them in batch (with an SQL `DELETE`) as opposed to repeatedly calling `remove_group()`. Isn't the only point of this new method to enable passing xmlids instead of ids?

> That method is not documented and never recommended in our reviews because it requires manual inspection of the model of the records being removed

Prior to running `remove_records` you will usually have an SQL query that's fetching `res_id`s based on xmlids. Using the new method you just "move" the part with xmlids from the SQL query to a list that's passed to the method directly. I don't see how this is an improvement in the context of this manual inspection.

> Generally speaking, it feels (to me, at least) as a semi-public api, mostly intended to be used as a helper for the other public APIs (all the removal methods)

I share this feeling, but I also think that `remove_records` could be transformed into a proper method instead of a helper such as it is right now. That is, if we really want to have a method that removes records in a batch, regardless of their models. I say this, because my understanding so far is that we all acknowledge that it's generally better to use explicit calls to `remove_view`/`remove_group`/`remove_menus`...

aj-fuentes:
I'd suggest we focus in the practicality of the change. Do we need it? Is it going to improve anything? Rather than ideal situations that, well, maybe they are nice things but in practice they don't matter much.

Note, IMO condensing 10 lines into one usually makes review _harder_. Being explicit helps in upgrade scripts.

Pirols:
> I'd suggest we focus in the practicality of the change. Do we need it? Is it going to improve anything? Rather than ideal situations that, well, maybe they are nice things but in practice they don't matter much.

Of course, we don't need it, it was just a more efficient to achieve what is already possible. Closing this, let's move on.

KangOl:
upgradeci skip

PR #151, opened 2024-10-10
[IMP] util.records.remove_menus: merge queries

Pirols:
Avoids passing the ids list back and forth.


KangOl (approved):
upgradeci skip

@robodoo r+

PR #155, opened 2024-10-24
[FIX] json: make json and orjson more similar

jjmaksoud:
1- dict with null keys
json allows the dump of a dict with a null key
but orjson raises an error `TypeError: Dict key must be str`
-> add option orjson.OPT_NON_STR_KEYS to orjson dumps

2- dumping namedtuple
orjson cannot serialize namedtuple objects
'TypeError: Type is not JSON serializable'
-> cast it to tuple

3- spaces after separators
orjson does not include a space after commas and columns in string dumps
and it is not configurable to do so, we can instead configure json to remove 
them as well so we have an expected output regardless of which lib is being
used

Fixes are for failures in ci tests

### DIFF
--- a/src/util/json.py
+++ b/src/util/json.py
@@ -6,7 +6,7 @@ except ImportError:
     import json
 
     def dumps(value, sort_keys=False):
-        return json.dumps(value, sort_keys=sort_keys)
+        return json.dumps(value, sort_keys=sort_keys, separators=(",", ":"))
 
     def loads(value):
         return json.loads(value)
--- a/src/util/json.py
+++ b/src/util/json.py
@@ -16,6 +16,10 @@ except ImportError:
 else:
 
     def dumps(value, sort_keys=False):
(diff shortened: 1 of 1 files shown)


src/util/json.py
```diff
@@ -16,7 +16,15 @@ def load(fp):
 else:
 
     def dumps(value, sort_keys=False):
-        option = orjson.OPT_SORT_KEYS if sort_keys else None
+        def is_namedtuple(obj):
+            return isinstance(obj, tuple) and hasattr(obj, "_fields")
+
+        if is_namedtuple(value):
```
KangOl, line 22: Can be simplified to 
```suggestion
        if isinstance(value, tuple):
            # downcast namedtuples
```

KangOl (changes requested):

KangOl (approved):
@robodoo rebase-ff r+

PR #160, opened 2024-10-30
[FIX] records: allow change of comodel for comp dep fields

jjmaksoud:
When replacing references by batch, the assertion that the model source == model dest is unique for company dependent fields, other indirect references get their model updated in this case.
So we can remove the assertion and change the comodel for jsonb company dependent fields too.

tbg-1590

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1479,7 +1479,15 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
             continue
         if ir.company_dependent_comodel:
             if ir.company_dependent_comodel == model_src:
-                assert model_src == model_dst
+                if model_src != model_dst:
+                    cr.execute(
+                        "UPDATE ir_model_fields SET relation = %s WHERE model = %s AND name = %s",
+                        [
+                            model_dst,
+                            model_of_table(cr, ir.table),
+                            ir.res_id,
+                        ],
+                    )
                 json_path = cr.mogrify(
                     "$.* ? ({})".format(" || ".join(["@ == %s"] * len(id_mapping))),
                     list(id_mapping),
@@ -1504,6 +1512,33 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
                     [list(map(list, id_mapping.items()))],
                 ).decode()
                 explode_execute(cr, query, table=ir.table)
+                # ensure all new ids exist
+                cr.execute(
+                    format_query(
+                        cr,
+                        """
+                        Select t.id AS id,
+                               j.key AS c_id,
+                               j.value AS ref
+                          FROM {table} t
+                          JOIN JSONB_EACH(t.{column}) j
+                            ON True
+                         WHERE NOT EXISTS (
+                                    SELECT 1 FROM {dest_table} WHERE id = j.value::int
+                               )
+                        """,
+                        table=ir.table,
+                        column=ir.res_id,
+                        dest_table=table_of_model(cr, model_dst),
+                    )
+                )
+                invalid_ref = cr.dictfetchall()
+                if invalid_ref:
+                    raise RuntimeError(
+                        "Invalid company dependent values for {}.{} referencing model {}: {}".format(
+                            model_of_table(cr, ir.table), ir.res_id, model_dst, invalid_ref
+                        )
+                    )
             continue
         res_model_upd = []
         if ir.res_model:


jjmaksoud:

src/util/records.py
```diff
@@ -1479,7 +1479,15 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
             continue
         if ir.company_dependent_comodel:
             if ir.company_dependent_comodel == model_src:
-                assert model_src == model_dst
+                if model_src != model_dst:
```
aj-fuentes, line 1482: Can't we just remove the `assert`? We can check instead that _all_ keys were replaced. In such cases we can "trust" that the model of the underlying field is being changed. I find it odd to update ir_model_fields relation column here...
@KangOl wdyt?

jjmaksoud: we can remove the assertion and trust that the relation will change later, I only added this part to be consistent with the other IR where we change the res_model, and I share the same concern

aj-fuentes: > I only added this part to be consistent with the other IR where we change the res_model

Can you put a link to it please?

jjmaksoud: https://github.com/odoo/upgrade-util/blob/c47b904633f1e3133c277b339ffeb534bffa056d/src/util/records.py#L1508-L1531

aj-fuentes: Well indeed. To be aligned we could assume that the "ir.res_model" of company dependent fields (jsonb version) is the model of the field. I think we could still add some sort of check after we change the model to ensure all values (ids) are present in the referenced table. Can you do a quick test to see in 18 what happens if the value points to an invalid key (id)? Is it ignored? Does it raises any error?

src/util/records.py
```diff
@@ -1479,7 +1479,15 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
...
+                    cr.execute(
+                        "UPDATE ir_model_fields SET relation = %s WHERE model = %s AND name = %s",
+                        (
+                            model_dst,
+                            model_of_table(cr, ir.table),
+                            ir.res_id,
+                        ),
```
aj-fuentes, line 1489: nit: IMHO it helps readability by reducing chances of confusing with a function call. Feel free to ignore it.
```suggestion
                        [
                            model_dst,
                            model_of_table(cr, ir.table),
                            ir.res_id,
                        ],
```

src/util/records.py
jjmaksoud: Did a small test, if the id does not exist then an [error](https://github.com/odoo/odoo/blob/3e4c4516f7d238e76e2e9b7b71a807f7cc7d55b3/odoo/fields.py#L1267-L1271) is raised and the view is inaccessible, i'll add the check, and I will do the same test for regular IR fields

src/util/records.py
```diff
@@ -1504,6 +1512,30 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
...
+                        """,
+                        table=ir.table,
+                        column=ir.res_id,
+                        dest_table=table_of_model(cr, model_dst),
+                    )
+                )
+                invalid_ref = list(cr.fetchall())
+                assert not invalid_ref, "Invalid company dependent values for {}.{} referencing model {}: {}".format(
+                    model_of_table(cr, ir.table), ir.res_id, model_dst, invalid_ref
+                )
```
KangOl, line 1538:
```suggestion
                if invalid_ref:
                    raise RuntimeError("Invalid company dependent values for {}.{} referencing model {}: {}".format(model_of_table(cr, ir.table), ir.res_id, model_dst, invalid_ref))
```

NOTE: the suggestion will still need to be formatted correctly.

KangOl (approved):
@robodoo r+

src/util/records.py
```diff
@@ -1504,6 +1512,33 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
                     [list(map(list, id_mapping.items()))],
                 ).decode()
                 explode_execute(cr, query, table=ir.table)
+                # ensure all new ids exist
+                cr.execute(
+                    format_query(
+                        cr,
+                        """
+                        Select t.id AS id,
```
aj-fuentes, line 1520: 🙀 
```suggestion
                        SELECT t.id AS id,
```

jjmaksoud: :scream_cat:

PR #161, opened 2024-11-05
[IMP] util.dispatch_by_dbuuid

Pirols:
See https://github.com/odoo/upgrade-platform/pull/705

Make dispatch_by_dbuuid able to dispatch callbacks to a db without affecting its
relatives[^1].

[^1]: ancestors, siblings and descendants.

### DIFF
--- a/src/util/specific.py
+++ b/src/util/specific.py
@@ -11,39 +11,58 @@ from .report import add_to_migration_reports
 _logger = logging.getLogger(__name__)
 
 
-@_cached
 def dbuuid(cr):
+    return _dbuuids(cr)[-1]
+
+
+@_cached
+def _dbuuids(cr):
     cr.execute(
         """
-        SELECT value
+        SELECT REPLACE(value, 'upg-neuter-', '')
           FROM ir_config_parameter
          WHERE key IN ('database.uuid', 'origin.database.uuid')
-      ORDER BY key DESC
-         LIMIT 1
-    """
+      ORDER BY key
+        """
     )
-    return cr.fetchone()[0]
+    return [uuid for (uuid,) in cr.fetchall()]
 
 
 def dispatch_by_dbuuid(cr, version, callbacks):
     """
-    Allow to execute a migration script for a specific database only, base on its dbuuid.
+    Allow to execute a migration script for a specific database only, based on its dbuuid.
 
-    Example:
-    -------
-    >>> def db_yellowbird(cr, version):
-            cr.execute("DELETE FROM ir_ui_view WHERE id=837")
+    .. example::
+        .. code-block:: python
+            def db_yellowbird(cr, version):
+                cr.execute("DELETE FROM ir_ui_view WHERE id=837")
 
-    >>> dispatch_by_dbuuid(cr, version, {
-            'ef81c07aa90936a89f4e7878e2ebc634a24fcd66': db_yellowbird,
-        })
+            util.dispatch_by_dbuuid(cr, version, {
+                "ef81c07aa90936a89f4e7878e2ebc634a24fcd66": db_yellowbird,
+            })
 
+    :param str version: Odoo version
+    :param dict[str, function] callbacks: mapping dbuuids to the functions to run against matching dbs
+
+    .. warning::
+        - Only the first match of (icp["database.uuid"], icp["origin.database.uuid"]) in `callbacks` is executed.
+
+    .. tip::
+        - If looking to prevent a callback from running against a descendant db, one can use a noop `callback`:
+        .. example::
+            .. code-block:: python
+                noop = lambda *args: None
+                util.dispatch_by_dbuuid(cr, version, {
+                    "dbuuid": noop,
+                    "ancestor's dbuuid": db_yellowbird,
+                })
     """
-    uuid = dbuuid(cr)
-    if uuid in callbacks:
-        func = callbacks[uuid]
-        _logger.info("calling dbuuid-specific function `%s`", func.__name__)
-        func(cr, version)
+    for uuid in _dbuuids(cr):
+        if uuid in callbacks:
+            func = callbacks[uuid]
+            _logger.info("calling dbuuid-specific function `%s`", func.__name__)
+            func(cr, version)
+            break
 
 
 def rename_custom_model(cr, model_name, new_model_name, custom_module=None, report_details=""):


KangOl:
Actually, prioritizing the `origin.database.uuid` was a mistake.
I would like to change it, but it will break a lot of custom scripts, even if most of them are now dead code as the databases have been upgraded.

I don't like the idea of introducing a new `neuter` flag only set by the upgrade platform.

Let's keep your `ancestor=True`, but change the implementation of `dispatch_by_dbuuid` to first try the real dbuuid. This should be retro-compatible

```python
def dispatch_by_dbuuid(cr, version, callbacks):
    query = """
       SELECT value
         FROM ir_config_parameter
        WHERE key IN ('database.uuid', 'origin.database.uuid')
     ORDER BY key ASC
    """
    cr.execute(query)
    for uuid, in cr.fetchall():
        if uuid in callbacks:
            func = callbacks[uuid]
            _logger.info("calling dbuuid-specific function `%s`", func.__name__)
            func(cr, version)
            break
```

Pirols:
> Actually, prioritizing the `origin.database.uuid` was a mistake.

I haven't looked into the history of its introduction but to me it looks like this was made necessary to have a consistent dispatch among test and production requests, due to [this](https://github.com/odoo/upgrade-platform/blob/master/upgrade-docker/sanitizer.py#L16).

> I would like to change it, but it will break a lot of custom scripts, even if most of them are now dead code as the databases have been upgraded.
> 
> I don't like the idea of introducing a new `neuter` flag only set by the upgrade platform.

The issue is caused by the overload of the `origin.database.uuid` icp key, which is why I opted for the introduction and usage of a dedicated (and overwritable[^1]) key.

[^1]: see the companion PR for the actual details.

KangOl:
> I haven't looked into the history of its introduction but to me it looks like this was made necessary to have a consistent dispatch among test and production requests, due to [this](https://github.com/odoo/upgrade-platform/blob/master/upgrade-docker/sanitizer.py#L16).

No, it's way older than that. It was added to handle the upgrade of copied databases made by the support. It date from the time when I mass upgraded all the SaaS databases in-place on the SaaS servers.

src/util/specific.py
```diff
@@ -13,21 +13,13 @@
...
-        SELECT value
-          FROM ir_config_parameter
-         WHERE key IN ('database.uuid', 'origin.database.uuid')
-      ORDER BY key DESC
-         LIMIT 1
-    """
-    )
-    return cr.fetchone()[0]
+    cr.execute("SELECT value FROM ir_config_parameter WHERE key = 'database.uuid'")
+    return cr.fetchone()[0].replace("upg-neuter-", "")
```
KangOl, line 17: I would instead keep the uuid unchanged for the `dbuuid` function and change the implementation of `dispatch_by_dbuuid` like I proposed in https://github.com/odoo/upgrade-util/pull/161#issuecomment-2457786241

aj-fuentes: ... and I agree :)

Pirols: Mh, I had something different in the works. Let me push it right away so that you can reject it before I finalize it. :D

Pirols: Actually, I'm basically just factoring into a separate (and cached method) the query, otherwise it's pretty much the same as you proposed.

aj-fuentes: You cannot change the return value of this function. It will break scripts in the wild using this.

src/util/specific.py
```diff
@@ -25,9 +25,22 @@ def dbuuid(cr):
     return cr.fetchone()[0]
 
 
-def dispatch_by_dbuuid(cr, version, callbacks):
+@_cached
+def dbuuids(cr):
```
aj-fuentes, line 29: Let's make this "private".
```suggestion
def _dbuuids(cr):
```

src/util/specific.py
aj-fuentes: Well, you had a now removed comment around here... ;) Linked to a now closed PR ;)

aj-fuentes: But in any case, just having a new function it for the purpose of caching is fine IMO. But I'd make it private.

Pirols: > Well, you had a now removed comment around here... ;) Linked to a now closed PR ;)

Aaah, yes. I confused 2 intents and opened that unnecessary PR, still I was not changing the output of that util, just changing the util used in some scripts (stupid idea, nonetheless).

src/util/specific.py
Pirols: Somewhat relevant discussion: should we deprecate the former `dbuuid()` util? Reason being that it will prioritise the origin dbuuid, when available.

aj-fuentes: I think we don't need to worry much about that for the issue at hand. Which is more related to the dispatcher.

src/util/specific.py
```diff
@@ -39,11 +52,13 @@ def dispatch_by_dbuuid(cr, version, callbacks):
...
-        func = callbacks[uuid]
-        _logger.info("calling dbuuid-specific function `%s`", func.__name__)
-        func(cr, version)
+    for uuid in _dbuuids(cr):
+        if uuid in callbacks:
+            func = callbacks[uuid]
+            _logger.info("calling dbuuid-specific function `%s`", func.__name__)
+            func(cr, version)
+        if strict:
+            break
```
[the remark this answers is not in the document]
aj-fuentes: OK, from DM: we need `strict` for some example cases where a callback should be applied to only the parent, but never to the child DBs. It makes sense.

src/util/specific.py
KangOl: It's not retro-compatible, then.

src/util/specific.py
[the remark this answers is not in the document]
KangOl: I don't understand the need.
Do we really have a lot of cases where a script should not be applied to *all* its children databases? I would instead explicitly add a no-op function for those children db that should do nothing.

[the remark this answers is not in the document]
Pirols: > Do we really have a lot of cases where a script should not be applied to _all_ its children databases? 

A lot of cases no. I was sent one: https://github.com/odoo/upgrade-specific/blob/d0459f4e1cc140c6524a69272814123fef71f65b/scripts/M8888-yyyy/base/17.0.1.0/pre-~migrate.py#L232-L256

> I would instead explicitly add a no-op function for those children db that should do nothing.

That's an interesting alternative approach, perhaps slighly hacky but rather flexible. I do like the idea. It could get ugly for cases where many related dbs diverge significantly, but we can adapt when and if the need arises. @aj-fuentes do you agree?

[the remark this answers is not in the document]
aj-fuentes: We can do that indeed. It may be worth adding that to the docstring.

src/util/specific.py
```diff
@@ -15,7 +15,7 @@
 def dbuuid(cr):
     cr.execute(
```
KangOl, line 16: Can then be written as
```suggestion
    return _dbuuids(cr)[-1]
```

Pirols:
Tested some more: seems ready to go to me. This should be merged before https://github.com/odoo/upgrade-platform/pull/705 is deployed.

KangOl (approved):
@robodoo r+

PR #162, opened 2024-11-06
[IMP] module auto install

KangOl:


nseinlet (approved):

src/util/modules.py
```diff
@@ -51,9 +51,28 @@
...
 
 INSTALLED_MODULE_STATES = ("installed", "to install", "to upgrade")
-NO_AUTOINSTALL = str2bool(os.getenv("UPG_NO_AUTOINSTALL", "0")) if version_gte("15.0") else False
 _logger = logging.getLogger(__name__)
 
+if version_gte("15.0"):
+    AUTO_INSTALL = os.getenv("UPG_AUTOINSTALL")
+    _NO_AUTOINSTALL = os.getenv("UPG_NO_AUTOINSTALL")
+    if AUTO_INSTALL:
+        if _NO_AUTOINSTALL:
```
diagnoza, line 60: np
```suggestion
    if AUTO_INSTALL and _NO_AUTOINSTALL:
```

src/util/modules.py
```diff
@@ -525,9 +544,18 @@ def force_install_module(cr, module, if_installed=None):
...
-    # auto_install modules...
     toinstall = [m for m in states if states[m] == "to install"]
-    if toinstall:
+
+    if module in toinstall:
+        _logger.info(
+            "force install of module %r (and its dependencies) because %s%s",
+            module,
+            reason,
+            " and modules {!r} are installed".format(if_installed) if if_installed else "",
```
aj-fuentes, line 554: I found it a bit confusing:
```suggestion
            " and modules {!r} are already installed".format(if_installed) if if_installed else "",
```

src/util/modules.py
```diff
@@ -51,9 +51,28 @@
...
+    _NO_AUTOINSTALL = os.getenv("UPG_NO_AUTOINSTALL")
+    if AUTO_INSTALL:
+        if _NO_AUTOINSTALL:
+            _logger.info("Ignoring the deprecated `UPG_NO_AUTOINSTALL` env variable as `UPG_AUTOINSTALL` is also set.")
+    elif _NO_AUTOINSTALL:
+        # TODO add link to doc
+        _logger.warning(
+            "Usage of the `UPG_NO_AUTOINSTALL` environment variable is deprecated. Please use `UPG_AUTOINSTALL`."
+        )
+        AUTO_INSTALL = "link_modules" if str2bool(_NO_AUTOINSTALL) else "all"
```
aj-fuentes, line 67: Clearer, IMO:
```suggestion
        AUTO_INSTALL = "only_link_modules" if str2bool(_NO_AUTOINSTALL) else "all"
```

src/util/modules.py
```diff
@@ -51,9 +51,28 @@
...
 
+if version_gte("15.0"):
+    AUTO_INSTALL = os.getenv("UPG_AUTOINSTALL")
+    _NO_AUTOINSTALL = os.getenv("UPG_NO_AUTOINSTALL")
+    if AUTO_INSTALL:
+        if _NO_AUTOINSTALL:
+            _logger.info("Ignoring the deprecated `UPG_NO_AUTOINSTALL` env variable as `UPG_AUTOINSTALL` is also set.")
+    elif _NO_AUTOINSTALL:
+        # TODO add link to doc
+        _logger.warning(
```
aj-fuentes, line 64: I'd move this to after setting AUTO_INSTALL and add the AUTO_INSTALL value we got to the log line, plus the original UPG_NO_AUTOINSTALL value.

src/util/modules.py
```diff
@@ -51,9 +51,28 @@
...
+            "Usage of the `UPG_NO_AUTOINSTALL` environment variable is deprecated. Please use `UPG_AUTOINSTALL`."
+        )
+        AUTO_INSTALL = "link_modules" if str2bool(_NO_AUTOINSTALL) else "all"
+    else:
+        AUTO_INSTALL = "all"
+else:
+    AUTO_INSTALL = "all"
+
+if AUTO_INSTALL not in ("all", "link_modules", "none"):
+    raise ValueError("Invalid value for the `UPG_AUTOINSTALL` environment variable: {!r}".format(AUTO_INSTALL))
```
aj-fuentes, line 74: If AUTO_INSTALL is not `all` I'd suggest we output an info in the logs.

aj-fuentes (approved):
I think this is for internal usage so my comments could be ignored.

src/util/modules.py
KangOl: No, this is not the same. The `else` part should then be changed to `elif not AUTO_INSTALL`.

I don't know if this make the code clearer or not...

KangOl:
@robodoo rebase-ff r+

PR #164, opened 2024-11-12
[REM] core: remove current_thread().testing

kmagusiak:
odoo/odoo#186884
task-4270485

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -5,6 +5,7 @@ import threading
 import unittest
 import uuid
 from ast import literal_eval
+from contextlib import contextmanager
 
 from lxml import etree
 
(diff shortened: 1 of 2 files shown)


KangOl:

src/base/tests/test_util.py
```diff
@@ -24,6 +26,19 @@
 from odoo.addons.base.maintenance.migrations.util.exceptions import MigrationError
 
 
+@contextmanager
+def without_testing():
+    thread = threading.current_thread()
+    testing = getattr(thread, "testing", False)
```
KangOl, line 32: Always `False` with new implementation.
```suggestion
    testing = getattr(modules.module, "current_test", False) or getattr(thread, "testing", False)
```

KangOl (changes requested):

ryv-odoo:
Can you check  it @KangOl ? Thx

KangOl (approved):
@robodoo r+

PR #165, opened 2024-11-15
[FIX] util.json: python2 compatibility

KangOl:
Force absolute import to avoid self import in python2.


KangOl:
upgradeci retry with always only account in 9.0

KangOl:
@robodoo r+

aj-fuentes (approved):

PR #166, opened 2024-11-18
[FIX] fields: ir.model.fields.selection only introduced in v13.0

xavieralt:
The `ir.model.fields.selection` was fist introduced in v13.0 (odoo/odoo@7593b887dff4f), do no try deleting entries from that table for earlier versions.


KangOl (approved):
upgradeci skip
@robodoo r+

KangOl:
upgradeci skip

PR #167, opened 2024-11-18
[FIX] util.change_field_selection_values

KangOl:
Make it usable in old versions. The `ir_model_fields_selection` table only appears in Odoo `saas~12.5`.


KangOl:
upgradeci skip

PR #170, opened 2024-12-10
[FIX] util/misc: make SelfPrintEvalContext handle starred nodes

aj-fuentes:
The star operator is "external" to the `SelfPrint` object we try to convert to string. Thus it is impossible to override a dunder method that would return always the right answer. Overriding `__iter__` is out of question. For example, it could be triggered in at least two different ways: `[*obj]`, `[for x in obj]`. From `obj` we cannot decide which one it was, thus printing `*obj` always is wrong.

The strategy in this patch is to prepare the self printing context before evaluation such that it can deal with starred `*obj` nodes. The basic idea is to replace all `*obj` occurrences with unique names that the self printing object can later transform back to the right starred expression. To ensure correctness we use `ast` and its `unparse` method.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1426,6 +1426,8 @@ class TestMisc(UnitTestCase):
                     "a",
                     "a.b" "a.b()",
                     "a.b(c)",
+                    "[('company_id', 'in', company_ids)]",
+                    "[]",
                 ]
                 + ["a {} 4".format(op) for op in ["+", "-", "*", "/", "//", "%"]]
                 + ["4 {} b".format(op) for op in ["+", "-", "*", "/", "//", "%"]]
@@ -1434,6 +1436,25 @@ class TestMisc(UnitTestCase):
     )
     def test_SelfPrint(self, value):
         evaluated = safe_eval(value, util.SelfPrintEvalContext(), nocopy=True)
+        self.assertEqual(str(evaluated), value, "Self printed result differs")
+
+        replaced_value, ctx = util.SelfPrintEvalContext.preprocess(value)
+        evaluated = safe_eval(replaced_value, ctx, nocopy=True)
+        self.assertEqual(str(evaluated), value, "Prepared self printed result differs")
+
+    @parametrize(
+        [
+            (value,)
+            for value in [
+                "[('company_id', 'in', [*company_ids, False])]",
+                "[('company_id', 'in', [False, *company_ids])]",
+            ]
+        ]
+    )
+    @unittest.skipUnless(util.ast_unparse is not None, "`ast.unparse` available from Python3.9")
+    def test_SelfPrint_prepare(self, value):
+        replaced_value, ctx = util.SelfPrintEvalContext.preprocess(value)
+        evaluated = safe_eval(replaced_value, ctx, nocopy=True)
         self.assertEqual(str(evaluated), value)
 
 
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -196,12 +196,11 @@ def _adapt_one_domain(cr, target_model, old, new, model, domain, adapter=None, f
     if not adapter:
         adapter = lambda leaf, _, __: [leaf]
 
-    evaluation_context = SelfPrintEvalContext()
-
     # pre-check domain
     if isinstance(domain, basestring):
         try:
-            eval_dom = expression.normalize_domain(safe_eval(domain, evaluation_context, nocopy=True))
+            replaced_domain, ctx = SelfPrintEvalContext.preprocess(domain)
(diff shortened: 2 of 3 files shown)


KangOl:
First, I would decouple the ast parsing from the `SelfPrintEvalContext` class. It's a defaultdict, so it can take initial values (like `dict`).
We could then just preprocess the domain and pass the result to `SelfPrintEvalContext`

```python
def preprocess(expr):
    replacer = RewriteName()
    root = ast.parse(expr.strip(), mode="eval").body
    visited = replacer.visit(root)
    return (ast.unparse(visited), replacer.replaces)


domain, replaces = preprocess(domain)
# replaces is a Dict[str, SelfPrint]
evaluation_context = SelfPrintEvalContext(replaces)

try:
    eval_dom = expression.normalize_domain(safe_eval(domain, evaluation_context, nocopy=True))
except Exception as e:
    ...
```   


Then, we may support older python3 versions if the `astunparse` package[^1] is installed[^2].
```python
try:
    from ast import unparse
except ImportError:
    try:
        from astunparse import unparse
    except ImportError:
        unparse = None
```

[^1]: https://pypi.org/project/astunparse/
[^2]: available on ubuntu: https://packages.ubuntu.com/jammy/python3-astunparse

src/util/misc.py
```diff
@@ -437,3 +446,40 @@ def __init__(self, *args, **kwargs):
...
+
+    ```
+    >>> prepared_domain, context = util.preprocess(domain)
+    >>> safe_eval(prepared_domain, context, nocopy=True)
+    ```
+
+    :meta private: exclude from online docs
+    """
+    if not ast_unparse:
+        return expr, {}
```
KangOl, line 469:
```suggestion
        return expr, SelfPrintEvalContext()
```

src/util/misc.py
```diff
@@ -437,3 +446,40 @@ def __init__(self, *args, **kwargs):
...
+
+    ```
+    >>> prepared_domain, context = util.preprocess(domain)
+    >>> safe_eval(prepared_domain, context, nocopy=True)
+    ```
+
+    :meta private: exclude from online docs
+    """
+    if not ast_unparse:
+        return expr, SelfPrintEvalContext()
```
KangOl, line 469: This case should also be tested.

src/base/tests/test_util.py
```diff
@@ -1436,6 +1437,21 @@ def test_SelfPrint(self, value):
         evaluated = safe_eval(value, util.SelfPrintEvalContext(), nocopy=True)
         self.assertEqual(str(evaluated), value)
 
+    @parametrize(
+        [
+            (value,)
+            for value in [
+                "[('company_id', 'in', [*company_ids, False])]",
+                "[('company_id', 'in', [False, *company_ids])]",
```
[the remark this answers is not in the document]
aj-fuentes: I refactored the tests to use the preprocess approach also in the other tests. That way we ensure it works as expected even when nothing was replaced. It would also cover the case when `ast_unparse is None`

KangOl:

src/util/domains.py
```diff
@@ -196,12 +196,12 @@ def _adapt_one_domain(cr, target_model, old, new, model, domain, adapter=None, f
     if not adapter:
         adapter = lambda leaf, _, __: [leaf]
 
-    evaluation_context = SelfPrintEvalContext()
-
     # pre-check domain
     if isinstance(domain, basestring):
+        replaced_domain, ctx = SelfPrintEvalContext.preprocess(domain)
         try:
```
KangOl, line 202: It could be placed in the try/except
```suggestion
        try:
            replaced_domain, ctx = SelfPrintEvalContext.preprocess(domain)
```

KangOl: > It could be placed in the try/except
Actually, it MUST to catch SyntaxError...

aj-fuentes: I pushed a try/except internal to preprocess. Also saquashed commits.

src/util/misc.py
```diff
@@ -437,3 +446,40 @@ def __init__(self, *args, **kwargs):
...
+        Prepare `expr` to be evaluated in a self printed context.
+
+        This is necessary to avoid skipping domains that `SelfPrint` cannot print.
+        Example: [('company_id', 'in', [*company_ids, False])
+
+        Returns a pair with the new expression and an evaluation context that should
+        be used in `safe_eval`.
+
+        ```
+        >>> prepared_domain, context = util.preprocess(domain)
```
KangOl, line 462:
```suggestion
        >>> prepared_domain, context = SelfPrintEvalContext.preprocess(domain)
```

aj-fuentes: done

nseinlet:
I see nothing wrong here. Mainly because it's out of my knowledge.

PR #172, opened 2024-12-10
[FIX] records: allow nulls in comp dep jsonb

jjmaksoud:
When replacing references for company dependent fields, we ensure that the new ids are valid.
Null is also a valid value when overriding a default and should not break the check.


jjmaksoud:

KangOl (approved):
@robodoo r+

PR #173, opened 2024-12-11
[IMP] util.misc.SelfPrint

KangOl:
While reviewing #170, I wanted to give an example of a real-world failure and stumble upon this expression:
```
("date_closed", ">=", datetime.datetime.combine(context_today() + relativedelta(months = -(context_today().month - 1) % 3, day = 1), datetime.time(0, 0, 0)).to_utc().strftime("%Y-%m-%d %H:%M:%S"))
```

This shown multiple things:
 - We didn't handle all operators (in this case, unary negative, but not only)
 - Since day 1 (11 years ago!), we wrongly supported operators order.

Further investigations also revealed that we didn't correctly support the `not` operator.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1430,16 +1430,21 @@ class TestMisc(UnitTestCase):
                 "a.b",
                 "a.b()",
                 "a.b(c)",
+                "a[b]",
+                "context['company_id']",
                 "[('company_id', 'in', company_ids)]",
                 "[]",
             ]
         ]
-        + [(f"a {op} 4", f"(a {op} 4)") for op in ["+", "-", "*", "/", "//", "%"]]
-        + [(f"4 {op} b", f"(4 {op} b)") for op in ["+", "-", "*", "/", "//", "%"]]
+        + [(f"a {op} 4", f"(a {op} 4)") for op in ["+", "-", "*", "/", "//", "%", "**"]]
+        + [(f"4 {op} b", f"(4 {op} b)") for op in ["+", "-", "*", "/", "//", "%", "**"]]
         + [
             ("a+b*c", "(a + (b * c))"),
             ("a+b/c-d", "((a + (b / c)) - d)"),
             ("(a+b) * c", "((a + b) * c)"),
+            ("+a", "+(a)"),
+            ("-a", "-(a)"),
+            ("-(a+b)", "-((a + b))"),
         ]
     )
     def test_SelfPrint(self, value, expected):
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1457,18 +1457,38 @@ class TestMisc(UnitTestCase):
 
     @parametrize(
         [
-            (value,)
+            (value, value)
             for value in [
+                # splat
                 "[('company_id', 'in', [*company_ids, False])]",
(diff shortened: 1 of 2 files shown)


KangOl:

KangOl:
@robodoo merge

src/base/tests/test_util.py
```diff
@@ -1424,13 +1424,24 @@ def test_expand_braces_failure(self, value):
...
-                + ["4 {} b".format(op) for op in ["+", "-", "*", "/", "//", "%"]]
+                + ["(a {} 4)".format(op) for op in ["+", "-", "*", "/", "//", "%", "**"]]
+                + ["(4 {} b)".format(op) for op in ["+", "-", "*", "/", "//", "%", "**"]]
+                + [
+                    "((a + b) * c)",
+                    "(a + (b * c))",
+                    "((a + ((b / c) - d)) % e)",
+                    "+(a)",
+                    "-(a)",
+                    "-((a + b))",
```
aj-fuentes, line 1443: I think we should add another tests that gets the non-parenthesized versions and ensures the parenthesized version we get back matches the expected result
Example:
```
a*-b -> (a*(-b)) # and the other examples here.
```

src/util/misc.py
```diff
@@ -380,59 +380,80 @@ def __init__(self, name):
...
+    def __neg__(self):
+        return SelfPrint("-(%r)" % (self,))
 
     def __repr__(self):
         return self.__name
 
     __str__ = __repr__
 
+    def __iter__(self):
+        raise RuntimeError
```
aj-fuentes, line 452:
```suggestion
        raise RuntimeError("Cannot self-print iteration")
```
I'd rather make the error less mysterious
```
  File "/home/odoo/src/odoo/16.0/odoo/tools/safe_eval.py", line 405, in safe_eval
    raise ValueError('%s: "%s" while evaluating\n%r' % (ustr(type(e)), ustr(e), expr))
ValueError: <class 'RuntimeError'>: "" while evaluating
```

src/util/misc.py
```diff
@@ -380,59 +380,80 @@ def __init__(self, name):
...
+
+    def __bool__(self):
+        raise RuntimeError
```
aj-fuentes, line 455:
```suggestion
        raise RuntimeError("Cannot self-print boolean evaluation")
```

src/base/tests/test_util.py
```diff
@@ -1457,6 +1468,23 @@ def test_SelfPrint_prepare(self, value):
...
 
+    @parametrize(
+        [
+            (value,)
+            for value in [
+                # iterators
+                "[a.b for a in b]",
+                "4 in b",
+                # bool conversions
+                "not a",
```
aj-fuentes, line 1479: For the record, `and`/`or` was also wrong since ever :/
```suggestion
                "not a",
                "a and b",
                "a or b"
```

aj-fuentes (commented):
LGTM, minimal comments.

aj-fuentes:
Small typo in one of the commit messages: `SeflPrint`

aj-fuentes:
I pushed a commit to see if the replacer idea can be extended to bool cases. In most situations the bool expression appears in the right side of a domain leaf. Let's see how the CI goes.

KangOl:
@robodoo rebase-ff r+

PR #174, opened 2024-12-12
[IMP] util.records.uninstall_module: use `remove_group`

jepbarasgarov:
In `uninstall_module` it is not using `remove_group` for records from model `res.groups` which leads to face **psycopg2.errors.ForeignKeyViolation** while removing the records from `res.groups`, if the group has relation
with _**ondelete=restrict**_.

**example**:
In [upg-2390919](https://upgrade.odoo.com/web#id=2390919&cids=1&menu_id=107&action=150&model=upgrade.request&view_type=form) the has `group_consolidation_user` has manually created referencing records in `ir.model.access` (ondelete=restrict).
So during the [removal](https://github.com/odoo/upgrade/blob/8191d819d6b0c630e2fec4b59c50b721f85ab8ce/migrations/base/saas~17.5.1.3/pre-10-modules.py#L38) of module `account_consolidation`, it causes the issue I mentioned above.

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -48,7 +48,7 @@ from .misc import on_CI, str2bool, version_gte
 from .models import delete_model
 from .orm import env, flush
 from .pg import column_exists, table_exists, target_of
-from .records import ref, remove_menus, remove_records, remove_view, replace_record_references_batch
+from .records import ref, remove_group, remove_menus, remove_records, remove_view, replace_record_references_batch
 
 INSTALLED_MODULE_STATES = ("installed", "to install", "to upgrade")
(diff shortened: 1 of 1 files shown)


jepbarasgarov:
At the [end](https://github.com/odoo/upgrade-util/blob/c65f16e851c121584966280825d485ba51989e84/src/util/records.py#L713) of `remove_group` it is already calling `remove_records`, additonally it is [removing](https://github.com/odoo/upgrade-util/blob/c65f16e851c121584966280825d485ba51989e84/src/util/records.py#L681) the on delete restricted dependencies

KangOl (approved):
@robodoo r+

PR #176, opened 2024-12-13
[IMP] util.modules: new hack to override load data mode

HydrionBurst:
The new runtime option 'load_data_init' will be used for each package when loading package graph. It works like the previous package.init but more comprehensive and complete. Now if any directly or indirectly dependent module is configured 'load_data_init', the module will also be configured 'load_data_init'

https://github.com/odoo/odoo/pull/189000

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -895,6 +895,11 @@ def _force_upgrade_of_fresh_module(cr, module, init, version):
     # Low level implementation
     # Force module state to be in `to upgrade`.
     # Needed for migration script execution. See http://git.io/vnF7f
+    if version_gte("saas~18.2") and init:
+        env_ = env(cr)
+        env_.registry._force_upgrade_scripts.add(module)
+        return
+
     cr.execute(
         """
             UPDATE ir_module_module


src/util/modules.py
```diff
@@ -907,9 +907,13 @@ def _force_upgrade_of_fresh_module(cr, module, init, version):
         [version, module],
     )
     if init and cr.rowcount:
-        # Force module in `init` mode beside its state is forced to `to upgrade`
-        # See http://git.io/vnF7O
-        odoo.tools.config["init"][module] = "oh yeah!"
+        if version_gte("saas~18.1"):
```
KangOl, line 910:
```suggestion
        if version_gte("saas~18.2"):
```

KangOl:

src/util/modules.py
```diff
@@ -907,9 +907,13 @@ def _force_upgrade_of_fresh_module(cr, module, init, version):
         [version, module],
     )
     if init and cr.rowcount:
-        # Force module in `init` mode beside its state is forced to `to upgrade`
-        # See http://git.io/vnF7O
-        odoo.tools.config["init"][module] = "oh yeah!"
+        if version_gte("saas~18.2"):
+            env_ = env(cr)
+            env_.registry._config_init_data.add(module)
```
HydrionBurst, line 912: Hi @KangOl ,
I make this hack logically equivalent to the old hack for the https://github.com/odoo/odoo/pull/189000
1. it will force loading data/demo in 'init' mode for `to upgrade` modules
2. it will be propagated through dependencies

i. Could you help me check if 2. is a side effect or not?
ii. Do you think we can impliment the demands by using logic `_force_call_migration_script` instead of `_force_upgrade_of_fresh_module`

KangOl: i. Yes, it was a side effect. And for the record, it already bites us in the past: https://github.com/odoo/upgrade/commit/62afc499c5f4e5ecf90e173ad45e7243a3dfa85b

ii. The point of the `fresh` term was to limit to modules in `to install` state. If you have a better naming that reflect that, we can add an alias.

aj-fuentes: > ii. The point of the fresh term was to limit to modules in to install state. If you have a better naming that reflect that, we can add an alias.

From IRL with Chong. The discussion was more about implementing a proper hook in the loading mechanism to say: run the upgrade scripts of this module before it is being installed. Instead of altering internal flags. The question is more then what's the right way to do this? Does it make sense to incorporate the use case in stable itself (upgrades has grown so much that it may indeed make sense)?

HydrionBurst: I am thinking to use
`env_.registry._config_force_upgrade_new_module.add(module)` without changing the `state` of the module from `to install` to `to upgrade`

and use `package.state == 'to upgrade' or package.force_upgrade` to decide if we should execute upgrade script or not.

KangOl: That would also work.

HydrionBurst: > I am thinking to use `env_.registry._config_force_upgrade_new_module.add(module)` here without changing the `state` of the module from `to install` to `to upgrade`
> 
> and use `package.state == 'to upgrade' or package.force_upgrade` in [`MigrationManager`](https://github.com/odoo/odoo/blob/819b2718dcf9526626efc36889f23ed15eec1b6a/odoo/modules/migration.py#L165) to decide if we should execute upgrade script or not.

OK, I will implement the new strategy since it is acutally less hacky and should have less side-effects
And in the migration script, we can even know if the script is forcely executed by checking `module.state == 'to install'` during the `pre-` and `post-` (`end-` is not possible, which we need `package.load_state`)

src/util/modules.py
```diff
@@ -869,6 +869,11 @@ def force_upgrade_of_fresh_module(cr, module, init=True):
     """
     Force the execution of upgrade scripts for a module that is being installed.
 
+    From saas~18.2
+    Odoo Standard Odoo allows to add module names to `registry._config_force_upgrade_scripts`
+    to force the execution of their upgrade scripts. `init=False` is not supported.
```
HydrionBurst, line 874: @KangOl
I am not sure if we should support `force_upgrade_of_fresh_module(cr, module, init=False)` in the new strategy.
Could you help me check previously when we set `init=False`, did we really want the data and demo to be loaded in the `update` mode or we just didn't want the `package.init` to be propagated through dependencies.

rco-odoo (approved):
Looks better (explicit) than the former hack (implicit).

rco-odoo:
@KangOl

rco-odoo:
@aj-fuentes

aj-fuentes (approved):
LGTM

rco-odoo:
@aj-fuentes thanks!

rco-odoo:
@robodoo r+

PR #178, opened 2024-12-23
[FIX] core: odoo.orm.domains

kmagusiak:
Use the implementation of `Domain` to normalize domains. Note that this also distributes NOT conditions.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -27,6 +27,9 @@ from odoo.addons.base.maintenance.migrations.util.domains import (
 )
 from odoo.addons.base.maintenance.migrations.util.exceptions import MigrationError
 
+USE_ORM_DOMAIN = util.misc.version_gte("saas~18.2")
+NOTNOT = () if USE_ORM_DOMAIN else ("!", "!")
+
 
 class TestAdaptOneDomain(UnitTestCase):
     def setUp(self):
@@ -176,7 +179,7 @@ class TestAdaptOneDomain(UnitTestCase):
         # double '!'
         self.mock_adapter.reset_mock()
         domain = ["!", "!", ("partner_id.user_id", "=", 1)]
-        match_domain = ["!", "!", ("partner_id.friend_id", "=", 2)]
+        match_domain = [*NOTNOT, ("partner_id.friend_id", "=", 2)]
         new_domain = _adapt_one_domain(
             self.cr, "res.partner", "user_id", "friend_id", "res.users", domain, adapter=self.mock_adapter
         )
@@ -186,7 +189,7 @@ class TestAdaptOneDomain(UnitTestCase):
         # triple '!'
         self.mock_adapter.reset_mock()
         domain = ["!", "!", "!", ("partner_id.user_id", "=", 1)]
-        match_domain = ["!", "!", "!", ("partner_id.friend_id", "=", 2)]
+        match_domain = [*NOTNOT, "!", ("partner_id.friend_id", "=", 2)]
         new_domain = _adapt_one_domain(
             self.cr, "res.partner", "user_id", "friend_id", "res.users", domain, adapter=self.mock_adapter
         )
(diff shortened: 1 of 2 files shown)


KangOl:
upgradeci retry with always only account_inter_company_rules hr_holidays stock

src/base/tests/test_util.py
KangOl, line 1: Those changes breaks the tests in older versions.

src/base/tests/test_util.py
```diff
@@ -28,6 +28,9 @@
 from odoo.addons.base.maintenance.migrations.util.exceptions import MigrationError
 
 
+USE_ORM_DOMAIN = util.misc.version_gte("saas~18.2")
+
```
KangOl, line 32: Most of the diff is about the `"!", "!"` that are removed. We can handle these with unpacking.
```suggestion
USE_ORM_DOMAIN = util.misc.version_gte("saas~18.2")
NOTNOT = () if USE_ORM_DOMAIN else ("!", "!")
```

src/base/tests/test_util.py
```diff
@@ -176,7 +179,10 @@ def test_adapter_more_domains(self):
         # double '!'
         self.mock_adapter.reset_mock()
         domain = ["!", "!", ("partner_id.user_id", "=", 1)]
-        match_domain = ["!", "!", ("partner_id.friend_id", "=", 2)]
+        if USE_ORM_DOMAIN:
+            match_domain = [("partner_id.friend_id", "=", 2)]
+        else:
+            match_domain = ["!", "!", ("partner_id.friend_id", "=", 2)]
```
KangOl, line 185:
```suggestion
        match_domain = [*NOTNOT, ("partner_id.friend_id", "=", 2)]
```

Same for all other changes.

KangOl (changes requested):

src/base/tests/test_util.py
```diff
@@ -859,14 +883,16 @@ def test_invert_boolean_field(self):
...
+        expected = [(old_name, "!=", True)] if USE_ORM_DOMAIN else ["!", (old_name, "=", True)]
+        self.assertEqual(literal_eval(fltr.domain), expected)
 
         # invert with same name; will invert domains and data
         with mock.patch.object(cr, "commit", lambda: ...):
             util.invert_boolean_field(cr, model, old_name, old_name)
 
         util.invalidate(fltr)
-        self.assertEqual(literal_eval(fltr.domain), ["!", "!", (old_name, "=", True)])
+        expected = ["!", (old_name, "!=", True)] if USE_ORM_DOMAIN else ["!", "!", (old_name, "=", True)]
```
KangOl, line 894: Note that it show that the domain simplification could still be improved.

src/base/tests/test_util.py
```diff
@@ -389,11 +392,31 @@ class TestRemoveFieldDomains(UnitTestCase):
...
+                ["|", FALSE_LEAF, ("state", "!=", "done")]
+                if USE_ORM_DOMAIN
+                else ["!", "&", TRUE_LEAF, ("state", "=", "done")],
+            ),
+            (
+                ["!", "|", ("updated", "=", 0), ("state", "=", "done")],
+                ["&", TRUE_LEAF, ("state", "!=", "done")]
+                if USE_ORM_DOMAIN
+                else ["!", "|", FALSE_LEAF, ("state", "=", "done")],
+            ),
```
KangOl, line 419: Same for those case.

KangOl (approved):
@robodoo r+

src/base/tests/test_util.py
kmagusiak: Actually, using the Domain results in `[(old_name, '=', True)]`, but the inverse boolean just does `[..., *(~Domain(some_leaf)), ...]` and does not normalize the whole domain.

src/util/domains.py
```diff
@@ -45,7 +45,9 @@
 
 # import from domains/expression
 try:
-    import odoo.domains as _dom
+    if not version_gte("saas~18.2"):
```
aj-fuentes, line 48: Shouldn't we make this 18.1 as the [original](https://github.com/odoo/upgrade-util/pull/99) patch (and the odoo/odoo part) was merged there?

kmagusiak: I waited especially for changes in domains for **18.2**.
The main one is having semantically equivalent complements of domains. For example: `['!', ('opt_date', '>', '2020-01-01')]` is not equivalent to `[('opt_date', '<=', '2020-01-01')]` because of null value handling - so when building/translating domains this is now taken into account and the rewrite does not happen. Before 18.2, `distribute_not` has this "bug" and it is part of domain building.

aj-fuentes: Don't we need to keep then both? Or we don't care about upgrading to 18.1 with the pre-18.1 way?

kmagusiak: In 18.1, you are migrated using the old `normalize_domain` function. As of 18.2, you are migrated using the Domain object which distributes the not values. In 18.2, we fix the distribution of *not* for some edge cases, so upgrading using either method is correct.

PR #179, opened 2024-12-26
[FIX] models: save automated rule

sagu-odoo:
[here](https://github.com/odoo/upgrade-util/blob/c65f16e851c121584966280825d485ba51989e84/src/util/models.py#L172) deletion of ``base_automation`` is handled seperatly while removing model. So while merging the model
saving the automated action accordingly to prevent from deletion.

opw-4420421

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -518,6 +518,15 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
                 "len_source": len(source) + 1,
             },
         )
+    if column_exists(cr, "base_automation", "model_id"):
+        cr.execute(
+            """
+            UPDATE base_automation
+               SET model_id = %s
+             WHERE model_id = %s
+            """,
+            [model_ids[target], model_ids[source]],
+        )
 
     remove_model(cr, source, drop_table=drop_table, ignore_m2m=ignore_m2m)
 


aj-fuentes:
Can you add details about that column? When does it exist? Why?

sagu-odoo:
https://github.com/odoo/upgrade-util/blob/c65f16e851c121584966280825d485ba51989e84/src/util/models.py#L172-L173
according to this I have added and one more thing base.automation is earlier base.act.rule  it got rename in [10.saas~14](https://github.com/odoo/upgrade/blob/387012beeaae1904020a4f919fb62d9b6f74f5ce/migrations/base_automation/10.saas~14.1.0/pre-10-models.py#L31) . should I handle it also?.

aj-fuentes:
That doesn't answer the request. Where is that column coming from?

sagu-odoo:
Thank you for asking regarding the ``model_id column``. The ``model_id`` column shouldn't exist in the first place, as the [delegate=True](https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/base_automation/models/base_automation.py#L39) attribute ensures that the model_id field is related. However, in the client's database (SaaS) I was working on, the column is present despite no customizations being found for model_id in ir_model_data or ir_model_fields.

To investigate further, I replicated the scenario using a demo database, starting from version 11 and migrating up to version 15.0. In this setup, the model_id column was not present, confirming that this issue is specific to the client's database.

Based on this finding, I am closing this PR   );

aj-fuentes:
It may still be a saas-specific module. @KangOl any insight?

PR #180, opened 2024-12-27
[FIX] models: fix filters model_id correclty

sagu-odoo:
This kind of where clause is preparing due to that while merging models filters  ``model_id`` is not updating
```
NOT EXISTS(SELECT 1 FROM ir_filters o WHERE o.user_id = t.user_id AND o.action_id = t.action_id AND o.name = t.name AND o.model_id=%(new)s)
AND NOT EXISTS(SELECT 1 FROM ir_filters o WHERE true AND o.model_id=%(new)s)
```
because while checking not exist ``SELECT 1 FROM ir_filters o WHERE true AND o.model_id=%(new)s`` there are filters exist of new model(in that model is merging) due to this condition didn't fulfill and ``model_id`` is set as ``_unknown`` during remove call( because here it didn't updated before remove_model method call). This above mentioned condition should be there if Indirect reference ``res_model`` it self have the unique constraint. So, that is only on ir_model handled accordingly like ``ir_model`` have ``model`` field which have unique constraint.  

original record
```
 select id,model_id from ir_filters where id in (100,210,101,87,194,103,214,105,202,177,178,207) ;
 id  |     model_id
-----+-------------------
  87 | sale.subscription
 100 | sale.subscription
 101 | sale.subscription
 103 | sale.subscription
 105 | sale.subscription
 210 | sale.subscription
 177 | sale.subscription
 178 | sale.subscription
 194 | sale.subscription
 202 | sale.subscription
 207 | sale.subscription
 214 | sale.subscription

```

before fix
```
sagu_2403383_upg=> select id,model_id from ir_filters where id in (100,210,101,87,194,103,214,105,202,177,178,207) ;
 id  | model_id
-----+----------
  87 | _unknown
 100 | _unknown
 101 | _unknown
 178 | _unknown
 103 | _unknown
 105 | _unknown
 202 | _unknown
 177 | _unknown
 194 | _unknown
 207 | _unknown
 210 | _unknown
 214 | _unknown
(12 rows)
```

after fix
```
sagu_2403383_17.0=> select id,model_id from ir_filters where id in (100,210,101,87,194,103,214,105,202,177,178,207) ;
 id  |  model_id
-----+------------
  87 | sale.order
 100 | sale.order
 101 | sale.order
 103 | sale.order
 105 | sale.order
 177 | sale.order
 178 | sale.order
 194 | sale.order
 202 | sale.order
 207 | sale.order
 210 | sale.order
 214 | sale.order
```

opw-4381532

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -466,7 +466,7 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
             # `replace_record_references_batch`
             wheres = []
             for _, uniqs in _get_unique_indexes_with(cr, ir.table, ir.res_model):
-                sub_where = " AND ".join("o.{0} = t.{0}".format(a) for a in uniqs if a != ir.res_model) or "true"
+                sub_where = " AND ".join("o.{0} = t.{0}".format(a) for a in uniqs if a != ir.res_model) or "false"
                 wheres.append(
                     "NOT EXISTS(SELECT 1 FROM {t} o WHERE {w} AND o.{c}=%(new)s)".format(
                         t=ir.table, c=ir.res_model, w=sub_where


KangOl:
upgradeci skip

This patch is wrong. The actual issue is the shadowing of the `ignores` variable. This is fixed by #187

PR #181, opened 2025-01-02
pg: add returning_id option to parallel_execute

cawo-odoo:
The default behavior is unchanged.

This adds the possibility to parallelize modifying queries that have a `RETURNING id` clause. For those, return the resulting ids (in a defined order) instead of the affected row count.

To avoid misuse add a warning to the docstring and try to detect queries other than the ones of the intended form. Raise an error if such are found.


src/util/pg.py
```diff
@@ -172,7 +177,14 @@ def parallel_execute(cr, queries, logger=_logger):
         if getattr(threading.current_thread(), "testing", False)
         else _parallel_execute_threaded
     )
-    return parallel_execute_impl(cr, queries, logger=_logger)
+
+    if returning_id:
+        returning_id_re = re.compile(r"(?s)(?:UPDATE|DELETE|INSERT|MERGE).*RETURNING\s+\S*\.?id\s*$")
+        if not all((bool(returning_id_re.search(q)) for q in queries)):
+            raise ValueError("The returning_id parameter can only be used with certain queries.")
```
Pirols, line 184:
```suggestion
    returning_id_re = re.compile(r"(?s)(?:UPDATE|DELETE|INSERT|MERGE).*RETURNING\s+\S*\.?id\s*$")
    returning_id = all((bool(returning_id_re.search(q)) for q in queries))
```

How about we automatically detect whether the queries at hand are returning something?

src/util/pg.py
```diff
@@ -172,7 +177,14 @@ def parallel_execute(cr, queries, logger=_logger):
         if getattr(threading.current_thread(), "testing", False)
         else _parallel_execute_threaded
     )
-    return parallel_execute_impl(cr, queries, logger=_logger)
+
+    if returning_id:
+        returning_id_re = re.compile(r"(?s)(?:UPDATE|DELETE|INSERT|MERGE).*RETURNING\s+\S*\.?id\s*$")
```
Pirols, line 182:
```suggestion
        returning_id_re = re.compile(r"(?s)(?:UPDATE|DELETE|INSERT|MERGE).*RETURNING\s+(?:\w+\.)?id\s*$")
```

Change `\S*\.?` to `(?:\w+\.)?` to avoid false positive like `"UPDATE ... RETURNING name,id"` and `"DELETE FROM ... RETURNING report_id"`.

Pirols (commented):
I know it's a draft, couldn't help peeking.

Could you further elaborate on the need to limit this to returning ids only?

aj-fuentes:
For the record: I think this is not necessary. See more comments in the linked PR in upgrades.

PR #184, opened 2025-01-06
[IMP] util/models: allow to skip parent inherits

aj-fuentes:
In this situation:
```
mod1: A -> mixin -> X
                    | inherits
mod2: A -> mixin -> Y
```
if after an upgrade `mod2` no longer adds `Y` as an inherit to model
`A`, we need to remove `Y` as inherit of `A` but not `X`. Thus we need
to skip parent inherits of `Y`.

Add proper documentation to this util as well.

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Utility functions for modifying models.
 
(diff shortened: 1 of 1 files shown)


KangOl:

src/util/models.py
```diff
@@ -522,11 +522,27 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
...
+    models are also removed, unless ``strict`` mode is requested. In such case only fields
+    from the specified ``inherit`` are removed, excluding parents. It is also possible to
+    specify some fields to ``keep``, or to skip some child models (descendants from
+    ``model``) to skip from the removal via ``skip_inherit``.
+
+    :param str model: name of the model from which to remove the inherit
+    :param str inherit: name of the inherited model (or mixin) to remove
+    :param tuple(str) keep: tuple with field names to keep
+    :param boolean strit: whether to remove fields coming only from ``inherit`` and not its
+                          parent models
```
KangOl, line 540:
```suggestion
    :param boolean strict: whether to remove fields coming only from ``inherit`` and not its
                           parent models
```

src/util/models.py
```diff
@@ -522,11 +522,27 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
...
+    This function removes all fields inherited via ``inherit`` from a ``model`` and all
+    its inheriting models. All fields from the inherit, including those from its parent
+    models are also removed, unless ``strict`` mode is requested. In such case only fields
+    from the specified ``inherit`` are removed, excluding parents. It is also possible to
+    specify some fields to ``keep``, or to skip some child models (descendants from
+    ``model``) to skip from the removal via ``skip_inherit``.
+
+    :param str model: name of the model from which to remove the inherit
+    :param str inherit: name of the inherited model (or mixin) to remove
+    :param tuple(str) keep: tuple with field names to keep
```
KangOl, line 538:
```suggestion
    :param tuple(str) keep: tuple with field names to keep
    :param tuple(str) skip_inherit: list of descendent models of `model` to not process.
```

src/util/models.py
```diff
@@ -522,11 +521,33 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
     remove_model(cr, source, drop_table=drop_table, ignore_m2m=ignore_m2m)
 
 
-def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=()):
+def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=(), strict=False):
```
KangOl, line 524: I know I'm the one who propose this name, but I don't find `strict` clear enough.
What about `only`? `direct_only`?

aj-fuentes: TBH the only keyword that I find "clear enough" is `keep_parent_fields`.
`only` -> only what?
`direct_only` -> direct from where?
`keep_parent_fields` -> it hints that we keep fields, from parents, it really doesn't make sense to think about parents of `model` since what we remove is from `inherit` thus it looks more aligned with what we do.

wdyt?

KangOl: Damn, this is really [hard](https://martinfowler.com/bliki/TwoHardThings.html).

aj-fuentes: That's why I personally don't care much about the actual name as long as the docstrings are clear... the latter is still not sure but we are getting there.

KangOl:
upgradeci retry

aj-fuentes:
Error in upgradeci looks unrelated. Anyway I rebase the linked upgrade PR.

KangOl:
upgradeci retry with always only hr_expense_extract

KangOl (approved):
@robodoo r+

PR #185, opened 2025-01-08
[FIX] util/orm: verify constraint rollback

aj-fuentes:
If there constraint wasn't there attempting to add it may cause errors.
Since the rollback should only revert what's done in this util we should
record if the constraint was there in the first place.


KangOl:

KangOl (approved):
@robodoo r+

PR #186, opened 2025-01-09
[FIX] util/modules: fix discovery of no-deps modules

aj-fuentes:
When a module has no explicit dependency (e.g. pure JS module) and it is `autoinstall=True` the current dicovery mechanism completely ignores them.

In this patch we ensure that such dependecy-less modules are also handled.


KangOl:

PR #187, opened 2025-01-09
[FIX] util.merge_model: don't shadow variable

KangOl:
Don't shadow the `ignores` variable that still used in the for loop below.

Oversight of commit 2f79c47cb78d19b36bff6efa2e6bd8bf3d0e198d

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -457,8 +457,13 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
                 )
 
     if field_ids_mapping:
-        ignores = ["ir_model_fields_group_rel", "ir_model_fields_selection"]
-        replace_record_references_batch(cr, field_ids_mapping, "ir.model.fields", replace_xmlid=False, ignores=ignores)
+        replace_record_references_batch(
+            cr,
+            field_ids_mapping,
+            "ir.model.fields",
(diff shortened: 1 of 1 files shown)


src/util/models.py
```diff
@@ -457,8 +457,10 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
                 )
 
     if field_ids_mapping:
-        ignores = ["ir_model_fields_group_rel", "ir_model_fields_selection"]
-        replace_record_references_batch(cr, field_ids_mapping, "ir.model.fields", replace_xmlid=False, ignores=ignores)
+        ref_fields_ignores = ["ir_model_fields_group_rel", "ir_model_fields_selection"]
+        replace_record_references_batch(
+            cr, field_ids_mapping, "ir.model.fields", replace_xmlid=False, ignores=ref_fields_ignores
```
aj-fuentes, line 462:
```suggestion
        replace_record_references_batch(
            cr, field_ids_mapping, "ir.model.fields", replace_xmlid=False, ignores=["ir_model_fields_group_rel", "ir_model_fields_selection"]
```

aj-fuentes (approved):
We can actually use no variable at all. Anyway it LGTM

KangOl:
@robodoo r+

PR #188, opened 2025-01-13
[FIX] util/report: improve report categories for better precision

bda-odoo:
Summary:
This update improves the handling of report categories in the utility module to ensure they are more precise and well-organized.

Reason for the Change:
It was observed that the report categories were not sorted lexicographically in order, leading to a lack of precision and organization. To address this, the categories have been sorted lexicographically to enhance their clarity and structure.

opw-4424140

**here is the attached visual capture**

main_before

**After this patch**

main_after

### DIFF
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -136,7 +136,7 @@ def announce_migration_report(cr):
         "version": release.major_version,
         "major_version": major_version,
         "minor_version": minor_version,
-        "messages": migration_reports,
+        "messages": dict(sorted(migration_reports.items())),
         "get_anchor_link_to_record": get_anchor_link_to_record,
     }
     _logger.info(migration_reports)


aj-fuentes:
I'm not sure we need this. The current order reflects the order the info was logged. Do we really get tickets related to this? I don't see anything related to order in the linked ticket. The description here is a bit misleading also, you are not "refining" anything, you are just sorting the categories lexicographically. I'd suggest if you do so to just sort the data that's passed in [`annouce_migration_report.`](https://github.com/odoo/upgrade-util/blob/7198cf966bbaaa2230cd8a9f68db338eb0be0067/src/util/report.py#L141)

bda-odoo:
>  I'd suggest if you do so to just sort the data that's passed in [`annouce_migration_report.`](https://github.com/odoo/upgrade-util/blob/7198cf966bbaaa2230cd8a9f68db338eb0be0067/src/util/report.py#L141)

@aj-fuentes Thank you for your review. The suggested changes have been implemented.

bda-odoo:

bda-odoo:
Hello @aj-fuentes 

could you please review this PR?

Thanks

aj-fuentes:
I'm not sure we need this. In current setup the report is build from the order in which things get reported during the upgrade. I don't see any real improvement in just ordering. I'm not against it anyway. @KangOl?

aj-fuentes:
There is no real advantage on doing this. (Discussed in DM).

PR #189, opened 2025-01-13
[IMP] testing: make MetaCase optional

xmo-odoo:
The metaclass is being removed from community because it's unnecessary complexity when `__init_subclass__` exists.

See also odoo/odoo#113461

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -7,10 +7,14 @@ import re
 import odoo
 from odoo import api, release
 from odoo.modules.registry import Registry
-from odoo.tests.common import BaseCase, MetaCase, TransactionCase, get_db_name
+from odoo.tests.common import BaseCase, TransactionCase, get_db_name
 from odoo.tools import config
 from odoo.tools.parse_version import parse_version
 
+try:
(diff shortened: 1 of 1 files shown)


KangOl:
Can we have a *why* (and a link to the odoo PR) in the commit message?

KangOl:

aj-fuentes:
> ... and useless

It's for upgradeci to pick up the patch and test this in _all_ versions, from 8.0 if requested. Maybe there is some weird thing we need to adapt there that we didn't notice... e.g [this comment](https://github.com/odoo/upgrade-util/blob/ea530f7fb878f31786e0945ca2f7571887449b6e/src/testing.py#L73-L74) is part of what we were discussing in DM considering whether we could fully remove the case for all versions.

> still dumb ...

Making it pick the branch by name could be an option but it would complexify the code while having a explicit reference helps for code history (that we need to dig all the time for issues in old versions). If you are up to it you could propose a patch in `odoo/upgrade-ci`.

KangOl:
upgradeci retry

PR #191, opened 2025-01-15
[FIX] Don't try to discover non-installable modules

yajo:
Without this patch, if there's any module with installable=False found while running `post-01-modules-auto-discovery.py`, it will fail.

<details>

```
2025-01-15 12:09:09,972 1 INFO ? odoo.modules.migration: module base: Running migration [0.0.0>] post-01-modules-auto-discovery 
2025-01-15 12:09:12,738 1 WARNING ? odoo.modules.loading: Transient module states were reset 
2025-01-15 12:09:12,742 1 ERROR ? odoo.modules.registry: Failed to load registry 
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 87, in new
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 430, in load_modules
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 237, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/opt/odoo/custom/src/odoo/odoo/modules/migration.py", line 189, in migrate_module
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/base/0.0.0/post-01-modules-auto-discovery.py", line 8, in migrate
    _trigger_auto_discovery(cr)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 958, in _trigger_auto_discovery
    module_deps_diff(cr, module, plus=plus, minus=minus)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 687, in module_deps_diff
    new_module_dep(cr, module, new_dep)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 643, in new_module_dep
    _assert_modules_exists(cr, module, new_dep)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 638, in _assert_modules_exists
    raise AssertionError("Unexisting modules: {}".format(", ".join(unexisting_modules)))
AssertionError: Unexisting modules: base_address_city
2025-01-15 12:09:12,744 1 ERROR ? click_odoo.env_options: exception 
  File "/usr/local/lib/python3.10/site-packages/click_odoo/env_options.py", line 194, in _invoke
    with self.environment_manager(
  File "/usr/local/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.10/site-packages/click_odoo_contrib/update.py", line 299, in OdooEnvironmentWithUpdate
    _update_db(
  File "/usr/local/lib/python3.10/site-packages/click_odoo_contrib/update.py", line 270, in _update_db
    _update_db_nolock(
  File "/usr/local/lib/python3.10/site-packages/click_odoo_contrib/update.py", line 248, in _update_db_nolock
    odoo.modules.registry.Registry.new(database, update_module=True)
  File "<decorator-gen-16>", line 2, in new
  File "/opt/odoo/custom/src/odoo/odoo/tools/func.py", line 87, in locked
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 87, in new
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 430, in load_modules
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 237, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/opt/odoo/custom/src/odoo/odoo/modules/migration.py", line 189, in migrate_module
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/base/0.0.0/post-01-modules-auto-discovery.py", line 8, in migrate
    _trigger_auto_discovery(cr)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 958, in _trigger_auto_discovery
    module_deps_diff(cr, module, plus=plus, minus=minus)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 687, in module_deps_diff
    new_module_dep(cr, module, new_dep)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 643, in new_module_dep
    _assert_modules_exists(cr, module, new_dep)
  File "/usr/local/lib/python3.10/site-packages/odoo/upgrade/util/modules.py", line 638, in _assert_modules_exists
    raise AssertionError("Unexisting modules: {}".format(", ".join(unexisting_modules)))
AssertionError: Unexisting modules: base_address_city
Error: Unexisting modules: base_address_city
```

</details>

@moduon MT-3426

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -939,6 +939,8 @@ def _trigger_auto_discovery(cr):
     graph = {}
     for module in odoo.modules.get_modules():
         manifest = get_manifest(module)
+        if not manifest.get("installable", True):
+            continue
         graph[module] = (
             set(manifest["depends"]),
             manifest["auto_install"],


yajo:
Yeah. The reproducer is a manifest for a non-installable module that depends on a non-existing module. Odoo can find the 1st one (the non-installable one), but can't find the non-existing one. Then, it fails.

```python
# non_installable_module/__mainfest__.py
{
    'name': 'non_installable_module',
    'version': '1.0',
    'author': 'Your Name',
    'installable': False,
    'depends': ['non_existing_module'],
}
```

aj-fuentes:
Then this is working as expected.

yajo:
I don't think so... I can see no reason why we'd need to fail an upgrade while processing a non-installable module dependencies. Only when it is installable, it should fail.

Sometimes old version modules are kept as non-installable because you want to check if the newer Odoo version is good enough. Maybe the addon's features are upstream now and it is not required anymore, and you want to check that before starting its migration effort. That sometimes includes the addon's dependencies.

aj-fuentes:
> I don't think so... I can see no reason why we'd need to fail an upgrade while processing a non-installable module dependencies. Only when it is installable, it should fail.

> Sometimes old version modules are kept as non-installable because you want to check if the newer Odoo version is good enough. Maybe the addon's features are upstream now and it is not required anymore, and you want to check that before starting its migration effort. That sometimes includes the addon's dependencies.

We do have uninstallable modules in our codebase, because they need to be certified by a government for example. They just have _correct_ dependencies. For your use case you have other options. For example you can remove the dependencies of the uninstallable module, since you are editing the manifest anyway.

We try to ensure here that the dependencies we process are _correct_. This avoids many upgrade issues, like scripts that didn't run, or autoinstall modules that didn't install. An advantage of this: once you are ready to make the module _installable_ (for example because it was already certified) you can simply switch the flag and the dependencies should be fine. In your case you need to switch the flag and then you will still fail because the dependencies are wrong. Thus better edit the dependencies and make sure they are correct. Ultimately you can put a fake module in your addons path with the same name of the missing dependency if you don't want to edit the dependencies of the uninstallable module. (Note: I'm assuming here you are in the "check ... before starting its migration efforts" thus this kind of edits are OK)

IMO we shouldn't support the use case you present here. But, if any, the patch shouldn't be to force no check of uninstallable modules to everybody. I'd suggest to use an env variable with a list of modules to "assume" present. Once again, IMO that's not necessary given that you have many alternatives, that being said we can have a second opinion from @KangOl.

yajo:
> if any, the patch shouldn't be to force no check of uninstallable modules to everybody

OK, but if so, why would somebody enable it? Just trying to understand here. 🤔

It's true that I can work around the issue following some of your suggestions. However, none of that explains why we need to process uninstallable addons in the fist place, what's the benefit or why it's better to work around it than to fix it.

KangOl:
Note that in this case, the error is due to *missing module*.
This is because the module has been either rename, removed or merged into another one[^1].

The `base_address_city` module has been merged into `base_address_extended` 3 years ago in saas~15.2 with odoo/odoo#81710

[^1] This info is somewhat public in the Odoo commit history, but we could do something to share it in a better format (extracted from our upgrade scripts)

aj-fuentes:
> However, none of that explains why we need to process uninstallable addons in the fist place

I already said it: the dependencies are checked and ensured correct. Later you can just switch the flag. Note that modules can be installable in some version but not in others, and a full multi major versions upgrade may go over many such intermediate states.

> what's the benefit or why it's better to work around it than to fix it.

It works as designed: we check dependencies when auto-discovering. If the dependency is wrong, just remove it and re-run the upgrade. I think this is a good feedback early in development of upgrade scripts for any module... you may disagree and that's fine.

yajo:
OK, closing then.

PR #194, opened 2025-01-16
[FIX] util/models: Handle company dependent field ref

sagu-odoo:
step to reproduce:
1. createdb in 17.0, create  a one2many field in which model going to remove from inherit with res partner(for testing) because there should be company_dependent indirect references.
2. migrate to 18.0 below traceback will raised

Why this happend because [here](https://github.com/odoo/upgrade-util/blame/e38e861ea2e77de69cd4c2478ed5c203b5e38f9d/src/util/models.py#L575) indirect reference of [company_dependent](https://github.com/odoo/upgrade-util/blob/e38e861ea2e77de69cd4c2478ed5c203b5e38f9d/src/util/indirect_references.py#L126) is not handled so correcting
that handling according to this [reference](
https://github.com/odoo/upgrade-util/blame/e38e861ea2e77de69cd4c2478ed5c203b5e38f9d/src/util/models.py#L79) indirect reference column will be many2one.
```
  File "/home/odoo/src/odoo/18.0/odoo/service/server.py", line 1306, in preload_registries
  File "<decorator-gen-13>", line 2, in new
  File "/home/odoo/src/odoo/18.0/odoo/tools/func.py", line 97, in locked
  File "/home/odoo/src/odoo/18.0/odoo/modules/registry.py", line 127, in new
  File "/home/odoo/src/odoo/18.0/odoo/modules/loading.py", line 476, in load_modules
    processed_modules += load_marked_modules(env, graph,
  File "/home/odoo/src/odoo/18.0/odoo/modules/loading.py", line 364, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/home/odoo/src/odoo/18.0/odoo/modules/loading.py", line 181, in load_module_graph
  File "/home/odoo/src/odoo/18.0/odoo/modules/migration.py", line 254, in migrate_module
    mod.migrate(self.cr, installed_version)
  File "/tmp/tmpgepznpey/migrations/account/saas~17.5.1.3/pre-migrate.py", line 216, in migrate
    util.remove_inherit_from_model(
  File "/tmp/tmpgepznpey/migrations/util/models.py", line 577, in remove_inherit_from_model
    cr.execute(query, [model])  # cannot be executed in parallel. See git blame.
  File "/home/odoo/src/odoo/18.0/odoo/sql_db.py", line 354, in execute
psycopg2.errors.UndefinedColumn: column "None" does not exist
LINE 1: DELETE FROM "res_partner" WHERE ("None"='account.payment')
                                         ^
HINT:  Perhaps you meant to reference the column "res_partner.phone".

```
upg-2437434
opw-4477130
TBG-1737

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -7,6 +7,8 @@ Model operations are best done in `pre-` script of the involved modules.
 import logging
 import re
 
+from psycopg2 import sql
+
 from . import json
 from .const import ENVIRON
 from .fields import IMD_FIELD_PATTERN, remove_field
@@ -573,8 +575,9 @@ def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=(), with
             table = table_of_model(cr, relation)
             irs = [ir for ir in indirect_references(cr) if ir.table == table and ir.res_id is not None]
             for ir in irs:
-                query = 'DELETE FROM "{}" WHERE ({})'.format(ir.table, ir.model_filter())
-                cr.execute(query, [model])  # cannot be executed in parallel. See git blame.
+                if not ir.company_dependent_comodel:
+                    query = format_query(cr, "DELETE FROM {} WHERE ({})", ir.table, sql.SQL(ir.model_filter()))
+                    cr.execute(query, [model])  # cannot be executed in parallel. See git blame.
         remove_field(cr, model, field, skip_inherit="*")  # inherits will be removed by the recursive call.
 
     # down on inherits of `model`


src/util/models.py
```diff
@@ -573,7 +573,34 @@ def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=(), with
             table = table_of_model(cr, relation)
             irs = [ir for ir in indirect_references(cr) if ir.table == table and ir.res_id is not None]
             for ir in irs:
-                query = 'DELETE FROM "{}" WHERE ({})'.format(ir.table, ir.model_filter())
+                if not ir.company_dependent_comodel:
+                    query = 'DELETE FROM "{}" WHERE ({})'.format(ir.table, ir.model_filter())
```
aj-fuentes, line 577: use `format_query` here as well, time to improve this.

src/util/models.py
```diff
@@ -573,7 +573,34 @@ def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=(), with
...
+                                          key,
+                                          CASE
+                                              WHEN value::int4 = %s THEN NULL
+                                              ELSE value::int4
+                                          END)
+                                     FROM jsonb_each_text({field})
+                                   )
+                             WHERE {field} IS NOT NULL
+                               AND {field} @? %s
+                            """,
```
aj-fuentes, line 598: fromatting nit
```suggestion
                                UPDATE {table}
                                   SET {field} = (
                                       SELECT jsonb_object_agg(
                                              key,
                                              CASE
                                                  WHEN value::int4 = %s THEN NULL
                                                  ELSE value::int4
                                              END)
                                         FROM jsonb_each_text({field})
                                       )
                                 WHERE {field} IS NOT NULL
                                   AND {field} @? %s
                                """,
```

aj-fuentes:
tbg-1737

sagu-odoo:
thanks for review, suggested changes implemented. :|

KangOl:

src/util/models.py
```diff
@@ -573,7 +573,34 @@ def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=(), with
...
+                                         FROM jsonb_each_text({field})
+                                        )
+                                 WHERE {field} IS NOT NULL
+                                   AND {field} @? %s
+                                """,
+                                table=ir.table,
+                                field=ir.res_id,
+                            ),
+                            [mod_id, "$.* ? (@ == {})".format(mod_id)],
+                        )
```
KangOl, line 603: This is wrong.

You copy/paste this part from `util.remove_model`, which is specific for the removal of one2many fields pointing to `ir.model` records.

Actually, we don't need this part at all.
We have 3 possible case:
1. `ir.company_dependent_comodel == model`: the one2many explicitly point to the model, which is really weird, as the many2one we remove come from the inherit model. Anyway, there are no records to remove here.
2. `ir.company_dependent_comodel in inherit_models`: the inherits models (and records) are still there; nothing to do.
3. `else`: unrelated field; skip.

```suggestion
```

sagu-odoo:
partial error is unrelated

sagu-odoo:
Closing in favour of this https://github.com/odoo/upgrade-util/pull/195

KangOl:
Actually, we could keep it. *Explicit is better than implicit*.

KangOl (approved):
@robodoo r+

sagu-odoo:
I have tested this fix on other database found this error 
```py
>>> sql.SQL("DELETE FROM {} WHERE ({})").format(sql.Identifier("mail_follower"),sql.Identifier('"res_model"=%s')).as_string(self.env.cr._cnx)
```
result
```sql
 'DELETE FROM "mail_follower" WHERE ("""res_model"" = %s")' 
```
which causing error  `` ""res_model""='account.*' "`` column doesn't exist 

 should I do  something like this instead  instead of format query 
```sql
cr.mogrify('DELETE FROM "{}"  WHERE {}'.format(ir.table, ir.model_filter()), [model]).decode()
```

aj-fuentes:
I don't get any error if I run the code in the patch with your example. The code you ran is not really the same.
```py
>> util.format_query(cr, "DELETE FROM {} WHERE ({})", "mail_follower", sql.SQL('"res_model"=%s').format(sql.Pl
aceholder()))
'DELETE FROM "mail_follower" WHERE ("res_model"=%s)'
```

aj-fuentes:
OK, so it should be:
```
query = format_query(cr, "DELETE FROM {} WHERE ({})", ir.table, sql.SQL(ir.model_filter()))
```
see:
```
>>> util.format_query(cr, "DELETE FROM {} WHERE ({})", "mail_follower", sql.SQL(
'"res_model"=%s'))
'DELETE FROM "mail_follower" WHERE ("res_model"=%s)'
```

aj-fuentes:
The problem is `sql.Identifier('"res_model"=%s')` this happens because `format_query` assumes that any `str` is an indentifier. As long as you pass any `sql.Composable` it just reuses it as-is.

src/util/models.py
```diff
@@ -573,8 +575,11 @@ def remove_inherit_from_model(cr, model, inherit, keep=(), skip_inherit=(), with
             table = table_of_model(cr, relation)
             irs = [ir for ir in indirect_references(cr) if ir.table == table and ir.res_id is not None]
             for ir in irs:
-                query = 'DELETE FROM "{}" WHERE ({})'.format(ir.table, ir.model_filter())
-                cr.execute(query, [model])  # cannot be executed in parallel. See git blame.
+                if not ir.company_dependent_comodel:
+                    query = format_query(
+                        cr, "DELETE FROM {} WHERE ({})", ir.table, sql.SQL(ir.model_filter()).format(sql.Placeholder())
```
KangOl, line 580:
```suggestion
                        cr, "DELETE FROM {} WHERE ({})", ir.table, sql.SQL(ir.model_filter())
```

KangOl:
upgradeci retry with only hr_expense_extract website_appointment in master

KangOl (approved):
@robodoo r+

PR #195, opened 2025-01-16
[IMP] IndirectReference.model_filter

KangOl:
Handle `company_dependent_comodel` in the `model_filter` method. When it's set, there is no model, so return `false` (sql boolean).

### DIFF
--- a/src/util/indirect_references.py
+++ b/src/util/indirect_references.py
@@ -27,7 +27,8 @@ class IndirectReference(
 
 
 # By default, there is no `res_id`, no `res_model_id` and it is deleted when the linked model is removed
-IndirectReference.__new__.__defaults__ = (None, None, False)  # https://stackoverflow.com/a/18348004
+# warning: defaults are from the last fields in the namedtuple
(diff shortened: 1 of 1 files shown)


KangOl:

src/util/indirect_references.py
```diff
@@ -19,11 +19,15 @@ def model_filter(self, prefix="", placeholder="%s"):
...
             column = self.res_model
 
+        if column is None:
+            # `model` is not set when `company_dependent_comodel` is.
+            return "(false and {} IS NULL)".format(placeholder)
+
         return '{}"{}"={}'.format(prefix, column, placeholder)
 
 
 # By default, there is no `res_id`, no `res_model_id` and it is deleted when the linked model is removed
```
aj-fuentes, line 29:
```suggestion
# By default, there is no `res_id`, no `res_model_id` and it is deleted when the linked model is removed
# warning: defaults are from the last fields in named tuples
```

aj-fuentes (commented):
some nits, I will ensure I understand this to be able to say more....

KangOl:
@robodoo rebase-ff r+

PR #196, opened 2025-01-17
[FIX] util/records: add root user check

apan-odoo:
**Description:**
- In most cases, this fix: odoo/upgrade@bc4a020d334ee15263b6f9992528645a17ea000f will not work. These cases occur for databases that already have a migration history `(i.e., 16 → 17 → 18)` or when a manual upgrade of a module has been performed by the client via the UI. In such scenarios, there is a high possibility that the `write_date` has been updated by the root user, exceeding the internal time limit (1 minute). Due to which the actual purpose of `if_unchanged` is not delivered.

**Solution:**
- To handle such cases, a check of the `write_uid` is essential.

Another scenario would be a record, `base.demo`, marked as `noupdate=True` in version A. 
Upon upgrading the database to version B, the record undergoes significant changes in the XML. 
To update the record, `update_record_from_xml` is performed during the migration. 
This process updates the `write_date` of the record with the root user.

Now, if a new version C upgrade occurs and the same record is updated using 
`util.if_unchanged(cr, "base.demo", util.update_record_from_xml)`, it will not be effective in this case.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -591,8 +591,9 @@ def is_changed(cr, xmlid, interval="1 minute"):
            -- Note: use a negative search to handle the case of NULL values in write/create_date
            AND COALESCE(r.write_date < p.value::timestamp, True)
            AND r.write_date - r.create_date > interval %s
+           AND r.write_uid <> %s
         """.format(table),
-        [res_id, interval],
+        [res_id, interval, ref(cr, "base.user_root")],
     )
     return bool(cr.rowcount)
 


apan-odoo:
> The linked fix can be achieved in a different way. If you provide details in its original PR then it can be done by editing the view directly.

I agree with you. The purpose of this patch was to address the missing link here in `is_changed` . The whole idea of this patch doesn't limited to the linked fix.

aj-fuentes:
I understand the idea in this patch. But I'm not sure I agree with it. It's usually better to understand what changed and why...

KangOl:
This is not because OdooBot is the last writer that the records hasn't been modified.

apan-odoo:
> This is not because OdooBot is the last writer that the records hasn't been modified.

Exactly.. That’s what this patch is meant to address—to ensure that the record is modified. This is the primary and major reason for this patch. Otherwise, the `util.if_unchanged` is not as effective as it is intended to be. The conclusion remains 🤔

aj-fuentes:
> Exactly.. That’s what this patch is meant to address—to ensure that the record is modified. This is the primary and major reason for this patch. Otherwise, the util.if_unchanged is not as effective as it is intended to be. The conclusion remains 🤔

No, you are ignoring edits by OdooBot. The fact that it was OdooBot that edited the view is enough reason for us to stop and consider whether the update can be applied. Even though in many cases it could be an update done by a previous upgrade... we simply cannot be sure.

apan-odoo:
Didn't find any flow explaining how changes were made to a standard record by the client with `write_uid=1`. However, based on the above comments, I am closing this PR.

PR #198, opened 2025-01-23
[IMP] util.pg: better format_query integration

KangOl:


KangOl:
upgradeci retry with always only account_followup base event hr_appraisal l10n_it mail_group project project_issue quality repair survey voip warning

src/util/pg.py
```diff
@@ -362,14 +362,26 @@ def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_log
     )
 
 
+class SQLStr(sql.SQL, str):
+    def __add__(self, other):
```
aj-fuentes, line 366: It needs a meta:private docstring or proper docstrings :/ (the latter is better)
image

src/util/pg.py
```diff
@@ -362,14 +362,26 @@ def explode_execute(cr, query, table, alias=None, bucket_size=10000, logger=_log
     )
 
 
+class SQLStr(sql.SQL, str):
```
aj-fuentes, line 365: I've been thinking about this, and do we really need to make this `sql.SQL`?
I'd propose we use a similar approach as `util.PGRegexp`
```suggestion
class SQLStr(str):
    pass
```
Then we check in `util.format_query` if the parameter is of this type, in which case we use literally (same as `sql.Composable`).

KangOl: Actually, this is a good idea.

KangOl:
upgradeci retry with always only account_followup base event hr_appraisal l10n_it mail_group project project_issue quality repair survey voip warning in all versions

src/util/pg.py
```diff
@@ -55,6 +55,14 @@ class PGRegexp(str):
     """
 
 
+class SQLStr(str):
+    """
+    Wrapper for semantic meaning of: this string is a valid snippet SQL.
```
aj-fuentes, line 60: Nitpick: the meaning is for when used as parameter.
```suggestion
    Wrapper for semantic meaning of parameters: this string is a valid SQL snippet.
```

src/util/pg.py
```diff
@@ -411,14 +429,20 @@ def pg_text2html(s, wrap="p"):
...
+        """.format(
+            opening_tag="<{}>".format(wrap) if wrap else "",
+            closing_tag="</{}>".format(wrap) if wrap else "",
+            src=s,
+            esc=pg_html_escape(s, quote=False),
+        )
     )
 
 
+def get_value_or_en_translation(cr, table, column):
```
aj-fuentes, line 441: Nice to have: this should had the `pg_` prefix... but maybe it isn't worth renaming it and keeping the old name as alias.

aj-fuentes (approved):

KangOl:
@robodoo rebase-ff r+

PR #199, opened 2025-01-27
[IMP] explode_query_range: prevent two small final buckets and remove deprecated kwarg from signature

Pirols:
### [IMP] explode_query_range: remove deprecated kwarg from signature

The `prefix` keyword argument has long been deprecated and while it is still
supported for legacy reasons, it can be removed from the method's signature.

### [IMP] explode_query_range: prevent two small final buckets

In the extreme case, the last bucket could target as little as 1 record.
If the two final buckets, combined, target less than 110% the `bucket_size`:
combine them.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -299,11 +299,15 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=10000, **kwarg
     else:
         ids = list(range(min_id, max_id + 1, bucket_size))
 
-    assert min_id == ids[0] and max_id + 1 != ids[-1]  # sanity checks
-    ids.append(max_id + 1)  # ensure last bucket covers whole range
     # `ids` holds a list of values marking the interval boundaries for all buckets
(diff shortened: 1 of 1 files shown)


Pirols:
The main work behind this PR was supposed to be an improvement over the bucketing strategy. However, private discussion with @aj-fuentes made me realize that it was a bad idea. These are two commits that happened during that work, I was going to throw them away but after some consideration, I realized that especially the second one is decent enough. However, feel free to close this PR if you don't want to waste time over such a small improvement, I wouldn't disagree.

aj-fuentes:
I think we can keep this store for future improvement. But, TBH, unless there is a provable improvement in running time, I'll spare a bit of my mental energy to review other stuff ... for the moment :)

Pirols:
Indeed, I think it's the right call. Anyways, at least I do not have to keep track of these commits locally.

KangOl:
upgradeci skip

PR #200, opened 2025-01-27
[FIX] util: import hr_payroll

asno-odoo:
hr_payroll file was added but never imported in __init__.py file.


aj-fuentes:
That was intentional.

PR #202, opened 2025-01-28
[IMP] util/fields: report computes that may need adaptation

diagnoza:
Updates compute methods when a field is renamed.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -1191,11 +1191,35 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
                )
     """.format(col_prefix=col_prefix, name=get_value_or_en_translation(cr, "ir_act_server", "name"))
     cr.execute(q, {"old_pattern": p["old_pattern"], "old": p["old"], "standard_modules": tuple(standard_modules)})
+    li = ""
     if cr.rowcount:
         li = "".join(
             "<li>{}</li>".format(get_anchor_link_to_record("ir.actions.server", aid, aname))
             for aid, aname in cr.fetchall()
         )
+
+    if column_exists(cr, "ir_model_fields", "compute"):
+        # Modifying compute methods is similarly as dangerous.
+        # Only report potential compute methods that may need an update.
+        q = """
+            SELECT f.id, f.name
+              FROM ir_model_fields f
+         LEFT JOIN ir_model_data d
+                ON d.res_id = f.id
+               AND d.model = 'ir.model.fields'
+             WHERE f.compute ~ %(old_pattern)s
+               AND (  d.module IS NULL -- custom action
+                   OR d.module NOT IN %(standard_modules)s
+                   OR (d.module IN %(standard_modules)s AND d.noupdate)
+                   )
+            """
+        cr.execute(q, {"old_pattern": p["old_pattern"], "standard_modules": tuple(standard_modules)})
+        if cr.rowcount:
+            li += "".join(
+                "<li>{}</li>".format(get_anchor_link_to_record("ir.model.fields", fid, fname))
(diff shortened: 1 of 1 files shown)


aj-fuentes:
This is not safe to do.

KangOl:
As said, it's not safe.

However, we could also search on the computed field for the warning: https://github.com/odoo/upgrade-util/blob/e8516ae58aa731c9180830eaf8b791f4c63036f7/src/util/fields.py#L1170-L1214



upgradeci skip

diagnoza:
> This is not safe to do.

Yeah, I thought it may be problematic.

> However, we could also search on the computed field for the warning

I'll make it so then.

diagnoza:
@aj-fuentes @KangOl fields are now listed in the upgrade report.

src/util/fields.py
```diff
@@ -1191,20 +1191,43 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
                )
     """.format(col_prefix=col_prefix, name=get_value_or_en_translation(cr, "ir_act_server", "name"))
     cr.execute(q, {"old_pattern": p["old_pattern"], "old": p["old"], "standard_modules": tuple(standard_modules)})
+    li = ""
     if cr.rowcount:
-        li = "".join(
+        li = li.join(
```
KangOl, line 1196: Not really.
```suggestion
        li = "".join(
```

src/util/fields.py
```diff
@@ -1191,20 +1191,43 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
+           AND d.model = 'ir.model.fields'
+         WHERE compute ~ %(old_pattern)s
+           AND (  d.module IS NULL -- custom action
+               OR d.module NOT IN %(standard_modules)s
+               OR (d.module IN %(standard_modules)s AND d.noupdate)
+               )
+        """
+    cr.execute(q, {"old_pattern": p["old_pattern"], "standard_modules": tuple(standard_modules)})
+    if cr.rowcount:
+        li = li.join(
```
KangOl, line 1217:
```suggestion
        li += "".join(
```

src/util/fields.py
```diff
@@ -1191,20 +1191,43 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
     if cr.rowcount:
-        li = "".join(
+        li = li.join(
             "<li>{}</li>".format(get_anchor_link_to_record("ir.actions.server", aid, aname))
             for aid, aname in cr.fetchall()
         )
+
+    # Modifying compute methods is similarly as dangerous.
+    # Only report potential compute methods that may need an update.
+    q = """
```
KangOl, line 1203: The `compute` fields has been added in 9.0
Be compatible with upgrade of older databases.
```suggestion
    if column_exists(cr, "ir_model_fields", "compute"):  # for db < 9.0
        # Modifying compute methods is similarly as dangerous.
        # Only report potential compute methods that may need an update.
        q = """
```

src/util/fields.py
```diff
@@ -1191,20 +1191,43 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
+        SELECT f.id, f.name
+          FROM ir_model_fields f
+     LEFT JOIN ir_model_data d
+            ON d.res_id = f.id
+           AND d.model = 'ir.model.fields'
+         WHERE compute ~ %(old_pattern)s
```
KangOl, line 1209:
```suggestion
         WHERE f.compute ~ %(old_pattern)s
```

KangOl (changes requested):

KangOl (approved):
@robodoo r+

PR #203, opened 2025-01-30
[FIX] util/records: do not perform callback if xmlid is gone

aj-fuentes:
Before acceff53 we returned early if the xmlid passed to `if_unchanged` was missing. After that commit was merged we always execute the `callback`. This is problematic when the callback is `update_record_from_xml`. It will create records even if they are in `forcecreate=0`.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -568,10 +568,12 @@ def is_changed(cr, xmlid, interval="1 minute"):
      * Have been updated in an upgrade preceding the current one
      * Have not been updated in the current upgrade
 
+    If the `xmlid` doesn't exist in the DB this function returns ``None``.
+
     :param str xmlid: `xmlid` of the record to check
     :param str interval: SQL interval, a record is considered as changed if
                          `write_date > create_date + interval`
-    :rtype: bool
+    :rtype: bool or None
     """
     assert "." in xmlid
     module, _, name = xmlid.partition(".")
@@ -620,7 +622,10 @@ def if_unchanged(cr, xmlid, callback, interval="1 minute", **kwargs):
     :param str interval: interval after `create_date` on which a record is considered as
                         _changed_, see :func:`~odoo.upgrade.util.misc.is_changed`
     """
-    if not is_changed(cr, xmlid, interval=interval):
+    changed = is_changed(cr, xmlid, interval=interval)
+    if changed is None:
+        return
+    if changed is False:
         callback(cr, xmlid, **kwargs)
     else:
         force_noupdate(cr, xmlid, noupdate=True)


src/util/records.py
```diff
@@ -620,7 +620,10 @@ def if_unchanged(cr, xmlid, callback, interval="1 minute", **kwargs):
     :param str interval: interval after `create_date` on which a record is considered as
                         _changed_, see :func:`~odoo.upgrade.util.misc.is_changed`
     """
-    if not is_changed(cr, xmlid, interval=interval):
+    changed = is_changed(cr, xmlid, interval=interval)
+    if changed is None:
+        return  # do not create a record or xmlid if the xmlid is gone
```
Pirols, line 625:
```suggestion
        return
```

This comment is not really correct in the current implementation: even without that early return it would not run the callback (thereby potentially creating the xmlid) in the `None` case (due to the `is False` check).

aj-fuentes: The comment was just to highlight about what's (potentially) not done. Now I didn't want to rely in `force_noupdate` being aligned with this implementation (i.e. it doesn't create the xmlid, which is the case right now). Anyway I will remove the comment.

aj-fuentes:
I updated the docs of `is_changed` as well.

KangOl:
> I updated the docs of `is_changed` as well.

Then, we can also correct the `rtype`.

aj-fuentes:
Yes I was considering it... but I'm not sure how to express union(bool,None) :/

aj-fuentes:
OK, it works well
image

KangOl:
> Yes I was considering it... but I'm not sure how to express union(bool,None) :/

From [the documentation](https://www.sphinx-doc.org/en/master/usage/domains/python.html#info-field-lists):
> Multiple types in a type field will be linked automatically if separated by either the vertical bar (|) or the word “or”:

KangOl:
```
upgradeci retry with always only account account_online_synchronization analytic appointment approvals auth_signup base calendar calendar_sms crm data_merge delivery digest documents documents_hr documents_hr_recruitment documents_project documents_sign documents_spreadsheet event event_booth event_sms gamification helpdesk helpdesk_fsm hr hr_appraisal hr_appraisal_survey hr_contract_salary hr_contract_salary_holidays hr_expense hr_holidays hr_holidays_attendance hr_payroll hr_presence hr_recruitment im_livechat industry_fsm industry_fsm_report industry_fsm_sale knowledge l10n_be_hr_contract_salary l10n_cl_edi l10n_cl_edi_stock l10n_id_hr_payroll l10n_lu_hr_payroll l10n_tr lunch mail marketing_card mass_mailing mrp_plm payment payment_sepa_direct_debit planning portal pos_coupon project project_enterprise project_todo purchase repair sale sale_amazon sale_coupon sale_gift_card sale_subscription sale_timesheet stock stock_sms survey test_mail timesheet_grid website website_crm_partner_assign website_event_track website_forum website_hr_recruitment website_mail_channel website_payment website_profile website_sale website_sale_collect website_slides website_slides_forum website_slides_survey worksheet
```

KangOl (approved):
@robodoo r+

PR #204, opened 2025-01-31
[IMP] util/domains: hide deprecation warnings

aj-fuentes:
This is noise in the upgrade logs since we intentionally use the tools from `expression`.

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -3,6 +3,7 @@ import collections
 import functools
 import logging
 import re
+import warnings
 
 import lxml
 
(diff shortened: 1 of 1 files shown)


src/util/domains.py
```diff
@@ -92,6 +93,8 @@ def is_leaf(leaf):
     AND_OPERATOR = expression.AND_OPERATOR
     OR_OPERATOR = expression.OR_OPERATOR
     DOMAIN_OPERATORS = expression.DOMAIN_OPERATORS
+    if not on_CI():
```
KangOl, line 96: Why only on CI?

KangOl:

src/util/domains.py
KangOl: It should still be shown for non-upgrade builds, as `odoo.upgrade.util` is not imported.

aj-fuentes: Yes but would those builds cover the _same_ lines... :/
Anyway I'll remove it.

aj-fuentes: Done

PR #205, opened 2025-01-31
[IMP] util.delete_unused

KangOl:
Also consider child records when searching for record usage.

i.e. when try to remove a product category, also search for its sub categories as they will be cascade deleted if unused.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1200,7 +1200,9 @@ def delete_unused(cr, *xmlids, **kwargs):
     Remove unused records.
 
     This function will remove records pointed by `xmlids` only if they are not referenced
-    from any table.
+    from any table. For hierarchical records (like product categories), it verifies
+    if the children marked as cascade removal are also not referenced. In which case
+    the record and its children are all removed.
 
     .. note::
        The records that cannot be removed are set as `noupdate=True`.
@@ -1250,9 +1252,45 @@ def delete_unused(cr, *xmlids, **kwargs):
         table = table_of_model(cr, model)
         res_id_to_xmlid = dict(zip(ids, xids))
 
-        sub = " UNION ".join(
+        cascade_children = [
+            fk_col
+            for fk_tbl, fk_col, _, fk_act in get_fk(cr, table, quote_ident=False)
+            if fk_tbl == table and fk_act == "c"
+        ]
+        if cascade_children:
+            if len(cascade_children) == 1:
+                join = format_query(cr, "t.{}", cascade_children[0])
+            else:
+                join = sql.SQL("ANY(ARRAY[{}])").format(
+                    sql.SQL(", ").join(sql.Composed([sql.SQL("t."), sql.Identifier(cc)]) for cc in cascade_children)
+                )
+
+            kids_query = format_query(
+                cr,
+                """
+                WITH RECURSIVE _child AS (
+                    SELECT id AS root, id AS child
+                      FROM {0}
+                     WHERE id = ANY(%(ids)s)
+                     UNION  -- don't use UNION ALL in case we have a loop
+                    SELECT c.root, t.id as child
+                      FROM {0} t
+                      JOIN _child c
+                        ON c.child = {1}
+                )
+                SELECT root AS id, array_agg(child) AS children
+                  FROM _child
+              GROUP BY root
+                """,
+                table,
+                join,
+            )
+        else:
+            kids_query = format_query(cr, "SELECT id, ARRAY[id] AS children FROM {0} WHERE id = ANY(%(ids)s)", table)
+
+        sub = " UNION ALL ".join(
             [
-                'SELECT 1 FROM "{}" x WHERE x."{}" = t.id'.format(fk_tbl, fk_col)
+                format_query(cr, "SELECT 1 FROM {} x WHERE x.{} = ANY(s.children)", fk_tbl, fk_col)
                 for fk_tbl, fk_col, _, fk_act in get_fk(cr, table, quote_ident=False)
                 # ignore "on delete cascade" fk (they are indirect dependencies (lines or m2m))
                 if fk_act != "c"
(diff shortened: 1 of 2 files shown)


KangOl:
```
upgradeci retry with always only account account_reports appointment_account_payment auth_signup base calendar crm_iap_lead documents documents_hr event fleet hr_appraisal hr_expense hr_fleet hr_holidays hr_payroll hr_payroll_account hr_recruitment hr_timesheet industry_fsm industry_fsm_sale knowledge l10n_ae_hr_payroll l10n_ar l10n_at l10n_au_hr_payroll l10n_au_hr_payroll_account l10n_be l10n_be_hr_contract_salary l10n_be_hr_payroll l10n_ch_hr_payroll l10n_cl l10n_co l10n_de l10n_do l10n_es l10n_fr_hr_payroll l10n_hu l10n_in_hr_payroll l10n_it l10n_ke_hr_payroll l10n_lu_hr_payroll l10n_mn l10n_pe l10n_pl l10n_th l10n_tr mass_mailing payment payment_demo planning_contract point_of_sale portal_anonymous pos_pricer pos_restaurant product project project_todo repair resource sale_subscription sales_team sign stock stock_dropshipping survey training_base utm website_sale_slides website_slides in all versions
```

aj-fuentes (commented):
To be sure I understand:
We extend the `delete_unused` method to take into account children of the record being removed. In effect if a child has a cascade removal it is fine to consider the parent unused as long a the child is also unused.

I still have mixed feelings bout this because how can we be sure it is a parent/child relationship? Maybe it is a same-table relation but not really a parent/child situation. I'd feel more reassured if at least we restrict to `parent_id` columns :/

Just for the sake of ensuring we discuss it. If there is a parent/child relationship already present in the list of records to remove I think current approach is still correct.

This patch requires a small comment in the docs if we move forward.

src/util/records.py
```diff
@@ -1245,9 +1245,42 @@ def delete_unused(cr, *xmlids, **kwargs):
...
+                          FROM {0}
+                         WHERE id = ANY(%(ids)s)
+                         UNION
+                        SELECT c.root, t.id as child
+                          FROM {0} t
+                          JOIN _child c
+                            ON t.{1} = c.child
+                    )
+                    SELECT root, child
+                      FROM _child
```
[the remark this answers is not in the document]
KangOl: This won't work.
It's not common, but we may have multiple parent/child relations on a table. The `with recursive` are merged with `UNION`[^1] (line 1254).

[^1]: actually, we could use `UNION ALL` as it doesn't matter if there are duplicated entries.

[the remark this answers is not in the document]
aj-fuentes: I was missing to add a surrounding CTE to the union of recursive ctes. Now current computation is still odd if you want to consider multiple columns:
```
A
| col1
B
| col2
C
```
Is C a descendant of A? Yes if we consider all columns to mean "child". No if we do that following always the same column.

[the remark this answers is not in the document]
aj-fuentes: Whatever we do to get the list of children per root, I believe that `kids_query` should already return it grouped. That way is easier to understand. The subsequent query using it can just assume the `id, [list of children]` format without anything extra.

src/util/records.py
```diff
@@ -1250,9 +1253,46 @@ def delete_unused(cr, *xmlids, **kwargs):
...
+        ]
+        if cascade_children:
+            if len(cascade_children) == 1:
+                join = format_query(cr, "t.{}", cascade_children[0])
+            else:
+                join = (
+                    sql.SQL("ANY(ARRAY[")
+                    + sql.SQL(", ").join(sql.Composed([sql.SQL("t."), sql.Identifier(cc)]) for cc in cascade_children)
+                    + sql.SQL("])")
+                )
```
aj-fuentes, line 1269: simpler?
```suggestion
                join = sql.SQL("ANY(ARRAY[{}])").format(
                    sql.SQL(", ").join(sql.Composed([sql.SQL("t."), sql.Identifier(cc)]) for cc in cascade_children)
                )
```

src/util/records.py
```diff
@@ -1250,9 +1253,46 @@ def delete_unused(cr, *xmlids, **kwargs):
...
+                      FROM {0}
+                     WHERE id = ANY(%(ids)s)
+                     UNION  -- don't use UNION ALL in case we have a loop
+                    SELECT c.root, t.id as child
+                      FROM {0} t
+                      JOIN _child c
+                        ON c.child = {1}
+                )
+                SELECT root, child
+                  FROM _child
```
aj-fuentes, line 1285: and now we can really group here already... `root, [child_ids]`
Anyway this is also fine.

src/util/records.py
```diff
@@ -1250,9 +1253,46 @@ def delete_unused(cr, *xmlids, **kwargs):
...
+                """,
+                table,
+                join,
+            )
+        else:
+            kids_query = format_query(cr, "SELECT id AS root, id AS child FROM {0} WHERE id = ANY(%(ids)s)", table)
+
+        sub = " UNION ALL ".join(
```
aj-fuentes, line 1293: maybe we can use SQLStr
```suggestion
        sub = SQLStr(" UNION ALL ".join(
```

src/util/records.py
```diff
@@ -1261,18 +1301,33 @@ def delete_unused(cr, *xmlids, **kwargs):
...
+
+                SELECT t.id
+                  FROM {} t
+                  JOIN _school s
+                    ON s.id = t.id
+                 WHERE NOT EXISTS({})
+                """,
+                sql.SQL(kids_query),
+                table,
+                sql.SQL(sub),
```
aj-fuentes, line 1324: `kids_query` is already SQLStr, `sub` could be set explicitly here or before?
```suggestion
                kids_query,
                table,
                SQLStr(sub),
```

src/util/records.py
```diff
@@ -1202,6 +1202,9 @@ def delete_unused(cr, *xmlids, **kwargs):
     This function will remove records pointed by `xmlids` only if they are not referenced
     from any table.
 
+    For hierarchical records whose children are also removed (like product categories),
+    verify that those children are also not referenced.
```
aj-fuentes, line 1206: I'd suggest we make clearer that we check cascade removal of children, then we remove them all.
```suggestion
    from any table. For hierarchical records (like product categories), it verifies
    if the children marked as cascade removal are also not referenced. In which case
    the record and its children are all removed.
```

nseinlet:
@robodoo r+

PR #207, opened 2025-02-05
[REF] core: move building model methods to a separate file

rco-odoo:
Simply adapt registry.setup_models() to its new name for saas-18.2.

Companion of https://github.com/odoo/odoo/pull/193559.

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -719,7 +719,10 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
 
     with all_patches():
         # 4. Reload the registry with the models and fields converted to manual.
-        env.registry.setup_models(env.cr)
+        setup_models = (
+            env.registry._setup_models__ if hasattr(env.registry, "_setup_models__") else env.registry.setup_models
+        )
+        setup_models(env.cr)
 
     # 5. Do the operation.
     yield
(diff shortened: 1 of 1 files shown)


src/util/orm.py
```diff
@@ -719,7 +719,10 @@ def _add_magic_fields(self):
 
     with all_patches():
         # 4. Reload the registry with the models and fields converted to manual.
-        env.registry.setup_models(env.cr)
+        if version_gte("saas~18.2"):
+            env.registry._setup_models__(env.cr)
+        else:
+            env.registry.setup_models(env.cr)
```
KangOl, line 725: What about using `hasattr` instead of comparing version?
```suggestion
        setup_models = env.registry._setup_models__ if hasattr(env.registry, "_setup_models__") else env.registry.setup_models
        setup_models(env.cr)
```

aj-fuentes:
I agree with @KangOl that's better to use `hasattr` (we already do that when we can in other places). You could also include the PR reference in the commit message (we prefer that to ease git blaming --we have to dig out code all the time in upgrades, so it's really useful to have direct references).

That being said this is OK to me.

rco-odoo:
@aj-fuentes done

KangOl (approved):
@robodoo r+

KangOl:
upgradeci skip

PR #210, opened 2025-02-18
[IMP] util/indirect_references: return SQLStr as model filter

aj-fuentes:
This would make it less verbose in `util.format_query`.


KangOl:

KangOl (approved):
@robodoo r+

PR #211, opened 2025-02-18
[IMP] inconsistencies: raise_if_inconsistent

Pirols:
tbg-1678

In some cases we know an inconsistency is certainly going to cause a failure
later on during the upgrade. In such cases, it makes sense to fail immediately
and avoid the informative error message from being lost among the other
warnings.

https://github.com/odoo/upgrade/pull/7252

### DIFF
--- a/src/util/inconsistencies.py
+++ b/src/util/inconsistencies.py
@@ -7,6 +7,7 @@ from psycopg2.extensions import quote_ident
 from psycopg2.extras import Json
 from psycopg2.sql import SQL
 
+from .exceptions import UpgradeError
 from .helpers import _validate_model, table_of_model
 from .misc import chunks, str2bool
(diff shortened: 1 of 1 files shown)


src/util/inconsistencies.py
```diff
@@ -209,6 +210,7 @@ def verify_uoms(
     include_archived_products=FROM_ENV,
     auto_fix=FROM_ENV,
     ids=None,
+    raise_if_inconsistent=False,
```
KangOl, line 213: I would instead take a policy as argument instead of a boolean option.
```suggestion
    on_missmatch="report",
```

The other accepted value would be `"fail"`.

KangOl (changes requested):
What about also changing `verify_products`?

KangOl:
Then, we can close it.

PR #212, opened 2025-02-19
[FIX] util/records: improve `remove_records`

aj-fuentes:
If we are going to remove too many records we better do so in parallel. This util was observed used to remove 23401 ids.

upg-2476330

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """Utility functions for record-level operations."""
 
 import logging
@@ -34,7 +33,7 @@ from .helpers import (
 )
 from .indirect_references import indirect_references
 from .inherit import direct_inherit_parents, for_each_inherit
-from .misc import parse_version, version_gte
+from .misc import chunks, parse_version, version_gte
 from .orm import env, flush
 from .pg import (
     PGRegexp,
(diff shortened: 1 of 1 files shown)


src/util/records.py
```diff
@@ -409,7 +409,13 @@ def remove_records(cr, model, ids):
...
 
     table = table_of_model(cr, model)
-    cr.execute('DELETE FROM "{}" WHERE id IN %s'.format(table), [ids])
+    parallel_execute(
+        cr,
+        [
+            cr.mogrify(format_query(cr, "DELETE FROM {} WHERE id IN %s", table), [chunk_ids]).decode()
+            for chunk_ids in chunks(ids, 1000, fmt=tuple)
+        ],
+    )
```
Pirols, line 418: np to not format the query many times:

```suggestion
    query = format_query(cr, "DELETE FROM {} WHERE id IN %s", table)
    parallel_execute(
        cr,
        [
            cr.mogrify(query, [chunk_ids]).decode()
            for chunk_ids in chunks(ids, chunk_size, fmt=tuple)
        ],
    )
```

Also, I'd argue for a `chunk_size` parameter.

Pirols (commented):
:tada: For parallel deletes!

Time to update your training. :p

KangOl (approved):
@robodoo r+

KangOl:

PR #214, opened 2025-02-21
[IMP] util/misc: new `literal_replace`

aj-fuentes:
Avoid replacing as string.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1609,6 +1609,24 @@ class TestMisc(UnitTestCase):
         with self.assertRaises(ValueError):
             safe_eval(value, util.SelfPrintEvalContext(), nocopy=True)
 
+    @parametrize(
+        [
+            ("[('company_id','in',allowed_company_ids)]", "[('company_id', 'in', companies.active_ids)]"),
+            (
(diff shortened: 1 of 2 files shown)


KangOl:

src/base/tests/test_util.py
```diff
@@ -1609,6 +1609,23 @@ def test_SelfPrint_failure(self, value):
...
+                "[('company_id','in',allowed_company_ids or [False])]",
+                "[('company_id', 'in', companies.active_ids or [False])]",
+            ),
+            (
+                "[('company_id','in', user.other.allowed_company_ids)]",
+                "[('company_id', 'in', user.other.allowed_company_ids)]",
+            ),
+        ]
+    )
+    def test_literal_replace(self, orig, expected):
```
KangOl, line 1625:
```suggestion
    @unittest.skipUnless(util.ast_unparse is not None, "`ast.unparse` available from Python3.9")
    def test_literal_replace(self, orig, expected):
```

aj-fuentes: done

KangOl (approved):
@robodoo r+

PR #215, opened 2025-02-25
[IMP] util/report: do not show None or empty link

aj-fuentes:
Sometimes we get unexpected empty (or NULL) values in some records when reporting about them. To avoid `None` or just an empty link we make a default value for `name` in such cases.


KangOl (approved):
@robodoo r+

PR #216, opened 2025-02-26
[IMP] inconsistencies, report

Pirols:
tbg-1840

Otherwise we eventually run into memory errors when announcing the report.

```
  File "/home/odoo/src/odoo/16.0/odoo/service/server.py", line 1323, in preload_registries
  File "<decorator-gen-16>", line 2, in new
  File "/home/odoo/src/odoo/16.0/odoo/tools/func.py", line 87, in locked
  File "/home/odoo/src/odoo/16.0/odoo/modules/registry.py", line 87, in new
  File "/home/odoo/src/odoo/16.0/odoo/modules/loading.py", line 620, in load_modules
    model._register_hook()
  File "/tmp/tmpa7ynrpyl/migrations/mail/0.0.0/pre-report-migration.py", line 32, in _register_hook
    util.announce_migration_report(self.env.cr)
  File "/tmp/tmpa7ynrpyl/migrations/util/report.py", line 149, in announce_migration_report
    _announce_to_db(cr, message)
  File "/tmp/tmpa7ynrpyl/migrations/util/report.py", line 158, in _announce_to_db
    if message.strip():
  File "/home/odoo/.odoo-venvs/15.0/lib/python3.8/site-packages/markupsafe/__init__.py", line 178, in func
    return self.__class__(orig(self, *args, **kwargs))
  File "/home/odoo/.odoo-venvs/15.0/lib/python3.8/site-packages/markupsafe/__init__.py", line 74, in __new__
    return text_type.__new__(cls, base)
MemoryError
```

---

[IMP] report: warn when writing long message to migration report

Making it easier to prevent and/or identify problematic (i.e. overly long)
message insertions.

### DIFF
--- a/src/util/inconsistencies.py
+++ b/src/util/inconsistencies.py
@@ -80,7 +80,10 @@ def break_recursive_loops(cr, model, field, name_field="name"):
         name=SQL(get_value_or_en_translation(cr, table, name_field)),
     )
     cr.execute(update_query, [tuple(ids)])
-    bad_data = cr.fetchall()
+    n_updates = cr.rowcount
+    _logger.warning("%s records in %s got their %r column unset to break a recursive loop.", n_updates, table, field)
+    N_UPDATES_IN_REPORT = 20
+    bad_data = cr.fetchmany(N_UPDATES_IN_REPORT)
 
     query = format_query(
         cr,
@@ -102,13 +105,18 @@ def break_recursive_loops(cr, model, field, name_field="name"):
         """
             <details>
             <summary>
-                The following {model} were found to be recursive. Their "{field}" field has been reset.
+                The following {model} were found to be recursive. Their "{field}" field has been reset.{disclaimer}
             </summary>
               <ul>{li}</ul>
             </details>
         """.format(
             model=html_escape(model_label),
             field=html_escape(field_label),
+            disclaimer=" Find below a list of the first {} (out of {}) affected records.".format(
+                N_UPDATES_IN_REPORT, n_updates
+            )
+            if n_updates > N_UPDATES_IN_REPORT
+            else "",
             li="".join("<li>{}</li>".format(get_anchor_link_to_record(model, id_, name)) for id_, name in bad_data),
         ),
         format="html",
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -105,6 +105,11 @@ def add_to_migration_reports(message, category="Other", format="text"):
         else:
             raw = True
     migration_reports.setdefault(category, []).append((message, raw))
+    migration_reports_length = sum(len(msg) for reps in migration_reports.values() for msg, _ in reps) + sum(
+        map(len, migration_reports)
+    )
+    if migration_reports_length > 1000000:
+        _logger.warning("Upgrade report is growing suspiciously long: %s characters so far.", migration_reports_length)
 
 
 def announce_release_note(cr):


aj-fuentes:
It was already fixed. I think we can close this PR for now. It was a bad util usage.

src/util/inconsistencies.py
```diff
@@ -80,7 +80,7 @@ def break_recursive_loops(cr, model, field, name_field="name"):
         name=SQL(get_value_or_en_translation(cr, table, name_field)),
     )
     cr.execute(update_query, [tuple(ids)])
-    bad_data = cr.fetchall()
+    bad_data = cr.fetchmany(100)
```
aj-fuentes, line 83: I'd instead raise an error if too many loops were found. When checking `cr.rowcount` above.

src/util/report.py
```diff
@@ -94,6 +94,8 @@ def html_escape(text):
 
 def add_to_migration_reports(message, category="Other", format="text"):
     assert format in {"text", "html", "md", "rst"}
+    if len(message) > 10000:
```
aj-fuentes, line 97: I'd track instead the total size of the report. Once big enough (TBD how big) output a warning on each new report added.

aj-fuentes (commented):
The original issue was already solved. It was a bad usage of the utils.

aj-fuentes (approved):
LGTM.

The limits look acceptable to me. 

For reports, tt's about 1Mb. I'm honestly not sure what's the maximum size of the, still sane, reports we return. I guess we can fine tune the value later.

src/util/report.py
```diff
@@ -105,6 +106,10 @@ def add_to_migration_reports(message, category="Other", format="text"):
         else:
             raw = True
     migration_reports.setdefault(category, []).append((message, raw))
+    global migration_reports_length  # noqa: PLW0603
+    migration_reports_length += len(message)
```
KangOl, line 110: Do we really needs a global?
This is roughly equivalent to
```suggestion
    migration_reports_length = len(repr(migration_reports))
```

Pirols: It's just much slower. In fact, if we don't want to keep the global variable we can compute the sum every time. Wdyt?

```python
from timeit import timeit


big_messages = {i: "a"*10**i for i in range(10)}
many_messages = {i: "a"*i for i in range(1000)}

def _repr_sol(d):
    return len(repr(d))


def _total_len(d):
    return sum(len(v) for v in d.values())


print(timeit(lambda: _repr_sol(big_messages), number=10))       # 14.92
print(timeit(lambda: _total_len(big_messages), number=10))      # 1.8e-05
print(timeit(lambda: _repr_sol(many_messages), number=10))      # 0.006
print(timeit(lambda: _total_len(many_messages), number=10))     # 0.0004
```

KangOl: Yes
```suggestion
    migration_reports_length = sum(len(m) for (m, r) in migration_reports.values())
```

src/util/report.py
```diff
@@ -105,6 +106,10 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+    if migration_reports_length > 1000000:
+        _logger.warning("Migration report is growing suspiciously long: %s chars", migration_reports_length)
```
KangOl, line 112:
```suggestion
        _logger.warning("Upgrade report is growing suspiciously long: %s characters so far.", migration_reports_length)
```

src/util/inconsistencies.py
```diff
@@ -58,6 +59,8 @@ def break_recursive_loops(cr, model, field, name_field="name"):
     cr.execute(query)
     if not cr.rowcount:
         return
+    if cr.rowcount > 100:
+        raise UpgradeError("{} records would be updated to break the loop".format(cr.rowcount))
```
KangOl, line 63: I'm not sure a hard fail is wanted here.

KangOl: Also, what's important is not this, but `len(ids)` (computed below)

aj-fuentes: Yes it should be the len of ids. OTOH I think that any DB with too many cycles (maybe 100 is too low) should be checked. An alternative is to output an error to the logs and put in the report something like "showing the first 100 from a total of ..."

src/util/report.py
```diff
@@ -105,6 +105,9 @@ def add_to_migration_reports(message, category="Other", format="text"):
         else:
             raw = True
     migration_reports.setdefault(category, []).append((message, raw))
+    migration_reports_length = sum(len(msg) for reps in migration_reports.values() for msg, _ in reps)
```
aj-fuentes, line 108: We could count the category title as well. Just in case somebody passes a huge chunk of text :/
```suggestion
    migration_reports_length = sum(len(msg) for reps in migration_reports.values() for msg, _ in reps) + sum(map(len, migration_reports))
```

src/util/inconsistencies.py
Pirols: > maybe 100 is too low

The threshold here varies dramatically based on the size of the table, I guess.
For that reason, it's in fact probably a good idea to avoid hard failing the upgrade. See the updated proposal.

KangOl: > The threshold here varies dramatically based on the size of the table, I guess.

It could be a percentage of the table size

aj-fuentes: Here is my take: the upgrade report is intended for a human operator. Anything too long would be useless. We should in such cases just warn about the high number of loops found (maybe showing the first few entries 20?).

If there are too many loops there is most probably some automatic bug creating them. Thus it should be solved or discussed at the source. Unfortunately the admin reading the report may be clueless... we can still output something in the logs for whoever investigates this.

In my experience, valid human-caused loops, are less than five records.

aj-fuentes: > It could be a percentage of the table size

Still it looks odd to me. :/

aj-fuentes: To be clear: my comments above are as an alternative to blocking the whole upgrade. That is I believe we could output an error in the logs for more than 20 loops, and show in the report only the first 20 cases (with an extra warning if there are more).

Pirols: > > It could be a percentage of the table size
> 
> Still it looks odd to me. :/

I think it's a decent metric to decide whether to log the message or not. If there is a bug, it's likely to affect a portion of the records, as opposed to a subset.

It still does not feel good enough to fail the upgrade: we would have to check that the table is large enough too but i think it's better if we limit ourselves to a single metric, to be later tuned/expanded need be. 

> In my experience, valid human-caused loops, are less than five records.

I guess it depends on the size of the table. For big ones I guess there may be a problematic flow that consistently produces a loop.

> Here is my take: the upgrade report is intended for a human operator. Anything too long would be useless. We should in such cases just warn about the high number of loops found (maybe showing the first few entries 20?).

I think on this part we all agree.

Pirols: > To be clear: my comments above are as an alternative to blocking the whole upgrade. That is I believe we could output an error in the logs for more than 20 loops, and show in the report only the first 20 cases (with an extra warning if there are more).

I'd be fine with that too.

aj-fuentes: > I think it's a decent metric to decide whether to log the message or not. If there is a bug, it's likely to affect a portion of the records, as opposed to a subset.

Logging is cheap, we can always log if there are more than 20 IMO. The report is a different beast.

src/util/inconsistencies.py
```diff
@@ -80,7 +83,17 @@ def break_recursive_loops(cr, model, field, name_field="name"):
         name=SQL(get_value_or_en_translation(cr, table, name_field)),
     )
     cr.execute(update_query, [tuple(ids)])
-    bad_data = cr.fetchall()
+    n_updates = cr.rowcount
+    if n_updates / total_records > 0.1:
```
KangOl, line 87: Don't warn on small tables.
```suggestion
    if total_records > 100 and n_updates / total_records > 0.1:
```

src/util/inconsistencies.py
```diff
@@ -80,7 +83,17 @@ def break_recursive_loops(cr, model, field, name_field="name"):
...
+        _logger.warning(
+            "Several records (%s out of %s) in %s got their %r unset to break a recursive loop.",
```
KangOl, line 89: Also log the model and the field.

KangOl (changes requested):

src/util/inconsistencies.py
aj-fuentes: Can't we just log always? This should have been the default IMO. One line in the logs is cheap. This will also help us understand how widespread the loops are in some models.

src/util/inconsistencies.py
Pirols: That's already the case.

src/util/inconsistencies.py
Pirols: > Can't we just log always? This should have been the default IMO. One line in the logs is cheap. This will also help us understand how widespread the loops are in some models.

I also prefer this approach. We should log what happens in this util.

KangOl (approved):
@robodoo rebase-ff r+

PR #217, opened 2025-02-26
[IMP] records.delete_unused: Change default of keep_xmlids

filisbits:
Changing the default kwarg value so the xmlids get deleted even if the record is kept, starting from version `saas~18.1` to keep the current behaviour.


aj-fuentes (approved):
cc: @KangOl

PR #218, opened 2025-02-27
[IMP] report: open chat window post upgrade

diagnoza:
Chat window with the report will be opened by default after the upgrade, only for the admin user.

image

### DIFF
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -68,7 +68,7 @@ except ImportError:
         from openerp.addons.base.module.module import MyWriter
 
 from .exceptions import MigrationError
-from .misc import has_enterprise, split_osenv, version_gte
+from .misc import has_enterprise, split_osenv, version_between, version_gte
 from .orm import env, get_admin_channel, guess_admin_id
 
 migration_reports = {}
@@ -294,6 +294,21 @@ def announce(
         poster(body=message, partner_ids=[user.partner_id.id], **kw)
     except Exception:
         _logger.warning("Cannot announce message", exc_info=True)
+    else:
+        # Chat window with the report will be open post-upgrade for the admin user
+        if version_between("9.0", "saas~18.2") and user.partner_id and recipient:
+            channel_member_model = (
+                "discuss.channel.member"
+                if version_gte("saas~16.3")
+                else "mail.channel.member"
+                if version_gte("16.0")
+                else "mail.channel.partner"
+            )
+            domain = [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
+            try:
+                registry[channel_member_model].search(domain)[:1].with_context(ctx).fold_state = "open"
+            except Exception:
+                _logger.warning("Cannot unfold chat window", exc_info=True)
 
 
 def get_anchor_link_to_record(model, id, name, action_id=None):


src/util/report.py
```diff
@@ -286,6 +286,21 @@ def ref(xid):
     except Exception:
         _logger.warning("Cannot announce message", exc_info=True)
 
+    # Chat window with the report will be open post-upgrade for the admin
+    if version_gte("9.0"):
```
diagnoza, line 290: We can get rid of one level of `if`s in favor of a dictionary with versions, but I thought it's going to be less readable.

aj-fuentes: You can get rid of this because we do not support upgrading to 9.0 already :)

src/util/report.py
```diff
@@ -286,6 +286,21 @@ def ref(xid):
...
+        elif version_gte("16.0"):
+            channel_member_model = "mail.channel.member"
+        else:
+            channel_member_model = "mail.channel.partner"
+        try:
+            registry[channel_member_model].search(
+                [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
+            )[0].with_context(ctx).fold_state = "open"
+        except Exception:
+            _logger.warning("Cannot open chat window", exc_info=True)
```
aj-fuentes, line 302: I don't think we need to log anything, in any case if we do logging exception info is better done with `_logger.exception` ;)
```suggestion
    pass
```
Maybe use `contextlib.suppress` and be more specific: what kind of exception we get here? and why?

src/util/report.py
```diff
@@ -286,6 +286,21 @@ def ref(xid):
...
+        if version_gte("saas~16.3"):
+            channel_member_model = "discuss.channel.member"
+        elif version_gte("16.0"):
+            channel_member_model = "mail.channel.member"
+        else:
+            channel_member_model = "mail.channel.partner"
+        try:
+            registry[channel_member_model].search(
+                [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
+            )[0].with_context(ctx).fold_state = "open"
```
KangOl, line 300: The field `fold_state` has been removed in `saas~18.2`: odoo/odoo#193245

KangOl:

src/util/report.py
```diff
@@ -286,6 +286,16 @@ def ref(xid):
...
 
+    # Chat window with the report will be open post-upgrade for the admin user
+    if not version_gte("saas~18.2"):
+        channel_member_model = "discuss.channel.member" if version_gte("saas~16.3") else "mail.channel.member"
+        try:
+            registry[channel_member_model].search(
+                [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
+            )[0].with_context(ctx).fold_state = "open"
+        except Exception:
+            _logger.exception("Could not open chat window with upgrade report.")
```
KangOl, line 297: 1/ extract the domain in a variable
2/ avoid exception by not using `[0]`
```suggestion
        domain = [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
        registry[channel_member_model].search(domain)[:1].with_context(ctx).fold_state = "open"
```

src/util/report.py
```diff
@@ -286,6 +286,12 @@ def ref(xid):
     except Exception:
         _logger.warning("Cannot announce message", exc_info=True)
 
+    # Chat window with the report will be open post-upgrade for the admin user
+    if not version_gte("saas~18.2"):
+        channel_member_model = "discuss.channel.member" if version_gte("saas~16.3") else "mail.channel.member"
```
KangOl, line 291:
```suggestion
        channel_member_model = "discuss.channel.member" if version_gte("saas~16.3") else "mail.channel.member" if version_gte("16.0") else "mail.channel.partner"
```

src/util/report.py
```diff
@@ -286,6 +286,12 @@ def ref(xid):
     except Exception:
         _logger.warning("Cannot announce message", exc_info=True)
 
+    # Chat window with the report will be open post-upgrade for the admin user
+    if not version_gte("saas~18.2"):
```
KangOl, line 290: `fold_state` appears in v9
```suggestion
    if version_between("9.0", "saas~18.2"):
```

KangOl (changes requested):

diagnoza:
@KangOl @aj-fuentes is it ok to merge?

src/util/report.py
```diff
@@ -286,6 +286,18 @@ def ref(xid):
...
+    if version_between("9.0", "saas~18.2"):
+        channel_member_model = (
+            "discuss.channel.member"
+            if version_gte("saas~16.3")
+            else "mail.channel.member"
+            if version_gte("16.0")
+            else "mail.channel.partner"
+        )
+        domain = [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
+        registry[channel_member_model].search(domain)[:1].with_context(ctx).fold_state = "open"
```
aj-fuentes, line 299: Isn't this going to fail if the search returns empty?

diagnoza: Should be better now

src/util/report.py
```diff
@@ -291,6 +291,21 @@ def ref(xid):
...
+    if version_between("9.0", "saas~18.2") and user.partner_id and recipient:
+        channel_member_model = (
+            "discuss.channel.member"
+            if version_gte("saas~16.3")
+            else "mail.channel.member"
+            if version_gte("16.0")
+            else "mail.channel.partner"
+        )
+        domain = [("partner_id", "=", user.partner_id.id), ("channel_id", "=", recipient.id)]
+        try:
```
aj-fuentes, line 304: Alternatively you can add an else to the try/except block above
```
else:
    # Chat window with the report will be open post-upgrade for the admin user
    if version_between("9.0", "saas~18.2") and user.partner_id and recipient:
        ...
```
If the message was post it's really unexpected that we fail here. Moreover if it failed there is no point in making it visible :/

diagnoza: @aj-fuentes why does it need to be in the `else` and not in the first `try` block? We still need this line wrapped in a `try/except`:

`registry[channel_member_model].search(domain)[:1].with_context(ctx).fold_state = "open"`

If we put it in the `else`, then I assume the point is to **not** catch any exception when it's executed?

aj-fuentes: When the posting failed this will fail _always_. That's what I meant. I didn't imply you should remove the other other try/except, you can keep it. In other words, why would we generate two errors in the logs if with the first one is already enough? OTOH if you put in the try/except above you may show a misleading error in the logs since maybe the posting succeeded but it was only the "opening" that failed. (that's what I meant that it is unexpected that this fail if the posting succeeded...)

diagnoza: Ah, I understand. I was confused by the use of _alternatively_ there :p

Xavier-Do:
Hello, it looks like it broke saas-18.1 and 18.2 upgrades

KangOl:
🤦 
Fixed by https://github.com/odoo/upgrade-util/pull/263

aj-fuentes:
So all this was pointless according to https://github.com/odoo/upgrade/pull/7056

> All existing chat window will be closed at upgrade, which is acceptable as they are temporary by nature.

diagnoza:
> So all this was pointless according to [odoo/upgrade#7056](https://github.com/odoo/upgrade/pull/7056)
> 
> > All existing chat window will be closed at upgrade, which is acceptable as they are temporary by nature.

I thought that we're well aware of it given that we had to add the upper bound for the version?

aj-fuentes:
I thought there was still a way to achieve that... to be done elsewhere. It looks like this is not really a thing anymore.

PR #221, opened 2025-02-28
[RFC][IMP] modules, models: remove inherits

Pirols:
When removing a module, all its models (and their fields) are also removed. However inherits of such models had to be cleaned up manually.


Pirols:
This is inspired by https://github.com/odoo/upgrade/pull/7280

Pirols:
upgradeci retry with always only account

Pirols:
In principle this should be the default behaviour. However that would break backwards compatibility with code that used to deal with the inherits after the removal of the module/model. 
The flags preserving worse-legacy-behavior are not ideal then, but probably necessary without a versioning system. Takes?

src/util/models.py
```diff
@@ -58,7 +58,7 @@ def _unknown_model_id(cr):
     return cr.fetchone()[0]
 
 
-def remove_model(cr, model, drop_table=True, ignore_m2m=()):
+def remove_model(cr, model, drop_table=True, remove_inherits=False, ignore_m2m=()):
```
aj-fuentes, line 61: Put the new parameter always at the end to avoid issues with existing code.

src/util/models.py
```diff
@@ -162,6 +163,10 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
 
     _remove_import_export_paths(cr, model)
 
+    if remove_inherits:
+        for inh in for_each_inherit(cr, model):
```
aj-fuentes, line 167: We could do this always and output a warning (with the name of the flag to set), advantages:
* In CI developers will know what to do
* In older versions, or client's upgrades, we will see in the log how widespread this issue is.

aj-fuentes (commented):
Since removing inherits may require, or have required, some extra steps, I believe that adding a new flag (False by default) is the best option.

src/util/models.py
```diff
@@ -164,6 +165,16 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
 
     _remove_import_export_paths(cr, model)
 
+    for inh in for_each_inherit(cr, model):
+        if remove_inherits:
+            remove_inherit_from_model(cr, inh.model, model)
```
jjmaksoud: Most of the time if one mixin is removed, it is replaced by its parent (if applicable) and removing the whole chain is not needed. If the parent model is also removed it will call this branch as well
```suggestion
            remove_inherit_from_model(cr, inh.model, model, with_inherit_parents=False)
```

PR #222, opened 2025-03-03
[IMP] util/misc: more literal replaces

aj-fuentes:
We want to avoid string replacement for code.


KangOl:

KangOl (approved):
@robodoo r+

PR #223, opened 2025-03-03
[IMP] specific: check author in `rename_custom_module`

Pirols:
When plenty of dbs exhibit the same custom (e.g. a `l10n`) module that needs to be renamed, we may use this util in generic scripts. In that case, it becomes paramount to make sure we only rename a specific module by a specific author. Actually, it would be better to do so in specific scripts too.

Also decrease the logger note level to info. It's not really a problem if the util is called against a db and the module is not found, perhaps the customer took care of the issue before submitting the new dump. Either way it's nothing to be concerned about.

### DIFF
--- a/src/util/specific.py
+++ b/src/util/specific.py
@@ -81,13 +81,13 @@ def rename_custom_model(cr, model_name, new_model_name, custom_module=None, repo
     )
 
 
-def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
-    cr.execute("SELECT 1 FROM ir_module_module WHERE name = %s", [old_module_name])
+def rename_custom_module(cr, old_module_name, new_module_name, report_details="", author="%"):
+    cr.execute("SELECT 1 FROM ir_module_module WHERE name = %s AND author ILIKE %s", [old_module_name, author])
     if not cr.rowcount:
-        _logger.warning("Module %r not found: skip renaming", old_module_name)
         return
 
     rename_module(cr, old_module_name, new_module_name)
+    _logger.warning("Custom module %r renamed to %r", old_module_name, new_module_name)
     add_to_migration_reports(
         category="Custom modules",
         message="The custom module '{old_module_name}' was renamed to '{new_module_name}'. {report_details}".format(


Pirols:
IMO this is good to have for specific scripts too, justifying the small complexity increase. @aj-fuentes

asno-odoo:
what about all the places using this util already without author name?

Pirols:
> I haven't seen enough examples to believe we need this. But anyway I'll let @KangOl decide. More [here](https://github.com/odoo/upgrade/pull/7310#discussion_r1977282688)

It's a simple util, we're never going to *need* this. There are 3 cases where this could have been used:

* https://github.com/odoo/upgrade/blob/2c71b4b9ec22f08764d405366e680ff1a8827349/migrations/base/saas~17.5.1.3/pre-10-modules.py#L175-L185
* https://github.com/odoo/upgrade/blob/2c71b4b9ec22f08764d405366e680ff1a8827349/migrations/base/16.0.1.3/pre-10-modules.py#L87-L97
* https://github.com/odoo/upgrade/blob/2c71b4b9ec22f08764d405366e680ff1a8827349/migrations/base/saas~14.5.1.3/pre-10-modules.py#L11-L21

Plus potentially https://github.com/odoo/upgrade/pull/7310. And as I said there, I think it would make sense to use this util more widely for recurring custom module removal in the name of reduced debugging time dedicated to those issues.

> what about all the places using this util already without author name?
> Edit: this method is used in lot of specific scripts

That's not super important for specific scripts. For old scripts we only want them to be kept backwards compatible.

src/util/specific.py
```diff
@@ -81,10 +81,10 @@ def rename_custom_model(cr, model_name, new_model_name, custom_module=None, repo
     )
 
 
-def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
-    cr.execute("SELECT 1 FROM ir_module_module WHERE name = %s", [old_module_name])
+def rename_custom_module(cr, old_module_name, new_module_name, report_details="", author="%"):
+    cr.execute("SELECT 1 FROM ir_module_module WHERE name = %s AND author ILIKE %s", [old_module_name, author])
     if not cr.rowcount:
-        _logger.warning("Module %r not found: skip renaming", old_module_name)
+        _logger.info("Module %r with author %r not found: skip renaming", old_module_name, author)
```
aj-fuentes, line 87: As I said, I don't think we should add a noisy line in the logs for something we don't care about. To avoid this line you would have to make a query anyway. Generic scripts are not meant to fix custom stuff. We do this only in exceptional cases (of which your last example doesn't look like being). An extra query for such cases is not an issue IMO.

aj-fuentes:
> There are 3 cases where this could have been used:

This is precisely my point: _could_ doesn't mean we _should_.

src/util/specific.py
Pirols: Logging one line is cheap. That said, I agree this can be improved. I moved the warning.

Pirols:
> > There are 3 cases where this could have been used:
> 
> This is precisely my point: _could_ doesn't mean we _should_.

Well I'm not sure that util existed when those 3 instances were introduced, otherwise I would say it should have been used (this patched version, anyways).

aj-fuentes:
> I would say it should have been used (this patched version, anyways).

My point is that you don't need to patch it to use it ;)

Pirols:
> > I would say it should have been used (this patched version, anyways).
> 
> My point is that you don't need to patch it to use it ;)

To the sole purpose of leveraging its `add_to_migration_reports` call? The value of this util (to me, at least) is that it is supposed to encode everything regarding the renaming of a **custom** module, which should include checking its `author` and logging about it.

src/util/specific.py
```diff
@@ -81,13 +81,13 @@ def rename_custom_model(cr, model_name, new_model_name, custom_module=None, repo
     )
 
 
-def rename_custom_module(cr, old_module_name, new_module_name, report_details=""):
-    cr.execute("SELECT 1 FROM ir_module_module WHERE name = %s", [old_module_name])
+def rename_custom_module(cr, old_module_name, new_module_name, report_details="", author="%"):
+    cr.execute("SELECT 1 FROM ir_module_module WHERE name = %s AND author ILIKE %s", [old_module_name, author])
     if not cr.rowcount:
-        _logger.warning("Module %r not found: skip renaming", old_module_name)
```
aj-fuentes, line 87: The idea of this line for specific scripts was to warn when a custom rename was not being applied. This could mean that more details in the specific upgrade needs to be addressed. Thus it is a warning to the people following the logs, client and/or developer.

src/util/specific.py
```diff
@@ -81,13 +81,13 @@ def rename_custom_model(cr, model_name, new_model_name, custom_module=None, repo
...
         return
 
     rename_module(cr, old_module_name, new_module_name)
+    _logger.warning("Custom module %r renamed to %r", old_module_name, new_module_name)
```
aj-fuentes, line 90: Since the util is doing what we instructed it to do it should be just info.
```suggestion
    _logger.info("Custom module %r renamed to %r", old_module_name, new_module_name)
```
... unless you want to transform this into something meant for generic upgrade which seem to be the root issue in the discussion.

aj-fuentes (commented):
The improvement looks well intended but I still see a conflict between a util meant for specific scripts and the generic upgrade. I won't push for more here. Below some comments.

src/util/specific.py
Pirols: Normally we want to rename a custom module because it otherwise conflicts and somehow causes the upgrade to fail. If the upgrade succeeds *usually* it does not matter whether the module was renamed or not and viceversa if it fails the presence of yet another warning message, will hardly be of any help to the debugger.

Notice I said *usually*. Because I'm sure there will be corner cases where we would care about that. Those are cases where I think your point would apply: utils should not be all encompassing. In those cases the developer will need to wrap it in more code to ensure the proper extra bits of logic are applied.

src/util/specific.py
aj-fuentes: > I think the warning level is appropriate here because there should be an inherent level of caution associated with renaming non-standard module.

We don't agree. For other reasons. There is always an inherent level of warning for almost everything we do in upgrades.

> IMO this shouldn't be addressed at specific or generic scripts. This should be the recipe to rename a custom module, period.

That seems like the goal of this PR. It was what I was referring above.

src/util/specific.py
aj-fuentes: I'm resolving this to move on: we agree in disagree.

aj-fuentes (approved):
This adds some value. Let's move on.

KangOl (approved):
upgradeci skip

@robodoo r+

PR #224, opened 2025-03-03
[FIX] records: use temp table in replacing refs

jjmaksoud:
For the jsonb company dependent indirect references when replacing record refs, building the query using the original id mapping input can generate a large query based on the number of references being updated, leading to a `stack depth limit exceeded` error.

We can use the mapping saved in the temp table `_upgrade_rrr` instead.

TBG-1752

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -2169,3 +2169,90 @@ class TestRenameXMLID(UnitTestCase):
         util.invalidate(test_view_2)
         self.assertIn('t-call="base.rename_view"', test_view_2.arch_db)
         self.assertIn('t-name="base.rename_view"', test_view_1.arch_db)
+
+
+class TestAssertUpdated(UnitTestCase):
+    def test_assert_updated(self):
+        p1 = self.env["res.partner"].create({"name": "Levi"})
+        p2 = self.env["res.partner"].create({"name": "Mikasa Ackerman"})
+        util.flush(p1)
+        util.flush(p2)
+
+        # when ids is None, assert any record is created or updated
+        with self.assertUpdated("res_partner"):
+            self.env["res.partner"].create({"name": "Sasha Braus"})
+        with self.assertUpdated("res_partner"):
+            p2.city = "Shiganshina"
+            util.flush(p2)
+        with self.assertRaises(AssertionError), self.assertUpdated("res_partner"):
+            pass
+
+        # when ids is [], assert a record is updated
+        with self.assertUpdated("res_partner", ids=[]):
+            p1.city = "Underground"
+            util.flush(p1)
+        with self.assertRaises(AssertionError), self.assertUpdated("res_partner", ids=[]):
+            self.env["res.partner"].create({"name": "Annie Leonhart"})
+
+        # when ids has multiple records, all records should be updated
+        with self.assertUpdated("res_partner", ids=[p1.id, p2.id]):
+            p1.company_name = "Survey Corps"
+            p2.company_name = "Survey Corps"
+            util.flush(p1)
+            util.flush(p2)
+        with self.assertRaises(AssertionError), self.assertUpdated("res_partner", ids=[p1.id, p2.id]):
+            p1.name = "Levi Ackerman"
+            util.flush(p1)
+
+        # when ids has a record, that record should be the one updated
+        with self.assertRaises(AssertionError), self.assertUpdated("res_partner", ids=[p1.id]):
+            p2.city = "Paradise Island"
+            util.flush(p2)
+
+    def test_assert_not_updated(self):
+        p1 = self.env["res.partner"].create({"name": "Eren Yeager"})
+        p2 = self.env["res.partner"].create({"name": "Armin Arlert"})
+        util.flush(p1)
+        util.flush(p2)
+
+        # when ids is None, assert no record is created or updated
+        with self.assertNotUpdated("res_partner"):
+            pass
+        with self.assertRaises(AssertionError), self.assertNotUpdated("res_partner"):
+            p1.city = "Shiganshina"
+            util.flush(p1)
+        with self.assertRaises(AssertionError), self.assertNotUpdated("res_partner"):
+            self.env["res.partner"].create({"name": "Bertolt Hoover"})
+
+        # when ids is [], assert no record is updated
+        with self.assertNotUpdated("res_partner", ids=[]):
+            self.env["res.partner"].create({"name": "Marco Bodt"})
+        with self.assertRaises(AssertionError), self.assertNotUpdated("res_partner", ids=[]):
+            p2.city = "Shiganshina"
+            util.flush(p2)
+
+        # when ids has a record, only that record should not be updated
+        with self.assertNotUpdated("res_partner", ids=[p2.id]):
+            p1.company_name = "Survey Corps"
+            util.flush(p1)
+
+        # when ids has multiple records, none of them should be updated
+        with self.assertRaises(AssertionError), self.assertNotUpdated("res_partner", ids=[p1.id, p2.id]):
+            p2.company_name = "Survey Corps"
+            util.flush(p2)
+
+    def test_assert_updated_combo(self):
+        p1 = self.env["res.partner"].create({"name": "Reiner Braun"})
+        p2 = self.env["res.partner"].create({"name": "Ymir Fritz"})
+        util.flush(p1)
+        util.flush(p2)
+
+        with self.assertUpdated("res_partner", ids=[p1.id]), self.assertNotUpdated("res_partner", ids=[p2.id]):
+            p1.company_name = "Marley Warriors"
+            util.flush(p1)
+
+        with self.assertRaises(AssertionError), self.assertUpdated("res_partner"), self.assertNotUpdated(
+            "res_partner", ids=[p2.id]
+        ):
+            p2.city = "Niflheim"
+            util.flush(p2)
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1460,6 +1460,51 @@ class TestRecords(UnitTestCase):
         [count] = self.env.cr.fetchone()
         self.assertEqual(count, 1)
 
+    @unittest.skipUnless(util.version_gte("18.0"), "Only work on Odoo >= 18")
+    def test_replace_record_references_batch__company_dependent(self):
+        partner_model = self.env["ir.model"].search([("model", "=", "res.partner")])
+        self.env["ir.model.fields"].create(
+            {
+                "name": "x_test_curr",
+                "ttype": "many2one",
+                "model_id": partner_model.id,
+                "relation": "res.currency",
+                "company_dependent": True,
+            }
+        )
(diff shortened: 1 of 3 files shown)


jjmaksoud:

aj-fuentes:
What examples do we have of such issue? Isn't this a problem of the caller to replace_records_reference_batch? Or that we should implement something perhaps splitting the id mapping?

src/util/records.py
```diff
@@ -1544,30 +1544,21 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
...
-                    ),
-                    [list(map(list, id_mapping.items()))],
-                ).decode()
-                explode_execute(cr, query, table=ir.table)
+                query = format_query(
+                    cr,
+                    """
+                     UPDATE {table} t
+                        SET {column} = (
+                                SELECT jsonb_object_agg(key, COALESCE((SELECT new FROM _upgrade_rrr WHERE old = value::int), value::int))
```
aj-fuentes, line 1552: I'm not sure a subquery here wouldn't end up making this painfully slow.

jjmaksoud:
i saw this so far only in documents upgrade when replacing folders with documents [here](https://github.com/odoo/upgrade/blob/6a3075e4c9d533a3e6a6fe15ec6d194eea2f7e82/migrations/documents/saas~17.5.1.4/pre-migrate.py#L193)
example request 2592340: has over 360,000 folder mappings and the new query was fast ( <2s on the test server)
I think the subquery should be okay because the number of companies is expected to be low, the explode bucket size is manageable, and the mapping table has an index

src/util/records.py
```diff
@@ -1544,30 +1544,21 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
...
+                                FROM jsonb_each_text(t.{column}))
+                    FROM _upgrade_rrr map
+                    WHERE t.{column} IS NOT NULL
+                        AND (SELECT * FROM JSONB_PATH_QUERY(t.{column}, '$.*'))::int = map.old
```
aj-fuentes, line 1556: I don't get how this work? Did you mean `map.old IN ...` Don't we get the error of casting jsonb null type to integer?

aj-fuentes: I think this is also updating the same row multiple times. I'd suggest we collect the info to update in a CTE to avoid that.

aj-fuentes: quick test:
```
> select 1 where (SELECT jsonb_path_query FROM JSONB_PATH_QUERY('{"1":3, "2":4, "3":null}'::jsonb, '$.*'))::int = 1
more than one row returned by a subquery used as an expression
```

jjmaksoud: seems the data i tried had a single company per record and no nulls, updated the syntax and tested
I also added values that will be overwritten if updated multiple times but they seem to be correct, checking more into that

KangOl:
Can you add a test?

src/util/records.py
```diff
@@ -1544,30 +1544,21 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
...
+                query = format_query(
+                    cr,
+                    """
+                     UPDATE {table} t
+                        SET {column} = (
+                                SELECT jsonb_object_agg(key, COALESCE((SELECT new FROM _upgrade_rrr WHERE old = value::int), value::int))
+                                FROM jsonb_each_text(t.{column}))
+                    FROM _upgrade_rrr map
+                    WHERE t.{column} IS NOT NULL
+                        AND map.old IN (SELECT JSONB_PATH_QUERY::int FROM JSONB_PATH_QUERY(t.{column}, '$.* ?(@ != null)'))
```
KangOl, line 1556: I also don't like those subqueries.
Here is a version using a CTE
```suggestion
                     WITH _upg_cd AS (
                         SELECT t.id, 
                                jsonb_object_agg(j.key, COALESCE(r.new, j.value::int)) as value
                           FROM {table} t
                           JOIN jsonb_each_text(t.{column}) j
                             ON true
                      LEFT JOIN _upgrade_rrr r 
                             ON r.old = j.value::integer
                          WHERE {{parallel_filter}}
                       GROUP BY t.id
                         HAVING bool_or(r.new IS NOT NULL)
                     )
                     UPDATE {table} t
                        SET {column} = u.value
                       FROM _upg_cd u
                      WHERE u.id = t.id
```

KangOl: I don't know how PG will behave with the `LEFT JOIN` after 2 tables in the `FROM`. At least, with a `JOIN ON true`, the behavior is clear.

jjmaksoud:
The test depends on https://github.com/odoo/odoo/pull/200509

jjmaksoud:
the community PR is merged and the test is now green

src/base/tests/test_util.py
```diff
@@ -1364,6 +1364,49 @@ def test_replace_record_references_batch__uniqueness(self):
...
+                "relation": "res.currency",
+                "company_dependent": True,
+            }
+        )
+        c1 = self.env["res.currency"].create({"name": "RC1", "symbol": "RC1"})
+        c2 = self.env["res.currency"].create({"name": "RC2", "symbol": "RC2"})
+        c3 = self.env["res.currency"].create({"name": "RC3", "symbol": "RC3"})
+
+        p1 = self.env["res.partner"].create({"name": "Captain Jack"})
+        p2 = self.env["res.partner"].create({"name": "River Song"})
```
KangOl, line 1384: To ensure the `replace_record_references_batch` don't do an UPDATE on all records, I would create a 3rd partner with a `x_test_curr` that shouldn't be updated (`{"4": c4.id}`) and then add a trigger on update of this row.

KangOl (changes requested):

src/base/tests/test_util.py
jjmaksoud: added the trigger, let me know if it's an acceptable testing utility

src/base/tests/test_util.py
```diff
@@ -1364,6 +1369,59 @@ def test_replace_record_references_batch__uniqueness(self):
...
+                """,
+                [value, id],
+            )
+        mapping = {
+            c1.id: c2.id,
+            c2.id: c3.id,
+            c3.id: c1.id,
+        }
+        add_test_trigger(self.env.cr, "replace_record_references", "res_partner", "UPDATE", f"new.id = {p3.id}")
+        util.replace_record_references_batch(self.env.cr, mapping, "res.currency")
```
KangOl, line 1413: This could be context manager
```suggestion
        with self.assertNotUpdated("res_partner", [p3.id]):
            util.replace_record_references_batch(self.env.cr, mapping, "res.currency")
```

Signature: `def assertNotUpdated(self, table, ids=None):`.
When ids is None, the whole should be handled, which mean no `INSERT`.

We can also add the positive version `def assertUpdated(self, table, ids=None)`

These functions need to be added in a dedicated commit.

KangOl: To be consistent with the other `assertX` functions, the signature needs to take a `msg` argument.

`assertNotUpdated(self, table, ids=None, msg=None)`

src/testing.py
```diff
@@ -122,6 +125,89 @@ def setUpClass(cls):
...
+            util.format_query(
+                cr,
+                """
+                CREATE TRIGGER no_update
+                BEFORE {when}
+                    ON {table}
+                FOR EACH ROW {cond} EXECUTE
+           FUNCTION fail_assert_not_updated()
+                """,
+                when=sql.SQL("UPDATE" if ids else "UPDATE or INSERT"),
```
KangOl, line 153:
```suggestion
                when=sql.SQL("UPDATE" if ids is not None else "UPDATE or INSERT"),
```

src/testing.py
```diff
@@ -122,6 +125,89 @@ def setUpClass(cls):
...
+                table=table,
+                cond=sql.SQL("WHEN (new.id IN %s)" if ids else ""),
```
KangOl, line 155:
```suggestion
                cond=sql.SQL("WHEN (new.id = ANY(%s))" if ids else ""),
```

src/testing.py
```diff
@@ -122,6 +125,89 @@ def setUpClass(cls):
...
+            ),
+            (tuple(ids) if ids else None,),
```
KangOl, line 157:
```suggestion
            [list(ids) if ids is not None else None],
```

src/testing.py
```diff
@@ -122,6 +125,89 @@ def setUpClass(cls):
...
+                BEFORE {when}
+                    ON {table}
+                FOR EACH ROW {cond} EXECUTE
+           FUNCTION save_updated()
+                """,
+                when=sql.SQL("UPDATE" if ids else "UPDATE or INSERT"),
+                table=table,
+                cond=sql.SQL("WHEN (NEW.id IN %s)" if ids else ""),
+            ),
+            (tuple(ids) if ids else None,),
```
KangOl, line 196: Ditto.

src/testing.py
```diff
@@ -122,6 +125,89 @@ def setUpClass(cls):
...
+        )
+        try:
+            yield
+            cr.execute("SELECT id FROM _upg_test_upd_id")
+            updated_ids = [r[0] for r in cr.fetchall()]
+            if not ids:
+                self.assertNotEqual(updated_ids, [], "No records were updated")
+            else:
+                for id in ids:
+                    self.assertIn(id, updated_ids, f"Record with id={id} was not updated")
```
KangOl, line 206: Raise directly `AssertionError` instead of relying on `self.assert*` functions.

src/testing.py
```diff
@@ -122,6 +125,89 @@ def setUpClass(cls):
...
+                """,
+                when=sql.SQL("UPDATE" if ids else "UPDATE or INSERT"),
+                table=table,
+                cond=sql.SQL("WHEN (new.id IN %s)" if ids else ""),
+            ),
+            (tuple(ids) if ids else None,),
+        )
+        try:
+            yield
+        finally:
```
KangOl, line 161: Not sure the right exception is raised.

KangOl (changes requested):
Maybe add tests for the new `assert` function to expect they behave correctly in all expected cases.

src/base/tests/test_util.py
```diff
@@ -1364,6 +1364,58 @@ def test_replace_record_references_batch__uniqueness(self):
...
+        }
+        for id, value in old.items():
+            self.env.cr.execute(
+                """
+                UPDATE res_partner
+                  SET x_test_curr = %s
+                WHERE id = %s
+                """,
+                [value, id],
+            )
```
KangOl, line 1401: Nit. Can be written on one line.
```suggestion
            self.env.cr.execute("UPDATE res_partner SET x_test_curr = %s WHERE id = %s", [value, id])
```

jjmaksoud:
Updated the assert methods, and ended up using a temp table for the assertNotUpdated as well instead of raising a psql exception.
Also added some test cases for the new assert functions. They can be used together, but the known limitation at the moment is that the same method cannot be used at the same time for a multiple tables.

KangOl (changes requested):

KangOl (approved):
@robodoo rebase-ff r+

d-fence:
robodoo retry
(failing test neutralized)

PR #225, opened 2025-03-04
[FIX] util/models: Fix compute for renamed model

sagu-odoo:
if model got renamed during migration but related field of model compute didn't update due to that below traceback raised for fixing it replaced with according this [patch](https://github.com/odoo/upgrade-util/blob/4230096526739591d15eb54652dc835ec8b3e3cc/src/util/models.py#L491)

```
'fleet.fleet_vehicle_menu', 521, 'Fleet > Fleet > Fleet', 726):
   File "/home/odoo/src/odoo/18.0/odoo/tools/safe_eval.py", line 397, in safe_eval
    return unsafe_eval(c, globals_dict, locals_dict)
   File "", line 2, in <module>
   File "/home/odoo/src/odoo/18.0/odoo/api.py", line 596, in __getitem__
    return self.registry[model_name](self, (), ())
   File "/home/odoo/src/odoo/18.0/odoo/modules/registry.py", line 240, in __getitem__
    return self.models[model_name]
 KeyError: 'stock.production.lot'
```

UPG-2588678
OPW-4592615


src/util/models.py
```diff
@@ -375,6 +375,22 @@ def rename_model(cr, old, new, rename_table=True):
     """.format(col_prefix=col_prefix, old=old.replace(".", r"\."), new=new)
     )
 
+    # adapt computes of manual fields
+    if column_exists(cr, "ir_model_fields", "compute"):
```
aj-fuentes, line 379: To avoid misalignment in the future if we need to update something in this query, could you extract this to a function (private, starting with `_`) please.

aj-fuentes (commented):
lgtm, small comment

src/util/models.py
sagu-odoo: suggested changes done. :t-rex:

src/util/models.py
```diff
@@ -375,6 +391,10 @@ def rename_model(cr, old, new, rename_table=True):
...
```
aj-fuentes, line 395: Move these two lines to the function as well.

aj-fuentes (approved):
@KangOl wdyt?

src/util/models.py
```diff
@@ -250,6 +250,24 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
 delete_model = remove_model
 
 
+def _replace_model_in_compute_field(cr, source, target):
```
KangOl, line 253:
```suggestion
def _replace_model_in_computed_custom_fields(cr, source, target):
```

KangOl:

KangOl (approved):
@robodoo r+

PR #226, opened 2025-03-05
[IMP] util.alter_column_type

KangOl:


src/util/pg.py
```diff
@@ -610,6 +614,15 @@ def remove_column(cr, table, column, cascade=False):
 
 
 def alter_column_type(cr, table, column, type, using=None, where=None, logger=_logger):
+    if where and not using:
```
aj-fuentes, line 617: I think we should add docs to this function to clarify the usage of each parameter. I feel we need something like "`where` is used to restrict the values converted via `using`"...

aj-fuentes: The drop of views should be documented if we add the docs.

aj-fuentes (approved):
LGTM

Docs would be better.

aj-fuentes (approved):

KangOl:
upgradeci retry with always only hr_holidays purchase_requisition

KangOl:
@robodoo rebase-ff r+

PR #229, opened 2025-03-06
[REF] odoo: remove test code from registry and sql_db

Williambraecky:
`Registry.enter_test_mode` has been removed in favor of a method on `BaseCase`.


KangOl:
upgradeci retry with always only base account stock

KangOl (approved):
@robodoo r+

PR #230, opened 2025-03-10
[IMP] util/records: use format_query

KangOl:
Avoid direct formatting and old-stye param expansion.


KangOl:
upgradeci retry with always only base hr sale_subscription

KangOl:
@robodoo r+

PR #231, opened 2025-03-11
[IMP] util/misc: dynamic literal_replace

aj-fuentes:
Allow to match in a glob-like fashion and produce the replacement node
dynamically.

Improvements:
* Do not replace if no match happens (allows to write less to the DB.)
* Do not buble syntax errors, log instead.
* Most valid nodes for Python expressions in the ORM are covered.
* Implement node matching in same order as the reference below.
* Much higher test coverage.
* Adapt tests for Python3.6 (Odoo 13 and 14)

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -7,6 +7,7 @@ import functools
 import logging
 import os
 import re
+import textwrap
 import uuid
 from contextlib import contextmanager
 from itertools import chain, islice
@@ -32,7 +33,13 @@ try:
     from ast import unparse as ast_unparse
 except ImportError:
     try:
-        from astunparse import unparse as ast_unparse
+        from astunparse import unparse as _ast_unparse
+
+        def ast_unparse(n):
+            # Older unparse adds weird new lines:
+            # >>> unparse(ast.parse("a",mode="eval").body)
+            # 'a\n'
+            return _ast_unparse(n).strip()
     except ImportError:
         ast_unparse = None
 
@@ -521,7 +528,7 @@ class SelfPrintEvalContext(collections.defaultdict):
                 return self._replace_node("_upg_UnaryOp_Not", node) if isinstance(node.op, ast.Not) else node
 
         replacer = RewriteName()
-        root = ast.parse(expr.strip(), mode="eval").body
+        root = ast.parse(expr.strip(), mode="exec")
         visited = replacer.visit(root)
         return (ast_unparse(visited).strip(), SelfPrintEvalContext(replacer.replaces))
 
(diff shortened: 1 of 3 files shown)


KangOl:

src/util/misc.py
```diff
@@ -510,47 +510,51 @@ def visit_UnaryOp(self, node):
...
+            key_ast = key if isinstance(key, ast.AST) else ast.parse(key, mode="eval").body
             self.mapping[key_ast.__class__.__name__].append((key_ast, value))
 
     def _no_match(self, left, right):
         return False
 
     def _match(self, left, right):
-        same_ctx = getattr(left, "ctx", None).__class__ is getattr(right, "ctx", None).__class__
+        same_ctx = getattr(left, "ctx", None).__class__ is getattr(right, "ctx", ast.Load()).__class__
+        same_ctx = True
```
KangOl, line 535: Is this diff wanted?

aj-fuentes: No, it's still a wip detail. Lemme check it once more.

src/util/misc.py
```diff
@@ -510,47 +510,51 @@ def visit_UnaryOp(self, node):
                 return self._replace_node("_upg_UnaryOp_Not", node) if isinstance(node.op, ast.Not) else node
 
         replacer = RewriteName()
-        root = ast.parse(expr.strip(), mode="eval").body
+        root = ast.parse(expr.strip(), mode="exec")
         visited = replacer.visit(root)
         return (ast_unparse(visited).strip(), SelfPrintEvalContext(replacer.replaces))
 
 
+GLOB_LIT = object()
```
KangOl, line 518: What about scoping it by being an attribute on the `literal_replace`
Also rename it to something more explicit.

```suggestion
literal_replace.WILDCARD = object()
```

aj-fuentes: Done. 
Subtleties:
1. `ast.Attributes` needs a string to unparse correctly
    ```py
    >>> ast.unparse(ast.Attribute(ast.Name("y"), object()))
      File "<input>", line 1, in <module>
        ast.unparse(ast.Attribute(ast.Name("y"), object()))
      File "/usr/lib/python3.12/ast.py", line 1791, in unparse
        return unparser.visit(ast_obj)
               ^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.12/ast.py", line 860, in visit
        return "".join(self._source)
               ^^^^^^^^^^^^^^^^^^^^^
    TypeError: sequence item 2: expected str instance, object found
    ```
2. Cannot really use `str is str`:
    ```py
    >>> x = "*"
    >>> y = "*"
    >>> x is y
    True
    ```
    ```py
    >>> class S(str):
    ...     pass
    ...
    >>> z = S("*")
    >>> z is x
    False
    ```
3. `ctx` is missing in manually created nodes:
    ```py
    >>> ast.parse("y.z", mode="eval").body.ctx
    <ast.Load object at 0x7c1cebe20bd0>
    >>> ast.Attribute(ast.Name("y"), "z").ctx
      File "<input>", line 1, in <module>
        ast.Attribute(ast.Name("y"), "z").ctx
    AttributeError: 'Attribute' object has no attribute 'ctx'
    ```

Thus for 3/ and the sake of less verbosity we assume missing ctx means sort of a wildcard context.

KangOl: Using `_Symbol` works when creating a `ast.Attribute` record.

Does it also for other `_match_X` methods where it is used?

I would like to see them also be covered by a test.

aj-fuentes: I will add tests for them.

aj-fuentes: Done. Note that this just a nice to have: good for debugging if we need to print the pattern.

aj-fuentes:
Still all green :D

KangOl (approved):
@robodoo rebase-ff r+

PR #234, opened 2025-03-18
[FIX] tests/test_util: add fallback to test with astunparse

aj-fuentes:
Old Python versions relying in astunparse will get a slightly different
expression with extra parenthesis.


KangOl (approved):
@robodoo r+

PR #235, opened 2025-03-19
[FIX] core: PEP 420 compatibility

xmo-odoo:
odoo/odoo#195664 removed explicit namespace support, as a result the `__init__` files break the ability to hook upgrade in via `PYTHONPATH`, as under PEP420 no namespace entry may have a`__init__` files within the shared namespace path.


aj-fuentes:
@KangOl I confirm it works while upgrading a DB from 7.0 up to 19.0 both locally and in the platform. If the CI is all good this looks good to go.

PR #236, opened 2025-03-19
[FIX] util.merge_model

KangOl:
Don't change the `res_model_id` of bound indirect references. Their `res_id` won't point to the correct record (if it even exists).

Those are records that haven't been remapped before the merge. They will be removed along the target model removal.

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -421,7 +421,10 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
     cr.execute("SELECT model, id FROM ir_model WHERE model in %s", ((source, target),))
     model_ids = dict(cr.fetchall())
     mapping = {model_ids[source]: model_ids[target]}
-    ignores = ["ir_model", "ir_model_fields", "ir_model_constraint", "ir_model_relation"]
+
+    ir_model_tables = ["ir_model", "ir_model_fields", "ir_model_constraint", "ir_model_relation"]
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only account documents hr_expense knowledge loyalty mrp product sale_subscription spreadsheet_edition survey website_slides

src/util/models.py
```diff
@@ -421,7 +421,10 @@ def merge_model(cr, source, target, drop_table=True, fields_mapping=None, ignore
     cr.execute("SELECT model, id FROM ir_model WHERE model in %s", ((source, target),))
     model_ids = dict(cr.fetchall())
     mapping = {model_ids[source]: model_ids[target]}
+
     ignores = ["ir_model", "ir_model_fields", "ir_model_constraint", "ir_model_relation"]
+    ignores.extend(ir.table for ir in indirect_references(cr, bound_only=True) if ir.res_model_id)
```
aj-fuentes, line 426: I'd prefer to not alter `ignores`. 

See, below in line 494
```py
if ir.res_model and not ir.res_id and ir.table not in ignores:
```
`bound_only=True` is the same as indirect references with `res_id` set, thus we are still ignoring them anyway.

IMO it's much better for debugging to keep the input `ignores` untouched and we know from experience that overriding variables could have unintended purposes.

KangOl:
@robodoo r+

PR #237, opened 2025-03-21
[ADD] util.get_common_columns

KangOl:
Function to get the ColumnList of common columns of two tables.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -31,6 +31,7 @@ except NameError:
 
 import psycopg2
 from psycopg2 import errorcodes, sql
+from psycopg2.extensions import quote_ident
 
 try:
     from odoo.modules import module as odoo_module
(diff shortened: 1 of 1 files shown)


KangOl:

src/util/pg.py
```diff
@@ -996,6 +996,11 @@ def __init__(self, list_=(), quoted=()):
         self._trailing_comma = False
         self._alias = None
 
+    @classmethod
+    def _new(cls, cr, list_=()):
```
aj-fuentes, line 1000: If we need an empty one we can always use the default `ColumnList()`. We could add some docs already :)
```suggestion
    def from_unquoted(cls, cr, list_):
        """
        Build a ColumnList item from a list of column names that may need quoting.
        
        ...
        """
```
🚲 🪛  on the name :D

aj-fuentes (commented):
The code anyway looks good to me.

KangOl:
@robodoo rebase-ff r+

PR #238, opened 2025-03-24
[FIX] ColumnList.using()

KangOl:


src/util/pg.py
```diff
@@ -1015,7 +1029,7 @@ def from_unquoted(cls, cr, list_):
         quoted = [quote_ident(c, cr._cnx) for c in list_]
         return cls(list_, quoted)
 
-    def using(self, leading_comma=False, trailing_comma=False, alias=None):
+    def using(self, leading_comma=KEEP_CURRENT, trailing_comma=KEEP_CURRENT, alias=KEEP_CURRENT):
```
aj-fuentes, line 1032: I think we could use "standard" (built-in) defaults. An empty alias cannot be confused with a valid alias ever.
```suggestion
    def using(self, leading_comma=None, trailing_comma=None, alias=""):
```

KangOl: I thought of this alternative, but I find `None` not really explicative of the behavior.

src/util/misc.py
```diff
@@ -39,8 +39,19 @@
...
 
+class Sentinel:
+    # waiting PEP661...
+    __slots__ = ("name",)
+
+    def __init__(self, name):
+        self.name = name
+
+    def __repr__(self):
+        return self.name
```
aj-fuentes, line 50: nice to have: in debugging sessions it's nice if this is explicitly printed as "Sentinel"
```suggestion
        return "Sentinel<{}>".format(self.name)
```

KangOl: Actually, it's for having it display nicely in the documentation: https://runbot131.odoo.com/runbot/static/build/77170216-master/logs/build/html/developer/reference/upgrades/upgrade_utils.html#odoo.upgrade.util.pg.ColumnList.using

image

KangOl:
upgradeci retry with always only base mrp

KangOl:
@robodoo rebase-ff r+

PR #239, opened 2025-03-26
[FIX] util/report: Fix unicode error

sagu-odoo:
fix unicode error
In python2 special character(``é``) with string formating gives error. So resolving accordingly using ``xmlcharrefreplace`` 

https://docs.python.org/2.7/howto/unicode.html#the-unicode-type
after fix 
image
```
  File "/home/odoo/src/odoo/10.0/odoo/service/server.py", line 974, in preload_registries
  File "/home/odoo/src/odoo/10.0/odoo/modules/registry.py", line 83, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/home/odoo/src/odoo/10.0/odoo/modules/loading.py", line 318, in load_modules
    report=report, models_to_check=models_to_check)
  File "/home/odoo/src/odoo/10.0/odoo/modules/loading.py", line 136, in load_module_graph
  File "/home/odoo/src/odoo/10.0/odoo/modules/migration.py", line 169, in migrate_module
  File "/home/odoo/src/odoo/10.0/odoo/addons/base/maintenance/migrations/base/0.0.0/pre-match-partner_company.py", line 53, in migrate
    for p_id, p_name, c_id, c_name in rows
  File "/home/odoo/src/odoo/10.0/odoo/addons/base/maintenance/migrations/base/0.0.0/pre-match-partner_company.py", line 53, in <genexpr>
    for p_id, p_name, c_id, c_name in rows
  File "/home/odoo/src/odoo/10.0/odoo/addons/base/maintenance/migrations/util/report.py", line 305, in get_anchor_link_to_record
    anchor_tag = '<a target="_blank" href="{}">{}</a>'.format(url, html_escape(name))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe8' in position 18: ordinal not in range(128)
```
upg-2642498

opw-4642911

### DIFF
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -2,6 +2,7 @@
 import logging
 import os
 import re
+import sys
 from textwrap import dedent
 
 import lxml
@@ -32,27 +33,30 @@ try:
     except ImportError:
         from odoo import SUPERUSER_ID
     from odoo import release
-    from odoo.tools import html_escape
     from odoo.tools.mail import html_sanitize
 except ImportError:
     from openerp import SUPERUSER_ID, release
     from openerp.tools.mail import html_sanitize
 
-    try:
-        from openerp.tools.misc import html_escape
-    except ImportError:
-        import werkzeug.utils
 
-        # Avoid DeprecationWarning while still remaining compatible with werkzeug pre-0.9
-        if parse_version(getattr(werkzeug, "__version__", "0.0")) < parse_version("0.9.0"):
+if sys.version_info > (3,):
+    from odoo.tools import html_escape
+else:
+    # In python2, `html_escape` always returns a byte-string with non-ascii characters replaced
+    # by their html entities.
 
-            def html_escape(text):
-                return werkzeug.utils.escape(text, quote=True)
+    import werkzeug.utils
 
-        else:
+    # Avoid DeprecationWarning while still remaining compatible with werkzeug pre-0.9
+    if parse_version(getattr(werkzeug, "__version__", "0.0")) < parse_version("0.9.0"):
 
-            def html_escape(text):
-                return werkzeug.utils.escape(text)
+        def html_escape(text):
+            return werkzeug.utils.escape(text, quote=True).encode("ascii", "xmlcharrefreplace")
+
+    else:
+
+        def html_escape(text):
+            return werkzeug.utils.escape(text).encode("ascii", "xmlcharrefreplace")
 
 
 try:


src/util/report.py
```diff
@@ -302,7 +302,9 @@ def get_anchor_link_to_record(model, id, name, action_id=None):
     else:
         url = "/web?debug=1#view_type=form&amp;model={}&amp;action={}&amp;id={}".format(model, action_id or "", id)
 
-    anchor_tag = '<a target="_blank" href="{}">{}</a>'.format(url, html_escape(name))
+    anchor_tag = '<a target="_blank" href="{}">{}</a>'.format(
+        url, html_escape(name).encode("ascii", "xmlcharrefreplace").decode("utf-8")
+    )
```
aj-fuentes, line 307: Isn't it more that we need to make the string unicode?
```suggestion
    anchor_tag = unicode('<a target="_blank" href="{}">{}</a>').format(url, html_escape(name))
```
```
>>> print(unicode("{}").format(x))
è
>>> print("{}".format(x))
  File "<input>", line 1, in <module>
    print("{}".format(x))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe8' in position 0: ordinal not in range(128)
```
cc: @KangOl

sagu-odoo: then I think  have to make [this](https://github.com/odoo/upgrade/blob/a7b641e29eddba23daad7102edfe4e6c9857bf21/migrations/base/0.0.0/pre-match-partner_company.py#L35) to unicode below traceback raised and further more unicode is not supported in python3 as I have tested 
```py
  File "/home/odoo/src/odoo/10.0/odoo/addons/base/maintenance/migrations/base/0.0.0/pre-match-partner_company.py", line 53, in migrate
    for p_id, p_name, c_id, c_name in rows
  File "/home/odoo/src/odoo/10.0/odoo/addons/base/maintenance/migrations/base/0.0.0/pre-match-partner_company.py", line 53, in <genexpr>
    for p_id, p_name, c_id, c_name in rows
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 108: ordinal not in range(128)
```
```py
odoo@odoo-sagu:~$ python3
Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> unicode("{}").format()
NameError: name 'unicode' is not defined
>>> 
```
please correct me if I am missing something

aj-fuentes: This parts makes the script work in python3
https://github.com/odoo/upgrade-util/blob/1109830b541cb00200f9234cd0c123ac52ceb9b8/src/util/report.py#L13-L17

And you need to run my example with python2, not python3
```
❯ python2 -m bpython
bpython version 0.20.1 on top of Python 2.7.18 /home/odoo/.local/bin/python2
WARNING: You are using `bpython` on Python 2. Support for Python 2 has been deprecated in version 0.19 and might disappear in a future version.
>>> x=u"\xe8"
>>> print(x)
è
>>> print(unicode("{}").format(x))
è
>>> print("{}".format(x))
  File "<input>", line 1, in <module>
    print("{}".format(x))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe8' in position 0:
ordinal not in range(128)
```

sagu-odoo: Yes, I tested  the suggestion in Python 2, but I also tested it in Python 3 for safety. As I mentioned earlier, even with ``unicode`` , the error still occurs https://github.com/odoo/upgrade-util/pull/239#discussion_r2013717416

KangOl: Shouldn't actually wrap all literal strings with `unicode` in the code?

aj-fuentes: Yes, that's the detail I wanted to avoid :/ I'm trying to understand if there is really any downside in using the encoding since supposedly the output of get_anchor_link_to_record should be used in html always, thus escaped chars should display correctly...

aj-fuentes: So, after checking this with the failing case, I agree with the first solution using `xmlcharrefreplace` but _only_ for Python2. That way we ensure we do not need to wrap all messages in 0.0.0 as unicode while at the same time keeping the replacement and xml encoding happening only in older versions (10.0 in practice, since 9.0 doesn't run 0.0.0 scripts)

aj-fuentes: This works
```diff
diff --git a/src/util/report.py b/src/util/report.py
index b6789b2..49bc19b 100644
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -13,8 +13,10 @@ from .misc import parse_version
 # python3 shims
 try:
     basestring  # noqa: B018
+    def unicode_escape(u):
+        return u.encode("ascii", "xmlcharrefreplace")
 except NameError:
-    basestring = unicode = str
+    basestring = unicode = unicode_escape = str

 try:
     from markupsafe import Markup, escape
@@ -302,7 +304,7 @@ def get_anchor_link_to_record(model, id, name, action_id=None):
     else:
         url = "/web?debug=1#view_type=form&amp;model={}&amp;action={}&amp;id={}".format(model, action_id or "", id)

-    anchor_tag = '<a target="_blank" href="{}">{}</a>'.format(url, html_escape(name))
+    anchor_tag = '<a target="_blank" href="{}">{}</a>'.format(url, unicode_escape(html_escape(name)))
     if Markup:
         anchor_tag = Markup(anchor_tag)
     return anchor_tag
```

image

KangOl: Shouldn't it be `html_escape` that do the ... escaping?

aj-fuentes (commented):
LGTM

Still, I'm OK with merging this into `html_escape` for python2. @KangOl ?

sagu-odoo:
upgrade retry with base

sagu-odoo:
upgradeci retry with base

KangOl:

sagu-odoo:
@KangOl is this good to go ?.

KangOl:

KangOl (approved):
@robodoo r+

PR #240, opened 2025-03-31
[IMP] util.update_record_from_xml

KangOl:
Make the default value of the `force_create` argument dynamic depending on the `fields` argument.


KangOl:
@robodoo r+

PR #241, opened 2025-04-01
[IMP] util,base: use cursor's mangled names

tbe-odoo:
- The psycopg2 technical objects are now mangled inside the Odoo cursors. This is made to avoid using them improperly. So we must adapt the code to use new names.

Related to:

https://github.com/odoo/odoo/pull/204273
https://github.com/odoo/enterprise/pull/82655
https://github.com/odoo/upgrade/pull/7478

### DIFF
--- a/src/base/0.0.0/end-no-respawn-fields.py
+++ b/src/base/0.0.0/end-no-respawn-fields.py
@@ -20,7 +20,7 @@ def migrate(cr, version):
     """
     )
     execute_values(
-        cr._obj,
+        getattr(cr, '_obj__', None) or cr._obj,
         "INSERT INTO no_respawn(model, field) VALUES %s",
         # fmt:off
         [
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -53,6 +53,7 @@ from .pg import (
     alter_column_type,
     column_exists,
     column_type,
+    cursor_get_connection,
     explode_execute,
     explode_query_range,
     format_query,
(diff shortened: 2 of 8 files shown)


KangOl:

src/base/0.0.0/end-no-respawn-fields.py
```diff
@@ -20,7 +20,7 @@ def migrate(cr, version):
     """
     )
     execute_values(
-        cr._obj,
+        getattr(cr, '_obj__', None) or cr._obj,
```
KangOl, line 23:
```suggestion
        cr,
```

src/util/fields.py
```diff
@@ -1484,7 +1485,7 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
         )
     else:
         psycopg2.extras.execute_values(
-            cr._obj,
+            getattr(cr, '_obj__') or cr._obj,
```
KangOl, line 1488:
```suggestion
            cr,
```

src/util/records.py
```diff
@@ -1476,7 +1476,7 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
         ignores.append("ir_model_data")
 
     cr.execute("CREATE UNLOGGED TABLE _upgrade_rrr(old int PRIMARY KEY, new int)")
-    execute_values(cr, "INSERT INTO _upgrade_rrr (old, new) VALUES %s", id_mapping.items())
+    execute_values(getattr(cr, '_obj__', cr), "INSERT INTO _upgrade_rrr (old, new) VALUES %s", id_mapping.items())
```
KangOl, line 1479:
```suggestion
    execute_values(cr, "INSERT INTO _upgrade_rrr (old, new) VALUES %s", id_mapping.items())
```

KangOl (changes requested):

PR #243, opened 2025-04-04
[FIX] util/records: deduplicate before batch update

aj-fuentes:
When updating record references we need to dedup uniq indexes. In the case of inderect references the case when two old references become the same new reference was missing.

Example to trigger the issue
```
test_15=> select * from mail_followers where res_model='res.partner.bank'
+----+------------------+--------+------------+
| id | res_model        | res_id | partner_id |
|----+------------------+--------+------------|
| 5  | res.partner.bank | 2      | 3          |
| 7  | res.partner.bank | 3      | 3          |
+----+------------------+--------+------------+
test_15=> select id,sanitized_acc_number,partner_id,company_id from res_partner_bank
+----+----------------------+------------+------------+
| id | sanitized_acc_number | partner_id | company_id |
|----+----------------------+------------+------------|
| 1  | 1                    | 3          | 1          |
| 2  | 1                    | 3          | 2          |
| 3  | 1                    | 3          | 3          |
+----+----------------------+------------+------------+
```

In such case the record replacer will replace references to the partner bank as in the mapping `{2:1, 3:1}`. But this causes an error
```
2025-03-07 17:12:12,145 502959 ERROR test_15_16 odoo.sql_db: bad query:

            UPDATE "mail_followers" t
               SET "res_model" = 'res.partner.bank'
                   , "res_id" = _upgrade_rrr.new
              FROM _upgrade_rrr
             WHERE "res_model"='res.partner.bank'
               AND _upgrade_rrr.old = res_id

                    AND
                        NOT EXISTS(SELECT 1 FROM mail_followers WHERE "res_model" = 'res.partner.bank' AND "res_id" = _upgrade_rrr.new AND "partner_id"=t."partner_id")
                    ;
                DELETE FROM mail_followers USING _upgrade_rrr WHERE "res_model"='res.partner.bank' AND res_id = _upgrade_rrr.old;

ERROR: duplicate key value violates unique constraint "mail_followers_mail_followers_res_partner_res_model_id_uniq"
DETAIL:  Key (res_model, res_id, partner_id)=(res.partner.bank, 1, 3) already exists.
```

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -389,6 +389,62 @@ class TestAdaptDomainView(UnitTestCase):
         self.assertIn("courriel", view_search_2.arch)
 
 
+@unittest.skipUnless(
+    util.version_gte("13.0"), "This test is incompatible with old style odoo.addons.base.maintenance.migrations.util"
+)
+class TestReplaceReferences(UnitTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env.cr.execute(
+            """
+            CREATE TABLE dummy_model(
+                             id serial PRIMARY KEY,
+                             res_id int,
+                             res_model varchar,
+                             extra varchar,
+                             CONSTRAINT uniq_constr UNIQUE(res_id, res_model, extra)
+                         );
+
+            INSERT INTO dummy_model(res_model, res_id, extra)
+                 VALUES -- the target is there with same res_id
+                        ('res.users', 1, 'x'),
+                        ('res.partner', 1, 'x'),
+
+                        -- two with same target and the target is there
+                        ('res.users', 2, 'x'),
+                        ('res.users', 3, 'x'),
+                        ('res.partner', 2, 'x'),
+
+                        -- two with same target and the target is not there
+                        ('res.users', 4, 'x'),
+                        ('res.users', 5, 'x'),
+
+                        -- target is there different res_id
+                        ('res.users', 6, 'x'),
+                        ('res.partner', 4, 'x')
+            """
+        )
+
+    def _ir_dummy(self, cr, bound_only=True):
+        yield util.IndirectReference("dummy_model", "res_model", "res_id")
+
+    def test_replace_record_references_batch__full_unique(self):
+        cr = self.env.cr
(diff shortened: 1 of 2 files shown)


KangOl:
upgradeci retry with always only base hr_recruitment sale_subscription helpdesk

src/base/tests/test_util.py
```diff
@@ -389,6 +389,69 @@ def test_adapt_domain_view(self):
...
+                        ('res.users', 6, 'x'),
+                        ('res.partner', 4, 'x')
+            """
+        )
+        self.orig_irs = list(util.INDIRECT_REFERENCES)
+        ir = util.IndirectReference("dummy_model", "res_model", "res_id")
+        util.INDIRECT_REFERENCES[:] = [ir]
+
+    def tearDown(self):
+        util.INDIRECT_REFERENCES[:] = self.orig_irs
```
KangOl, line 428: I don't like monkey patching the global list (which may be considered as an implementation detail)

You may just mock the `util.indirect_references()` function

```python
def ir_dummy(cr, bound_only=True):
    yield util.IndirectReference("dummy_model", "res_model", "res_id")

with mock.patch("odoo.upgrade.util.indirect_references.indirect_references", ir_dummy):
    util.replace_record_references_batch(...)
```

src/base/tests/test_util.py
```diff
@@ -389,6 +389,57 @@ def test_adapt_domain_view(self):
...
+            """
+        )
+
+    def _ir_dummy(self, cr, bound_only=True):
+        yield util.IndirectReference("dummy_model", "res_model", "res_id")
+
+    def test_replace_record_references_batch__full_unique(self):
+        mapping = {1: 1, 2: 2, 3: 2, 4: 3, 5: 3, 6: 4}
+        with mock.patch("odoo.addons.base.maintenance.migrations.util.records.indirect_references", self._ir_dummy):
+            util.replace_record_references_batch(self.cr, mapping, "res.users", "res.partner")
```
KangOl, line 430: Because of the dynamic nature of the `odoo.addons.base.maintenance.migrations` package, we cannot mock it.

So we need to mock the `odoo.upgrade` version and limit the test to versions >= 13

KangOl (approved):
@robodoo rebase-ff r+

PR #253, opened 2025-04-14
[IMP] util/records: skip replace if all ids are the same

aj-fuentes:
If all ids are the same and we are replacing in the same model we
currently cause an error. This is not necessary and could lead to false
errors if the list of mapping ids is correctly the same.

For example sometimes we transform a model A into model B by updating
metadata. Then we insert into B's table, and finally perform a replace
of references from A's table to B's table. This would fail if the ids
are the same. This case will happen with a high probability if there is
only one A record with id=1 since the single B record would get id=1 as
well.

Example logs after the patch:
```
2025-04-14 07:28:32,883 42811 WARNING test_16 odoo.upgrade.util.records: Replace references in model `res.country`, ignoring same-id mapping `{1: 1, 2: 2}`
2025-04-14 07:28:32,883 42811 WARNING test_16 odoo.upgrade.util.records: Nothing to replace in model `res.country`, ignoring empty mapping
```


src/util/records.py
```diff
@@ -1469,6 +1469,10 @@ def replace_record_references_batch(cr, id_mapping, model_src, model_dst=None, r
         if same_ids:
             _logger.warning("Replace references in model `%s`, ignoring same-id mapping `%s`", model_src, same_ids)
             id_mapping = {k: v for k, v in id_mapping.items() if k != v}
+            if not id_mapping:
+                if not id_mapping:
+                    _logger.warning("Nothing to replace in model `%s`, ignoring empty mapping", model_src)
```
KangOl, line 1474:
```suggestion
            if not id_mapping:
                if not id_mapping:
                    if not id_mapping:
                        _logger.warning("Nothing to replace in model `%s`, ignoring empty mapping", model_src)
```

aj-fuentes: rebase mishap :/

PR #258, opened 2025-04-16
[IMP] util.remove_field

aj-fuentes:
Add option to keep binary data as attachment.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -50,6 +50,7 @@ from .inherit import for_each_inherit
 from .misc import SelfPrintEvalContext, log_progress, version_gte
 from .orm import env, invalidate
 from .pg import (
+    SQLStr,
     alter_column_type,
     column_exists,
     column_type,
@@ -182,7 +183,7 @@ def _remove_field_from_context(context, fieldname):
     return changed
 
 
-def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inherit=()):
+def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inherit=(), keep_as_attachments=False):
     """
     Remove a field and its references from the database.
 
@@ -196,6 +197,8 @@ def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inh
     :param bool drop_column: whether the field's column is dropped
     :param list(str) or str skip_inherit: list of inheriting models to skip the removal
                                           of the field, use `"*"` to skip all
+    :param bool keep_as_attachments: for binary fields, whether the data should be kept
+                                     as attachments
     """
     _validate_model(model)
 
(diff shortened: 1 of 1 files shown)


src/util/fields.py
```diff
@@ -182,7 +183,7 @@ def filter_value(key, value):
     return changed
 
 
-def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inherit=()):
+def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inherit=(), keep_attachments=False):
```
aj-fuentes, line 186: `keep_attachments` sounds like if we were keeping _existing_ attachments
```suggestion
def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inherit=(), keep_as_attachments=False):
```

src/util/fields.py
```diff
@@ -196,6 +197,7 @@ def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inh
     :param bool drop_column: whether the field's column is dropped
     :param list(str) or str skip_inherit: list of inheriting models to skip the removal
                                           of the field, use `"*"` to skip all
+    :param bool keep_attachments: for binary fields, whether the data should be kept as attachments.
```
aj-fuentes, line 200: Ideally we should align the text to the max column (not done below due to limitations of GH review textarea)
```suggestion
    :param bool keep_as_attachments: for binary fields, whether the data should be kept as attachments
```

src/util/fields.py
```diff
@@ -353,8 +355,20 @@ def adapter(leaf, is_or, negated):
             defaults.pop(fieldname, None)
             cr.execute("UPDATE mail_alias SET alias_defaults = %s WHERE id = %s", [repr(defaults), alias_id])
 
-    # if field was a binary field stored as attachment, clean them...
-    if column_exists(cr, "ir_attachment", "res_field"):
+    table = table_of_model(cr, model)
+    # NOTE table_exists is needed to avoid altering views
+    stored = table_exists(cr, table) and column_exists(cr, table, fieldname)
+
+    if keep_attachments:
```
aj-fuentes, line 362:
```suggestion
    if keep_as_attachments:
```

src/util/fields.py
```diff
@@ -367,13 +381,20 @@ def adapter(leaf, is_or, negated):
...
     for inh in for_each_inherit(cr, model, skip_inherit):
-        remove_field(cr, inh.model, fieldname, cascade=cascade, drop_column=drop_column, skip_inherit=skip_inherit)
+        remove_field(
+            cr,
+            inh.model,
+            fieldname,
+            cascade=cascade,
+            drop_column=drop_column,
+            skip_inherit=skip_inherit,
+            keep_attachments=keep_attachments,
```
aj-fuentes, line 396:
```suggestion
            keep_as_attachments=keep_as_attachments,
```

src/util/fields.py
```diff
@@ -884,14 +905,39 @@ def _convert_field_to_property(
...
-    name_query = "COALESCE({0}, '{1}('|| id || ').{2}')".format(
-        name_field if name_field else "NULL",
-        model.title().replace(".", ""),
-        field,
+    name_query = cr.mogrify(
+        format_query(cr, "COALESCE({}, CONCAT(%s, '(', id,  ').', %s)", name_field if name_field else SQLStr("NULL")),
+        [model.title().replace(".", ""), field],
+    ).decode()
+
+    if not column_exists(cr, "ir_attachment", "res_field"):
```
aj-fuentes, line 924: Hm... was it a mistake in the previous implementation? We always assumed res_field was there.

KangOl: No. This function was explicitly written for handling the storage of the binary fields (`bytea` to `attachment`), and it's not possible without the `res_field` column.

Now that we repurpose this function, we need to handle the case.

KangOl:
upgradeci retry with always only hr marketing_card

KangOl (approved):
@robodoo r+

PR #259, opened 2025-04-16
[IMP] util/fields: skip convert if the column is missing

aj-fuentes:
It helps for RR versions where we currently need to pre-check if the
field is already created.


KangOl:

PR #264, opened 2025-04-18
[FIX] util/domains: ast.Constant doesn't exist in Python2

aj-fuentes:
Exaple error:
```
2025-04-18 16:00:25,979 531760 ERROR test_9_10 odoo.modules.migration: module base: Each pre-migration file must have a "migrate(cr, installed_version)" function
```

The problem is that the AttributeError in Odoo 10 hides the real issue:
```
'module' object has no attribute 'Constant'
```
See:
https://github.com/odoo/odoo/blob/28c3f51c4878fbcd79b2e819948465fcf2160ebc/odoo/modules/migration.py#L166-L167

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -102,7 +102,13 @@ except ImportError:
     del expression  # unset so it's not used directly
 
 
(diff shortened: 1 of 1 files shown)


aj-fuentes:
upgradeci retry with always account

KangOl (approved):
@robodoo priority r+

PR #265, opened 2025-04-23
[IMP] adapt code to new simplified `safe_eval` API

KangOl:
`safe_eval` loose the `nocopy` option. As we rely on it when evaluating expression, we need to reimplement `safe_eval` to do the evaluation using the `SelfPrintEvalContext`.

See odoo/odoo#206846
See odoo/enterprise#83818

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -12,6 +12,7 @@ import base64
 import logging
 import re
 import warnings
+from ast import literal_eval
 
 import psycopg2
 from psycopg2 import sql
@@ -25,11 +26,9 @@ except ImportError:
 try:
     from odoo import release
     from odoo.tools.misc import mute_logger
-    from odoo.tools.safe_eval import safe_eval
 except ImportError:
     from openerp import release
     from openerp.tools.misc import mute_logger
-    from openerp.tools.safe_eval import safe_eval
 
 from .domains import FALSE_LEAF, TRUE_LEAF
 
@@ -47,7 +46,7 @@ from .domains import _adapt_one_domain, _replace_path, _valid_path_to, adapt_dom
 from .exceptions import SleepyDeveloperError
 from .helpers import _dashboard_actions, _validate_model, resolve_model_fields_path, table_of_model
 from .inherit import for_each_inherit
-from .misc import SelfPrintEvalContext, log_progress, version_gte
+from .misc import log_progress, safe_eval, version_gte
 from .orm import env, invalidate
 from .pg import (
     SQLStr,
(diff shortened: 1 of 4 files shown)


KangOl:

src/util/misc.py
```diff
@@ -533,6 +533,40 @@ def visit_UnaryOp(self, node):
...
+        from odoo.tools.safe_eval import safe_eval as _safe_eval_func
+    except ImportError:
+        from openerp.tools.safe_eval import safe_eval as _safe_eval_func
+
+    def safe_eval(expr, context=None):
+        if context is None:
+            context = SelfPrintEvalContext()
+        assert isinstance(context, SelfPrintEvalContext)
+
+        return _safe_eval_func(expr, context, nocopy=True)
```
aj-fuentes, line 567: Why don't we assert in both cases that expr is str ot bytes?

If we can assert it in both then I'd suggest we move the shared part to a toplevel `safe_eval` that then calls `_safe_eval_impl` with bot expr and context mandatory. Sort of:
```py
if version_gte("saas~18.4"):
    import odoo.tools.safe_eval as _safe_eval_mod

    def _safe_eval_impl(expr, context):
        c = _safe_eval_mod.test_expr(expr, _safe_eval_mod._SAFE_OPCODES, mode="eval", filename=None)
        context["__builtins__"] = dict(_safe_eval_mod._BUILTINS)
        try:
            return _safe_eval_mod.unsafe_eval(c, context, None)
        except _safe_eval_mod._BUBBLEUP_EXCEPTIONS:
            raise
        except Exception as e:
            raise ValueError("{!r} while evaluating\n{!r}".format(e, expr))
        finally:
            del context["__builtins__"]
else:
    try:
        from odoo.tools.safe_eval import safe_eval as _safe_eval_func
    except ImportError:
        from openerp.tools.safe_eval import safe_eval as _safe_eval_func

    def _safe_eval_impl(expr, context):
        return _safe_eval_func(expr, context, nocopy=True)

def safe_eval(expr, context=None):
    if context is None:
        context = SelfPrintEvalContext()

    assert isinstance(expr, (str, bytes))
    assert isinstance(context, SelfPrintEvalContext)
    return _safe_eval_impl(expr, context)
```

src/util/fields.py
```diff
@@ -350,7 +349,7 @@ def adapter(leaf, is_or, negated):
         )
         for alias_id, defaults_s in cr.fetchall():
             try:
-                defaults = dict(safe_eval(defaults_s))  # XXX literal_eval should works.
+                defaults = dict(literal_eval(defaults_s))
```
aj-fuentes, line 352: Potential change of behaviour, LGTM just flagging it it new errors start appearing after the merge.

KangOl:
@robodoo r+

src/util/misc.py
antonrom1: Actually compile already checks the type of the source [here](https://github.com/python/cpython/blob/76e911a040aaf16f0c76b7d2a2f29582cb98c32d/Python/bltinmodule.c#L967). It's can either be a string, bytesarray or AST. We already check that the source object is not a codeobject, so I think the `assert isinstance(expr, (str, bytes))` is not really needed here

PR #266, opened 2025-04-25
[FIX] util/models: remove crons before server actions

jjmaksoud:
When removing the indirect references for a model, the records for `ir.cron` should be removed before the `ir.actions.server` because of the `on_delete=restrict` fk to the server actions. The direct IR on `ir.cron` was removed since v11, but the FK blocks the removal of server action records. Removing the crons should be done explicitly.

upg-2455357

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -135,7 +135,13 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
             # remove in batch
             size = (cr.rowcount + chunk_size - 1) / chunk_size
             it = chunks([id for (id,) in cr.fetchall()], chunk_size, fmt=tuple)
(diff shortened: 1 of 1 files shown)


KangOl:

jjmaksoud:
> It's already be handled in `remove_records`.

I'm guessing you mean [here](https://github.com/odoo/upgrade-util/blob/bdab4ced0060a1a1970630f1a14754d5a1f2ed91/src/util/records.py#L406), which now that I checked works in versions < 17.0 but when we get the inheitance data from `ir_model_inherit` the link between server actions and ir.cron doesnt exist. looking further into that

KangOl:
The error come from an error in Odoo where the `ir_model_inherit` data is not correctly computed. Fixed by https://github.com/odoo/odoo/pull/199109

PR #267, opened 2025-04-30
[FIX] remove_inherit_from_model

KangOl:
Use default `bucket_size` when deleting from indirect references tables. Those tables contains a lot of data (`mail_message` can be huge) and using a low bucket size can make the deletion slower.


KangOl:
upgradeci retry with always only account account_accountant base documents hr_timesheet mail maintenance mrp

PR #156, opened 2024-10-24
[IMP] util/fields: update default when converting to html

diagnoza:
One could use Python escape sequences in text fields. When converting a field to HTML, they will no longer work, so replace them with HTML tags.

Before:
```
+--------------------------------------------------------------------------------------------------------------------------------------+
| json_value                                                                                                                           |
|--------------------------------------------------------------------------------------------------------------------------------------|
| "Mise en place du chantier\n\nIntervention pour une mise en arrêt de la chaudière\n\nVérification responsable\nRepli fin de travaux" |
+--------------------------------------------------------------------------------------------------------------------------------------+
```
After:
```
+-----------------------------------------------------------------------------------------------------------------------------------------------------+
| json_value                                                                                                                                          |
|-----------------------------------------------------------------------------------------------------------------------------------------------------|
| "Mise en place du chantier<br/><br/>Intervention pour une mise en arrêt de la chaudière<br/><br/>Vérification responsable<br/>Repli fin de travaux" |
+-----------------------------------------------------------------------------------------------------------------------------------------------------+
```

[opw-4133947](https://www.odoo.com/odoo/project/6443/tasks/4133947)

PR #268, opened 2025-05-09
[FIX] util/pg: Fix escaping issue

sagu-odoo:
due to this escaping issue ``\n`` never converted
to ``<br>`` so fixing it. as per psql documentation 
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE

before fix:
```sql
website_crm=> select description from crm_lead where id=25;
 description
-------------
 test       +
 test       +
 etest      +

(1 row)

website_crm=> \q
```
after fix:

```sql
website_crm_18.0=> select description from crm_lead where id=25;
           description
----------------------------------
 <p>test<br>test<br>etest<br></p>
(1 row)
```
upg-2709886
opw-4781422

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -2109,13 +2109,13 @@ class TestConvertFieldToHtml(UnitTestCase):
         f1 = self.env["ir.model.fields"].create(
             {"name": "x_testx", "model": "res.partner", "ttype": "text", "model_id": model.id, "translate": True}
         )
-        partner = self.env["res.partner"].create({"name": "test Pxtner", "x_testx": "test partner field"})
-        default = self.env["ir.default"].create({"field_id": f1.id, "json_value": '"Test text"'})
+        partner = self.env["res.partner"].create({"name": "test Pxtner", "x_testx": "test\npartner field"})
+        default = self.env["ir.default"].create({"field_id": f1.id, "json_value": '"Test\\ntext"'})
         util.convert_field_to_html(cr, "res.partner", "x_testx")
         util.invalidate(default)
 
-        self.assertEqual(default.json_value, '"<p>Test text</p>"')
-        self.assertEqual(partner.x_testx, "<p>test partner field</p>")
+        self.assertEqual(default.json_value, '"<p>Test<br>text</p>"')
+        self.assertEqual(partner.x_testx, "<p>test<br>partner field</p>")
 
 
 class TestRemoveView(UnitTestCase):
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -428,9 +428,9 @@ def pg_text2html(s, wrap="p"):
                                            'https?://[-A-Za-z0-9+&@#/%%?=~_()|!:,.;]*[-A-Za-z0-9+&@#/%%=~_()|]',
                                            '<a href="\&" target="_blank" rel="noreferrer noopener">\&</a>',
                                            'g'),
-                            E'\\n',
+                            E'\n',
                             '<br>'),
-                    E'\\t',
+                    E'\t',
                     '&Tab;'),
                 '{closing_tag}')
          END
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -698,10 +698,10 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
 
     # Update ir.default
     if table_exists(cr, "ir_default"):
-        json_value_html = pg_text2html("json_value::json #>> '{}'")
+        json_value_html = pg_text2html("json_value::jsonb->>0")
         query = """
                 UPDATE ir_default AS d
-                   SET json_value = to_json(({})::text)::text
+                   SET json_value = to_json({})::text
                   FROM ir_model_fields AS imf
                  WHERE imf.name = %s
                    AND imf.model = %s


aj-fuentes:
I think this was intended in https://github.com/odoo/upgrade-util/pull/156

To be sure we aren't breaking it add a test for the specific case in https://www.odoo.com/odoo/project/6443/tasks/4133947 where we have `\n` in json_value in ir_default

src/util/pg.py
```diff
@@ -428,7 +428,7 @@ def pg_text2html(s, wrap="p"):
                                            'https?://[-A-Za-z0-9+&@#/%%?=~_()|!:,.;]*[-A-Za-z0-9+&@#/%%=~_()|]',
                                            '<a href="\&" target="_blank" rel="noreferrer noopener">\&</a>',
                                            'g'),
-                            E'\\n',
+                            E'\n',
                             '<br>'),
                     E'\\t',
```
KangOl, line 433: Same error here
```suggestion
                    E'\t',
```

sagu-odoo: as I have checked tab is not possible to add directly in text field but from defaults we can add so for that this is fine

aj-fuentes: For the record: I think this isn't going to address the original issue which was a literal `\n` in a json value. Actually the `E` seems to be not necessary for the original intent. The PR was kind of hacked into doing something else.
Example in 14.0 (fresh db with only `account` installed):
image
image
```
test_14=> select id,json_value from ir_default
+----+-----------------------+
| id | json_value            |
|----+-----------------------|
| 1  | "en_US"               |
| 3  | "line1\nline2\nline3" |
+----+-----------------------+
```

sagu-odoo: I am testing on 13 version here are some results
```sql
test_default= select replace(json_value,'\n','<br>') from ir_default where id=25;
    replace     
----------------
 "test<br>test"
(1 row)

test_default= select replace(json_value,E'\\n','<br>') from ir_default where id=25;
    replace     
----------------
 "test<br>test"
(1 row)

test_default= select json_value from ir_default where id=25;
  json_value  
--------------
 "test\ntest"
(1 row)
```

sagu-odoo: and one more thing should I add test case for ``tab`` also ?.

sagu-odoo:
upgradeci retry with always only crm in all versions

src/base/tests/test_util.py
```diff
@@ -2109,13 +2109,23 @@ def test_convert_field_to_html(self):
         f1 = self.env["ir.model.fields"].create(
             {"name": "x_testx", "model": "res.partner", "ttype": "text", "model_id": model.id, "translate": True}
         )
+        f2 = self.env["ir.model.fields"].create(
+            {"name": "x_testx2", "model": "res.partner", "ttype": "text", "model_id": model.id, "translate": True}
+        )
```
KangOl, line 2114: No need to create a second field.

src/util/pg.py
```diff
@@ -421,17 +421,20 @@ def pg_text2html(s, wrap="p"):
...
+                                                -- double the %% to allow this code chunk to be used in parameterized queries
+                                                'https?://[-A-Za-z0-9+&@#/%%?=~_()|!:,.;]*[-A-Za-z0-9+&@#/%%=~_()|]',
+                                                '<a href="\&" target="_blank" rel="noreferrer noopener">\&</a>',
+                                                'g'),
+                                    E'\\n',
+                                    '<br>'),
+                            E'\\t',
+                            '&Tab;'),
+                        E'\n',
+                        '<br>'),
```
KangOl, line 437: No need to escape `\n` twice.
I think we can just get rid of the `E` prefix in the existing replaces.

KangOl (changes requested):

src/util/pg.py
sagu-odoo: without E prefix it won't work for the simple text field  here are some result 
``` sql

test_default= select replace(description,E'\n','<br>')  from crm_lead where id=45;
     replace      
------------------
 test<br>test<br>
(1 row)

test_default= select replace(description,E'\\n','<br>')  from crm_lead where id=45;
 replace 
---------
 test   +
 test   +
 
(1 row)

test_default= select replace(description,'\\n','<br>')  from crm_lead where id=45;
 replace 
---------
 test   +
 test   +
 
(1 row)

test_default= select replace(description,'\n','<br>')  from crm_lead where id=45;
 replace 
---------
 test   +
 test   +
 
(1 row
```

src/base/tests/test_util.py
sagu-odoo: @KangOl  done

src/util/pg.py
KangOl: Ok, so we just need to use a singe `\`.

aj-fuentes: I think that we need to use `E\n` and `E\t` here. But then in this part
https://github.com/odoo-dev/upgrade-util/blob/d360838e515af2408add216a64f7953c43710cff/src/util/fields.py#L660

We need to do add the `\\n` -> `<br/>` similar for tabs I guess. Because that value is doubly encoded :/

KangOl: It is not doubled encoded. It is just JSON encoded.
We can convert from/to json
```python
        json_value_html = pg_text2html("json_value::jsonb->>0")
        query = """
                UPDATE ir_default AS d
                   SET json_value = to_json({})::text
                  FROM ir_model_fields AS imf
                 WHERE imf.name = %s
                   AND imf.model = %s
                   AND imf.id = d.field_id
                """
```

aj-fuentes: > It is not doubled encoded. It is just JSON encoded.

It's text (that should be then html) inside a json encoding, that's what I meant. Normally `pg_text2html` assumes that the column is text representing as html (not html inside a json text literal).

Anyways I think your solution is basically what needs to be done in this part :)

sagu-odoo: thanks for review. Suggested changes are done

KangOl (approved):
@robodoo r+

PR #274, opened 2025-05-20
[IMP] util.uniq_tags

KangOl:
Rewrite function to use `replace_record_references_batch`


KangOl:
upgradeci retry with always only note crm project in 11.0

src/util/data.py
```diff
@@ -25,77 +25,26 @@ def uniq_tags(cr, model, uniq_column="name", order="id"):
...
-    query = """
-        WITH dups AS (
-            SELECT (array_agg(id order by {order}))[1] as id,
-                   (array_agg(id order by {order}))[2:array_length(array_agg(id), 1)] as others
+
+    query = format_query(
+        cr,
+        """
+            SELECT unnest((array_agg(id order by {order}))[2:]),
+                   (array_agg(id order by {order}))[1]
```
aj-fuentes, line 33: nit
```suggestion
            SELECT unnest((array_agg(id ORDER BY {order}))[2:]),
                   (array_agg(id ORDER BY {order}))[1]
```

aj-fuentes (commented):
This function now is much more general. Maybe we should introduce a new function (`uniquify`?) and make `uniq_tags` reuse it for compatibility. The idea is that we can start suggesting this function in our PRs for simple single-column deduplication.

KangOl:
@robodoo r+

PR #276, opened 2025-05-22
[IMP] get rid of savepoints

KangOl:


src/util/fields.py
```diff
@@ -510,49 +510,60 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
...
+    [fid] = cr.fetchone() or [None]
+
     if fid:
+        # recreate the xmlids
         name = IMD_FIELD_PATTERN % (model.replace(".", "_"), new)
-        # In some cases the same field may be present on ir_model_data with both the double __ and single _ name
-        # version. To avoid conflicts (module, name) on the UPDATE below we keep only the double __ version
+        # duplicate key. May happen du to conflict between
+        # some_model.sub_id and some_model_sub.id
+        # (before saas~11.2, where pattern changed)
```
aj-fuentes: Some typos. I'd make this clear enough to avoid people figuring out what pattern we are talking about.
```suggestion
        # Duplicated key. May happen due to some_model.sub_id and some_model_sub.id
        # both leading to some_model_sub_id xmlid before saas~11.2 where the pattern
        # was fixed to be some_module__sub_id and some_model_sub__id respectively.
```

aj-fuentes (commented):
posting since rebased

src/util/fields.py
```diff
@@ -510,49 +510,69 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
...
-        except psycopg2.IntegrityError:
-            # duplicate key. May happen du to conflict between
-            # some_model.sub_id and some_model_sub.id
-            # (before saas~11.2, where pattern changed)
-            name = "%s_%s" % (name, fid)
-            cr.execute("UPDATE ir_model_data SET name=%s WHERE model='ir.model.fields' AND res_id=%s", [name, fid])
+        if cr.rowcount:
+            modules = {mod for (mod,) in cr.fetchall()}
+
+            inserted = psycopg2.extras.execute_values(
```
aj-fuentes, line 546: Was about to suggest this. :)

src/util/fields.py
```diff
@@ -510,49 +510,69 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
...
+                # some_model.sub_id and some_model_sub.id
+                # (before saas~11.2, where pattern changed)
+                on_conflict_name = "{}_{}".format(name, fid)
+                psycopg2.extras.execute_values(
+                    cr,
+                    """
+                    INSERT INTO ir_model_data(module, name, model, res_id, noupdate)
+                         VALUES %s
+                    """,
+                    [(module, on_conflict_name, 'ir.model.fields', fid, True) for module in conflicted],
```
aj-fuentes, line 569: Better?
```suggestion
                cr.execute(
                    """
                    INSERT INTO ir_model_data(module, name, model, res_id, noupdate)
                         SELECT UNNEST(%s), %s, 'ir.model.fields', %s, True
                    """,
                    [sorted(conflicted), on_conflict_name, fid],
```

src/util/fields.py
```diff
@@ -510,49 +510,69 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
...
+
+            inserted = psycopg2.extras.execute_values(
+                cr,
+                """
+                INSERT INTO ir_model_data(module, name, model, res_id, noupdate)
+                     VALUES %s
+                ON CONFLICT DO NOTHING
+                  RETURNING module
+                """,
+                [(module, name, 'ir.model.fields', fid, True) for module in modules],
```
aj-fuentes, line 554: Since we only vary `module` maybe an `unnest(%s)` with a list of modules is enough instead of execute_many

src/util/fields.py
```diff
@@ -511,49 +510,69 @@ def rename_field(cr, model, old, new, update_references=True, domain_adapter=Non
...
+            conflicted = modules - {mod for (mod,) in inserted}
+            if conflicted:
+                # duplicate key. May happen du to conflict between
+                # some_model.sub_id and some_model_sub.id
+                # (before saas~11.2, where pattern changed)
+                on_conflict_name = "{}_{}".format(name, fid)
+                psycopg2.extras.execute_values(
+                    cr,
+                    """
+                    INSERT INTO ir_model_data(module, name, model, res_id, noupdate)
```
aj-fuentes, line 566: Why do we insert conflicted modules? Shouldn't we update instead as we did before?

src/util/fields.py
KangOl: I like it!
And it resolves the old-psycop2-version-that-dont-have-the-fetch-argument issue[^1]

[^1]: https://upgradeci.odoo.com/upgradeci/run/205248

src/util/fields.py
KangOl: I now remove all the existing xmlids. Before we were keeping one that we updated.

aj-fuentes: Indeed. Got it. Looks good then.

aj-fuentes (approved):
LGTM

KangOl:
@robodoo rebase-ff r+

KangOl:
@robodoo r+

PR #279, opened 2025-05-27
[IMP] pg: tiny simplification

cawo-odoo:
When using a `Cursor` in a `with` statement, it will be committed automatically. No need to do it explicitly and no need to save the rowcount in a local var.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -111,9 +111,7 @@ if ThreadPoolExecutor is not None:
         def execute(query):
             with cursor() as tcr:
                 tcr.execute(query)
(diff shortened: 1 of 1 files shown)


aj-fuentes (commented):
Don't make this type of PRs, unless they really add some measurable value (performance, avoid errors, etc). The codebase is full of such "tiny" improvements. If you need to touch this code then, yes, include such small details. But don't waste time in such simple things otherwise.


That being said, this LGTM.

KangOl (approved):
@robodoo r+

PR #280, opened 2025-05-27
[IMP] tools/compile23.py

cawo-odoo:
Recent systems don't have python2 installed.
Let the pre-commit hook pass, but warns, when it must verify scripts that should be compatible with python2.


aj-fuentes (approved):

KangOl (approved):
@robodoo r+

PR #283, opened 2025-06-06
[FIX] util/fields: rename index

jjmaksoud:
The index rename query had the index names switched :(


jjmaksoud:
upgradeci retry with only base

aj-fuentes (approved):
🤦🏼

PR #284, opened 2025-06-11
[FIX] spreadsheet: batch process `spreadsheet_revision.commands`

vval-odoo:
[FIX] spreadsheet: batch process `spreadsheet_revision.commands`

Some dbs have `spreadsheet_revision` records with over 10 millions characters in `commands`. If the number of record is high, this leads to memory errors [here](https://github.com/odoo/upgrade-util/blob/86409e52629c0d77c8ff83871fee45078fd9960e/src/util/spreadsheet/misc.py#L7). We use a named cursor so the fetching is done in postgres and data are processed in smaller batches by the upgrade process.
```
select count(*),SUM(CHAR_LENGTH(commands)) as characters from spreadsheet_revision where commands like any (array['%"DELETE_SHEET"%', '%"MOVE_COLUMNS_ROWS"%', '%"REMOVE_COLUMNS_ROWS"%', '%"ADD_COLUMNS_ROWS"%', '%"RENAME_SHEET"%'])
+-------+------------+
| count | characters |
|-------+------------|
| 2317  | 3419017646 |
+-------+------------+
```
```
select id,CHAR_LENGTH(commands) from spreadsheet_revision where commands like any (array['%"DELETE_SHEET"%', '%"MOVE_COLUMNS_ROWS"%', '%"REMOVE_COLUMNS_ROWS"%', '%"ADD_COLUMNS_ROWS"%', '%"RENAME_SHEET"%']) order by CHAR_LENGTH(commands) desc limit 10
+------+-------------+
| id   | char_length |
|------+-------------|
| 3871 | 13197157    |
| 4788 | 13197157    |
| 3290 | 13197157    |
| 3557 | 13197157    |
| 4179 | 13197157    |
| 4481 | 13197157    |
| 2757 | 13197157    |
| 3022 | 13197157    |
| 2492 | 13197157    |
| 5097 | 13197157    |
+------+-------------+
```

Fixes [upg-2899961](https://upgrade.odoo.com/odoo/action-150/2899961)


src/util/spreadsheet/misc.py
```diff
@@ -1,10 +1,13 @@
-from .. import json
+from .. import json, pg
+
+BATCH_SIZE = 10
```
Pirols:
```suggestion
BATCH_SIZE = 10
```

Have you tried larger values too? Is there any impact on the performance?

src/util/spreadsheet/misc.py
```diff
@@ -1,10 +1,13 @@
...
 
 
 def iter_commands(cr, like_all=(), like_any=()):
     if not (bool(like_all) ^ bool(like_any)):
         raise ValueError("Please specify `like_all` or `like_any`, not both")
-    cr.execute(
+    ncr = pg.named_cursor(cr, itersize=BATCH_SIZE)
```
Pirols: Using a context manager you do not need to close it explicitely[^1].

```suggestion
    with pg.named_cursor(cr, itersize=BATCH_SIZE) as ncr:
```

That said, this is just in the name of a more pythonic implementation. IOW: imo you can keep your current version, if you like it better.

[^1]: https://github.com/odoo/odoo//blob/bdc28e53b7c426640e30b900064442f19c946d9e/odoo/sql_db.py#L221-L238

Pirols (commented):
Good work! :)

src/util/spreadsheet/misc.py
vval-odoo: I did, it takes roughly the same time. Since some records holds up to ~13Mo, I thought a 100 records could already be too much.

Pirols: In fact I was not sure how much we could gain by copying those tuples into memory in batches. Answer: whatever it is, it's probably dwarfed by the dominant factor (which is the processing of those commands). If that's confirmed to be the case, we might as well remove any batching (batch_size = 1); which would enable us to process any command that can fit in the available memory.

cawo-odoo: > In fact I was not sure how much we could gain by copying those tuples into memory in batches

It's less about memory copy, it's about network round trips. Would we expect other DBs to potentially have a lot of small tuples instead of a few big ones in this table? If so, then it may be worth it to be clever about it, like selecting the max-sized commands and from that determine an itersize.

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,33 @@
...
-        if "commands" not in data_loaded:
-            continue
-        data_old = json.dumps(data_loaded, sort_keys=True)
+    with pg.named_cursor(cr, itersize=1) as ncr:
+        ncr.execute(
+            """
+            SELECT id,
+                commands
+            FROM spreadsheet_revision
+            WHERE commands LIKE {}(%s::text[])
```
aj-fuentes: Keep the formatting.
```suggestion
            SELECT id,
                   commands
              FROM spreadsheet_revision
             WHERE commands LIKE {}(%s::text[])
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,33 @@
...
-         WHERE commands LIKE {}(%s::text[])
-        """.format("ALL" if like_all else "ANY"),
-        [list(like_all or like_any)],
-    )
-    for revision_id, data in cr.fetchall():
-        data_loaded = json.loads(data)
-        if "commands" not in data_loaded:
-            continue
-        data_old = json.dumps(data_loaded, sort_keys=True)
+    with pg.named_cursor(cr, itersize=1) as ncr:
```
aj-fuentes, line 10: You can either leave the default `itersize`, or optimize the value to something that would work depending on the data. Another alternative is to use `fetchmany` directly.
```suggestion
    with pg.named_cursor(cr) as ncr:
```

src/util/spreadsheet/misc.py
vval-odoo: Something like this?
` itersize = int(available_memory / biggest_command)`
How can I determine `available_memory` ?

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,89 @@
...
+                        ELSE a.sum + o.length
+                   END AS sum
+              FROM assign a
+              JOIN ordered_rows o
+                ON o.rownum = a.rownum + 1
+        )
+        SELECT count(*),ARRAY_AGG(id)
+          FROM assign
+      GROUP BY bucket
+      ORDER BY bucket;
```
aj-fuentes: I think this can be simplified. You only need for each `id` the number of the bucket where it would be. In case one length is longer than one bucket you want to keep the record alone in it's bucket to avoid some smaller records making the group take even more space.
You can return the ids and commands already per bucket. That way you can iterate over the named cursor one by one.

Example query: (/!\warning: untested/!\)
```sql
  buckets AS (
      SELECT id,
             mod(SUM(length(commands) OVER (ORDER BY id), %(bucket_size)s) AS num,
             length(commands) > %(bucket_size)s AS alone
        FROM spreadsheet_revision
       WHERE commands LIKE {}(%(filters)s::text[])
    ORDER BY id
  )
  SELECT ARRAY_AGG(id ORDER BY id)
         ARRAY_AGG(commands ORDER BY id)
    FROM buckets
GROUP BY num, alone
```
then
```py
for ids, datas in ncr:  # you can iterate one by one here
    for id, data in zip(ids, datas):
        ...
```

Pirols:
Another affected request: https://upgrade.odoo.com/odoo/action-150/2988031

src/util/spreadsheet/misc.py
vval-odoo: I believe what we need `AS num` is:
```
SUM(LENGTH(commands)) OVER (ORDER BY id) / {memory_cap}
```
And, when trying to iterate directly through ncr, I still get the memory error... So I used fetchmany(size=1) instead.
Except for that, I implemented your solution, and it is indeed way simpler and as effecient.

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,46 @@
...
+              ORDER BY id
             )
+            SELECT ARRAY_AGG(id ORDER BY id),
+                   ARRAY_AGG(commands ORDER BY id)
+              FROM buckets
+          GROUP BY num
+            """.format(memory_cap=MEMORY_CAP, condition="ALL" if like_all else "ANY"),
+            [list(like_all or like_any)],
+        )
+        for ids, datas in ncr.fetchmany(size=1):
```
Pirols: This will only output the first bucket.

A simple solution is what you had already:

```suggestion
        for ids, datas in ncr:
```

With `with pg.named_cursor(cr, itersize=1) as ncr`

src/util/spreadsheet/misc.py
aj-fuentes: My bad, indeed. I meant just division :)

src/util/spreadsheet/misc.py
aj-fuentes: I agree with Edoardo. You just need to limit the itersize. Under the hood psycopg will fetch one by one and each line already contains multiple records (as many as they fit in one bucket).

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,46 @@
...
-
-        changed = yield data_loaded["commands"]
-        if changed is None:
-            changed = data_old != json.dumps(data_loaded, sort_keys=True)
-
-        if changed:
-            cr.execute(
-                "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s", [json.dumps(data_loaded), revision_id]
+
+    with pg.named_cursor(cr) as ncr:
```
aj-fuentes:
```suggestion
    with pg.named_cursor(cr, itersize=1) as ncr:
```
https://www.psycopg.org/docs/usage.html#server-side-cursors

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,46 @@
...
+                       SUM(LENGTH(commands)) OVER (ORDER BY id) / {memory_cap} AS num,
+                       commands
+                  FROM spreadsheet_revision
+                 WHERE commands LIKE {condition}(%s::text[])
+              ORDER BY id
             )
+            SELECT ARRAY_AGG(id ORDER BY id),
+                   ARRAY_AGG(commands ORDER BY id)
+              FROM buckets
+          GROUP BY num
```
aj-fuentes: This is fine but it has the downside that if you have a very large record you will get it along with all other smaller records that fit in its bucket.
Example
```
|--butket1--|--butket1--|--butket2--|
***big*record*************
                          *small1
                           *small2
                            ...
                                  *smallN
```
Total size in group associated to bucket 3: 3*bucket_size, whether we could have gotten _just_ big_record's size if we left out the smaller ones. It would depend on whether we actually get cases where there is a HUGE record that is still OK alone but too much with the extra ~200MB.

All this to say that's why I added the `len > memory_cap AS alone` then use it in the groping ;)

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,47 @@
...
+                       commands
+                  FROM spreadsheet_revision
+                 WHERE commands LIKE {condition}(%s::text[])
+              ORDER BY id
             )
+            SELECT ARRAY_AGG(id ORDER BY id),
+                   ARRAY_AGG(commands ORDER BY id)
+              FROM buckets
+          GROUP BY num, alone
+            """.format(memory_cap=MEMORY_CAP, condition="ALL" if like_all else "ANY"),
```
aj-fuentes: Use `format_query` to set `ALL` or `ANY`. Then pass memory_cap as parameter to the query in `ncr.execute`
EDIT: `condition = util.SQLStr("ALL" if like_all else "ANY")`

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,47 @@
...
-        if changed:
-            cr.execute(
-                "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s", [json.dumps(data_loaded), revision_id]
+
+    with pg.named_cursor(cr, itersize=1) as ncr:
+        ncr.execute(
+            """
+            WITH buckets AS (
+                SELECT id,
+                       SUM(LENGTH(commands)) OVER (ORDER BY LENGTH(commands),id) / {memory_cap} AS num,
```
aj-fuentes:
```suggestion
                       SUM(LENGTH(commands)) OVER (ORDER BY LENGTH(commands),id) / %s AS num,
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,47 @@
...
+                       LENGTH(commands) > {memory_cap} AS alone,
```
aj-fuentes:
```suggestion
                       LENGTH(commands) > %s AS alone,
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,47 @@
-from .. import json
+from .. import json, pg
+
+MEMORY_CAP = 2 * 10**8  # 200Mo
```
cawo-odoo: moooo :cow2: 
```suggestion
MEMORY_CAP = 2 * 10**8  # 200MB
```

vval-odoo: :fr: :-1:  :us: :+1:

vval-odoo:
> Note that this could still lead to a big record being grouped with many more "smaller" records potentially adding up to 200mb to the size of the fetched data --vs what would be fetched if the record was alone. If we don't avoid that case I don't see any advantage in ordering by length.

That was the idea behind my original query. Compute buckets that would only go above `memory_cap` in case of lone record bigger than `memory_cap`. But for this we need to know for each row what will be the size of following one, so it complicates the logic a bit. Should I go with this one then?

```
        ncr.execute(
            """
            WITH RECURSIVE start_bucket AS (
                SELECT 1 AS bucket
            ),
            ordered_rows AS (
                SELECT id,
                       commands,
                       LENGTH(commands) AS length,
                       ROW_NUMBER() OVER (ORDER BY LENGTH(commands), id) AS rownum
                  FROM spreadsheet_revision
                 WHERE commands LIKE {condition}(%s::text[])
            ),
            assign AS (
                SELECT o.id AS id,
                       o.commands as commands,
                       o.length,
                       o.rownum,
                       sb.bucket AS bucket,
                       o.length AS sum
                  FROM ordered_rows o, start_bucket sb
                 WHERE o.rownum = 1

                UNION ALL

                SELECT o.id AS id,
                       o.commands as commands,
                       o.length,
                       o.rownum,
                       CASE
                            WHEN a.sum + o.length > %s THEN a.bucket + 1
                            ELSE a.bucket
                       END AS bucket,
                       CASE
                            WHEN a.sum + o.length > %s THEN o.length
                            ELSE a.sum + o.length
                       END AS sum
                  FROM assign a
                  JOIN ordered_rows o
                    ON o.rownum = a.rownum + 1
            )
            SELECT ARRAY_AGG(id),
                   ARRAY_AGG(commands)
              FROM assign
          GROUP BY bucket
            """.format(condition=pg.SQLStr("ALL" if like_all else "ANY")),
            [list(like_all or like_any), MEMORY_CAP, MEMORY_CAP]
        )
```

aj-fuentes:
> That was the idea behind my original query. Compute buckets that would only go above `memory_cap` in case of lone record bigger than `memory_cap`. But for this we need to know for each row what will be the size of following one, so it complicates the logic a bit. Should I go with this one then?

What's wrong with using the "alone" trick I sent you? If a record is bigger than the bucket it will get `alone=True`, now it's impossible for two records bigger than a bucket to get `alone=True` in the **same** bucket. Therefore if we group by `num, alone` we achieve our goal: big records won't be bucketed with smaller ones. You can think of it as having two groupings one for big records that will be fetched individually, and another for smaller records that would be clustered by bucket.

KangOl:
No need for `lag`/`lead`, we can just use a cumulative sum to get the ids to fetch in successive queries
```sql
  with _groups as (
    select id, sum(length(commands)) over (ORDER BY id) / {mem_cap} as cs
      from spreadsheet_revision
     where {condition}
  )
  select array_agg(id)
    from _groups
group by cs
```

Pirols:
> What's wrong with using the "alone" trick I sent you? If a record is bigger than the bucket it will get `alone=True`, now it's impossible for two records bigger than a bucket to get `alone=True` in the **same** bucket.

I don't think that the `alone` trick is enough, it still does not reason about wasted bucket space. Imagine the following scenario: bucket size is 10 chars, the first command (c1) is 6 chars long, second one (c2) is 6 chars long, third one (c3) is 5 chars long. No one qualifies as `alone`. First command goes into b0 (first bucket), second command goes into b1 (because 12 / 10 > 1), third command also goes into b1 (17/10 < 2), however `len(c2) + len(c3) = 11 > bucket_size = 10`:

```
|----b0----|----b1----|----b2----|
 ******c1
            ******c2
                  *****c3
```

vval-odoo:
> What's wrong with using the "alone" trick I sent you? If a record is bigger than the bucket it will get `alone=True`, now it's impossible for two records bigger than a bucket to get `alone=True` in the **same** bucket. Therefore if we group by `num, alone` we achieve our goal: big records won't be bucketed with smaller ones. You can think of it as having two groupings one for big records that will be fetched individually, and another for smaller records that would be clustered by bucket.
I think that two records bigger than `memory_cap` can't get in the same bucket, even without the "alone" trick. What this tricks prevents is for smaller records to "complete" a bucket that is already too large, but that doesn't solve any problem because a bucket of `15,something * memory_cap` would make the program crash as would a bucket of `15 * memory_cap`. What I try to prevent is getting buckets between `memory_cap` and `2 * memory_cap`.
I'll try what you proposed in your 'edit', but we have to agree that the real memory cap is `2 * memory_cap`, right?

aj-fuentes:
> > What's wrong with using the "alone" trick I sent you? If a record is bigger than the bucket it will get `alone=True`, now it's impossible for two records bigger than a bucket to get `alone=True` in the **same** bucket.
> 
> I don't think that the `alone` trick is enough, it still does not reason about wasted bucket space. Imagine the following scenario: bucket size is 10 chars, the first command (c1) is 6 chars long, second one (c2) is 6 chars long, third one (c3) is 5 chars long. No one qualifies as `alone`. First command goes into b0 (first bucket), second command goes into b1 (because 12 / 10 > 1), third command also goes into b1 (17/10 < 2), however `len(c2) + len(c3) = 11 > bucket_size = 10`:
> 
> ```
> |----b0----|----b1----|----b2----|
>  **c1**
>             **c2**
>                   *c3**
> ```

I think it is correct. See we don't care about 2*max_cap size since we **know** that's the hard max. We care about a really big record, think about 5*max_cap, getting even _more_ smaller ones. In your example, yes c2+c3 > max_cap, but still < 2*max_cap. That's my point. The logic is simpler and as proven by the example here the biggest record is bigger than 2*max_cap and still acceptable for memory.

src/util/spreadsheet/misc.py
Pirols:
```suggestion
                       SUM(LENGTH(commands)) OVER (ORDER BY LENGTH(commands)) / {memory_cap} AS num,
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,47 @@
...
+        ncr.execute(
+            """
+            WITH buckets AS (
+                SELECT id,
+                       SUM(LENGTH(commands)) OVER (ORDER BY LENGTH(commands),id) / {memory_cap} AS num,
+                       LENGTH(commands) > {memory_cap} AS alone,
+                       commands
+                  FROM spreadsheet_revision
+                 WHERE commands LIKE {condition}(%s::text[])
+              ORDER BY id
```
Pirols:
```suggestion
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,47 @@
...
             )
+            SELECT ARRAY_AGG(id ORDER BY id),
+                   ARRAY_AGG(commands ORDER BY id)
```
Pirols:
```suggestion
            SELECT ARRAY_AGG(id),
                   ARRAY_AGG(commands)
```

Pirols (commented):
Remove seemingly useless ordering?

src/util/spreadsheet/misc.py
aj-fuentes: The order is necessary. Example just to illustrate that we don't really want the sum to go over a random order.
```
=> select id,sum(id) over(order by id) from res_users order by id
+----+-----+
| id | sum |
|----+-----|
| 1  | 1   |
| 2  | 3   |
| 3  | 6   |
| 4  | 10  |
+----+-----+
=> select id,sum(id) over(order by id desc) from res_users order by id
+----+-----+
| id | sum |
|----+-----|
| 1  | 10  |
| 2  | 9   |
| 3  | 7   |
| 4  | 4   |
+----+-----+
```

src/util/spreadsheet/misc.py
aj-fuentes: The order is necessary here as well otherwise you get a wrong match when we do the `zip(ids, commands)` in python side

src/util/spreadsheet/misc.py
aj-fuentes: We can remove the order but we gain little. OTOH is very nice to be sure we get a deterministic output each time we run the whole process.

KangOl:
upgradeci retry

KangOl:
upgradeci retry

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,58 @@
...
+              FROM smaller
+          GROUP BY num
+
+             UNION ALL
+
+            SELECT ARRAY[id],
+                   ARRAY[commands]
+              FROM filtered
+             WHERE commands_length > %s
+            """.format(condition=pg.SQLStr("ALL" if like_all else "ANY")),
```
aj-fuentes:
```suggestion
            """,
            condition=pg.SQLStr("ALL" if like_all else "ANY"),
            )  # close format_query
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,58 @@
...
-        if changed is None:
-            changed = data_old != json.dumps(data_loaded, sort_keys=True)
-
-        if changed:
-            cr.execute(
-                "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s", [json.dumps(data_loaded), revision_id]
+
+    with pg.named_cursor(cr, itersize=1) as ncr:
+        ncr.execute(
+            """
```
aj-fuentes: Use `format_query`. The query is becoming increasingly complex, better use the right formatting tool to avoid issues later.
```suggestion
            util.format_query(
                cr,
                """
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,62 @@
...
+                    SELECT id,
+                           commands,
+                           sum(commands_length) OVER (ORDER BY id) / %s AS num
+                      FROM filtered
+                     WHERE commands_length <= %s
+                )
+                SELECT array_agg(id ORDER BY id),
+                       array_agg(commands ORDER BY id)
+                  FROM smaller
+              GROUP BY num
```
aj-fuentes, line 32:
```suggestion
              GROUP BY num
              ORDER BY num
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,62 @@
...
+
+                 UNION ALL
+
+                SELECT ARRAY[id],
+                       ARRAY[commands]
+                  FROM filtered
+                 WHERE commands_length > %s
```
aj-fuentes, line 40:
```suggestion
                 WHERE commands_length > %s
              ORDER BY id
```

aj-fuentes (approved):
LGTM

I left an optional suggestion to ensure we return data deterministically.

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,65 @@
...
+              GROUP BY num
+
+                 UNION ALL
+
+                SELECT ARRAY[id],
+                       ARRAY[commands],
+                       NULL AS num
+                  FROM filtered
+                 WHERE commands_length > %s
+              ORDER BY num NULLS LAST
```
aj-fuentes: This is not what we need. It isn't actually ordering the result since all are NULL. You are doing an UNION ALL thus you should order on each query. In the first you order by `num` here you then need to order by `id`. You don't need to return `num` since it isn't used in Python side.

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,65 @@
...
+                           commands,
+                           sum(commands_length) OVER (ORDER BY id) / %s AS num
+                      FROM filtered
+                     WHERE commands_length <= %s
+                )
+                SELECT array_agg(id ORDER BY id),
+                       array_agg(commands ORDER BY id),
+                       num
+                  FROM smaller
+              GROUP BY num
```
aj-fuentes, line 32:
```suggestion
              GROUP BY num
              ORDER BY num
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,65 @@
...
+
+                 UNION ALL
+
+                SELECT ARRAY[id],
+                       ARRAY[commands],
+                       NULL AS num
```
aj-fuentes:
```suggestion
                       ARRAY[commands]
```

src/util/spreadsheet/misc.py
aj-fuentes:
```suggestion
              ORDER BY id
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,65 @@
...
+                  FROM filtered
+                 WHERE commands_length > %s
+              ORDER BY num NULLS LAST
+                """,
+                condition=pg.SQLStr("ALL" if like_all else "ANY"),
+            ),
+            [list(like_any or like_all), MEMORY_CAP, MEMORY_CAP, MEMORY_CAP],
+        )
+        for ids, commands, _ in ncr:
```
aj-fuentes, line 47:
```suggestion
        for ids, commands in ncr:
```

src/util/spreadsheet/misc.py
```diff
@@ -1,32 +1,65 @@
...
+                ), smaller AS (
+                    SELECT id,
+                           commands,
+                           sum(commands_length) OVER (ORDER BY id) / %s AS num
+                      FROM filtered
+                     WHERE commands_length <= %s
+                )
+                SELECT array_agg(id ORDER BY id),
+                       array_agg(commands ORDER BY id),
+                       num
```
aj-fuentes:
```suggestion
                       array_agg(commands ORDER BY id)
```

aj-fuentes (commented):
Ensure each query in the UNION ALL is ordered independently. For the first one the order that makes sense is `num`, for the second one is `id`.

src/util/spreadsheet/misc.py
vval-odoo: I get the problem. However, independently ordering sets of a union is not authorized by psql. I can either do that with subqueries, either use something more hacky like this:
```
                SELECT array_agg(id ORDER BY id),
                       array_agg(commands ORDER BY id),
                       min(id) AS sort_key
                  FROM smaller
              GROUP BY num

                 UNION ALL

                SELECT ARRAY[id],
                       ARRAY[commands],
                       id AS sort_key
                  FROM filtered
                 WHERE commands_length > %s
              ORDER BY sort_key
```

aj-fuentes: The sort_key idea looks good to me :)

aj-fuentes (approved):

rish-odoo:
Hello @aj-fuentes
Is there any chance of getting r+ on this PR?

PR #285, opened 2025-06-17
[IMP] util/pg: cache column_exists for common columns

aj-fuentes:
In many cases we know that columns must exist, it's better to cache them
to avoid thousands of small queries.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -43,7 +43,7 @@ except ImportError:
 
 from .exceptions import MigrationError, SleepyDeveloperError
 from .helpers import _validate_table, model_of_table
-from .misc import Sentinel, log_progress
+from .misc import Sentinel, log_progress, version_gte
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


src/util/pg.py
```diff
@@ -472,6 +473,30 @@ def _column_info(cr, table, column):
...
+            if not kwargs and tuple(args) in cache:
+                return cache[tuple(args)]
+            return f(cr, *args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+@_forced_cache(_COLUMNS)
```
KangOl: I don't think the decorator is needed. It can be inline into the function directly.
```python
def column_exists(cr, table, column):
    if (table, column) in _COLUMNS:
        return _COLUMNS[(table, column)]
    return _column_info(cr, table, column) is not None
```

KangOl:
```
upgradeci retry with always only hr* *_hr* in all versions
```

src/util/pg.py
aj-fuentes: Well, originally I was using it in `get_value_or_en_translation`. It proved not as useful. I'm still working on this, I will simplify as much as possible before the final version. There are still some other columns being checked more than 70 times in the local test I'm doing.

PR #287, opened 2025-06-19
[FIX] util/records:Mark records noupdate

sagu-odoo:
If particular xml don't have noupdate marked but parent ``data`` node have noupdate in that case from update_records_from_xml creating record as noupdate ``false`` which is causing issue in future upgrade because it going for delete. For preventing such case adding noupdate flag to record and inherit created record acccoring to this [commit](
https://github.com/odoo/odoo/commit/ef709fd92b7b22a58e60715c9ff7ab4476a83ca5#diff-7144f88ea32f36feb17ce1b8dda7dee1631f5ada34075414587df3948c6b3d1bL4098-R4118) by that ``//data`` node or that ``xml node``  itself.

use case:
client db don't have the [``documents.document_internal_folder``](https://github.com/odoo/enterprise/blob/6c677f388c934e3fc545fc7cb2f9d65a958433da/documents/data/documents_document_data.xml#L2-L18) xmlid which is created when migrated to 18.0 from this [script](
https://github.com/odoo/upgrade/blob/9327a2c12d9faf67d71f09bbe4d91e6105c75dca/migrations/documents/saas~17.5.1.4/post-migrate.py#L24) with noupdate false and because data node have noupdate true but particular xml don't have noupdate that is why it created as noupdate false and other inherit records also with noupdate and after migrating from onwards version related to that mail.alias records xml creating according to [this](
https://github.com/odoo/odoo/blob/9822f5cd1fd3ac60689fc5e3101e5f6fa6a5722d/odoo/models.py#L5514)

and later on version in 18.3 if related inherit model is changed but still some how use case of those records for preventing in that case.

before upgrade
```
timer_test_document_test=# select * from ir_model_data where name like '%document_internal_folder%'
;
 id | create_uid | create_date | write_date | write_uid | res_id | noupdate | name | module | model
----+------------+-------------+------------+-----------+--------+----------+------+--------+-------
(0 rows)
```

before fix
```
timer_test_document_test=# select * from ir_model_data where name like '%document_internal_folder%'
;
  id   | create_uid |        create_date         |         write_date         | write_uid | res_id | noupdate |                name                 |  module   |       model
-------+------------+----------------------------+----------------------------+-----------+--------+----------+-------------------------------------+-----------+--------------------
 39576 |            | 2025-06-19 13:38:37.987267 | 2025-06-19 13:38:37.987267 |           |     20 | f        | document_internal_folder_mail_alias | documents | mail.alias
 39577 |            | 2025-06-19 13:38:37.987267 | 2025-06-19 13:38:37.987267 |           |     20 | f        | document_internal_folder            | documents | documents.document
```

after fix
```
timer_test_document_test=# select * from ir_model_data where name like '%document_internal_folder%'
;
  id   | create_uid |        create_date         |         write_date         | write_uid | res_id | noupdate |                name                 |  module   |       model
-------+------------+----------------------------+----------------------------+-----------+--------+----------+-------------------------------------+-----------+--------------------
 39576 |            | 2025-06-19 13:41:40.343454 | 2025-06-19 13:41:40.343454 |           |     20 | t        | document_internal_folder_mail_alias | documents | mail.alias
 39577 |            | 2025-06-19 13:41:40.343454 | 2025-06-19 13:41:40.343454 |           |     20 | t        | document_internal_folder            | documents | documents.document

```

```
  File "/home/odoo/src/odoo/saas-18.3/odoo/service/server.py", line 1396, in preload_registries
    registry = Registry.new(dbname, update_module=update_module, install_modules=config['init'], upgrade_modules=config['update'])
  File "<decorator-gen-6>", line 2, in new
  File "/home/odoo/src/odoo/saas-18.3/odoo/tools/func.py", line 83, in locked
  File "/home/odoo/src/odoo/saas-18.3/odoo/orm/registry.py", line 167, in new
    load_modules(
  File "/home/odoo/src/odoo/saas-18.3/odoo/modules/loading.py", line 509, in load_modules
    env['ir.model.data']._process_end(registry.updated_modules)
  File "/tmp/tmp2ipekce9/migrations/base/0.0.0/pre-models-no-model-data-delete.py", line 108, in _process_end
    return super(IrModelData, self)._process_end(modules)
  File "/home/odoo/src/odoo/saas-18.3/odoo/addons/base/models/ir_model.py", line 2589, in _process_end
    self._process_end_unlink_record(record)
  File "/home/odoo/src/odoo/saas-18.3/addons/website/models/ir_model_data.py", line 35, in _process_end_unlink_record
    return super()._process_end_unlink_record(record)
  File "/home/odoo/src/odoo/saas-18.3/odoo/addons/base/models/ir_model.py", line 2518, in _process_end_unlink_record
    record.unlink()
  File "/home/odoo/src/odoo/saas-18.3/odoo/orm/models.py", line 3913, in unlink
    cr.execute(SQL(
  File "/home/odoo/src/odoo/saas-18.3/odoo/sql_db.py", line 422, in execute
psycopg2.errors.ForeignKeyViolation: update or delete on table "mail_alias" violates foreign key constraint "documents_document_alias_id_fkey" on table "documents_document"
DETAIL:  Key (id)=(40) is still referenced from table "documents_document".
```

upg-2987024
opw-4874805

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1107,7 +1107,8 @@ def __update_record_from_xml(
         return
     else:
         # The xmlid doesn't already exists, nothing to reset
-        reset_write_metadata = noupdate = reset_translations = False
+        reset_write_metadata = reset_translations = False
+        noupdate = True
         fields = None
 
     write_data = None
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -901,11 +901,19 @@ def force_noupdate(cr, xmlid, noupdate=True, warn=False):
              WHERE module = %s
                AND name = %s
                AND noupdate != %s
+         RETURNING model
     """,
         [noupdate, module, name, noupdate],
     )
-    if noupdate is False and cr.rowcount and warn:
-        _logger.warning("Customizations on `%s` might be lost!", xmlid)
+    if cr.rowcount:
+        if noupdate is False and warn:
+            _logger.warning("Customizations on `%s` might be lost!", xmlid)
+
+        [model] = cr.fetchone()
+        for parent_model, _ in inherits_parents(cr, model):
+            parent = parent_model.replace(".", "_")
+            force_noupdate(cr, "{}_{}".format(xmlid, parent), noupdate=noupdate, warn=warn)
+
     return cr.rowcount
 
 
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -33,7 +33,7 @@ from .helpers import (
 )
 from .inconsistencies import break_recursive_loops
 from .indirect_references import indirect_references
-from .inherit import direct_inherit_parents, for_each_inherit
+from .inherit import for_each_inherit, inherits_parents
 from .misc import Sentinel, chunks, parse_version, version_gte
 from .orm import env, flush
 from .pg import (
@@ -841,16 +841,15 @@ def rename_xmlid(cr, old, new, noupdate=None, on_collision="fail"):
             """
             cr.execute(query, {"old": r"\y{}\y".format(re.escape(old)), "new": new})
 
-        for parent_model, inh in direct_inherit_parents(cr, model):
-            if inh.via:
-                parent = parent_model.replace(".", "_")
-                rename_xmlid(
-                    cr,
-                    "{}_{}".format(old, parent),
-                    "{}_{}".format(new, parent),
-                    noupdate=noupdate,
-                    on_collision=on_collision,
-                )
+        for parent_model, _ in inherits_parents(cr, model):
+            parent = parent_model.replace(".", "_")
+            rename_xmlid(
+                cr,
+                "{}_{}".format(old, parent),
+                "{}_{}".format(new, parent),
+                noupdate=noupdate,
+                on_collision=on_collision,
+            )
         return new_id
     return None
 
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1096,13 +1096,10 @@ def __update_record_from_xml(
 
     cr.execute(
         """
-        UPDATE ir_model_data d
-           SET noupdate = false
-          FROM ir_model_data o
-         WHERE o.id = d.id
-           AND d.module = %s
-           AND d.name = %s
-     RETURNING d.model, d.res_id, o.noupdate
+        SELECT model, res_id, noupdate
+          FROM ir_model_data
+         WHERE module = %s
+           AND name = %s
     """,
         [module, name],
     )
@@ -1110,6 +1107,7 @@ def __update_record_from_xml(
         model, res_id, noupdate = cr.fetchone()
         if model == "ir.model":
             return
+        force_noupdate(cr, xmlid, noupdate=False)
     elif not force_create:
         _logger.warning("Record %r not found in database. Skip update.", xmlid)
         return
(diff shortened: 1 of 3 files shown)


aj-fuentes:
I'm not sure we want this. We could call `force_noupdate` in the caller [script](https://github.com/odoo/upgrade/blob/9327a2c12d9faf67d71f09bbe4d91e6105c75dca/migrations/documents/saas~17.5.1.4/post-migrate.py#L24). Or _perhaps_ add a `noupdate=(True/False/None)` (`None` -> keep whatever is already there) to `update_record_from_xml`.

@KangOl wdyt?

sagu-odoo:
Hello @aj-fuentes , @KangOl  Thanks for your quick review. I have update patch so it will work only whenever new record will be  created. your insight will be valuable and it will prevent blocking for future case also That's the main purpose of this fix

src/util/records.py
```diff
@@ -1153,6 +1159,12 @@ def add_ref(ref):
                             node.remove(fn)
                 new_root[0].append(node)
 
+                data_node = doc.find(".//data")
+                if (data_node is not None and eval(data_node.attrib.get("noupdate", "0"))) or (
+                    eval(node.attrib.get("noupdate", "0"))
+                ):
+                    force_noupdate_to_record = True
+
```
aj-fuentes: The idea is to NOT do this.
```suggestion
```

sagu-odoo: but if record is noupdate ``False`` which is created  in that case ?.  :(

src/util/records.py
```diff
@@ -1210,6 +1223,17 @@ def add_ref(ref):
     if reset_write_metadata and write_data:
         cr.execute("UPDATE {} SET write_uid=%s, write_date=%s WHERE id=%s".format(table), write_data)
 
+    # force noupdate newly created record and inherit records
```
aj-fuentes: I'm not sure what you meant here. But I think what you could do is to put the `noupdate` flag in `new_root` such that any new record created from it gets the noupdate flag.

Doing this post load may miss some newly created records (and children).

src/util/records.py
aj-fuentes: > but if record is noupdate `False` which is created in that case ?. :(

I don't understand what you meant.

sagu-odoo:
Indeed @KangOl , I was thinkin for the case where the record with ``noupdate False``  come for ``update_record_from_xml`` but that will never happen my bad and made fix according to that I have updated the patch and tested also. 

Thanks

src/util/records.py
```diff
@@ -899,10 +899,10 @@ def force_noupdate(cr, xmlid, noupdate=True, warn=False):
             UPDATE ir_model_data
                SET noupdate = %s
              WHERE module = %s
-               AND name = %s
+               AND name ILIKE %s
                AND noupdate != %s
     """,
-        [noupdate, module, name, noupdate],
+        [noupdate, module, "{}%".format(name), noupdate],
```
KangOl: No, this is wrong. You will match unrelated records. I will push the correct fix.

KangOl:
@sagu-odoo can you test with the last version?

KangOl:
Ho! The usage of `direct_inherit_parents` is not enough.
Thanks for pointing it.

sagu-odoo:
Hello @KangOl , I have tested with this new patch but here due to [_cached](https://github.com/odoo/upgrade-util/pull/287/files#diff-e6fd6c801528da788715ef7fac4d24f1e9f0a5b992a021fddc513944d799322bR15) decorator it not working because if model inherit is changed during upgrade  those are not updated because it already in [``_result``](https://github.com/odoo/upgrade-util/blob/daa5888115bf4c9ba89899d4d614c6d0a2c1cafc/src/util/misc.py#L65) and changes regarding noupdate didn't reflected accordingly to that ``_inherits`` record because it never got that model in direct_inherits.

sagu-odoo:
this is no more needed

PR #288, opened 2025-06-27
[IMP] jinja_to_qweb: avoid MemoryError

cawo-odoo:
upg-2994884

Avoid `MemoryError` and/or killed process because of `malloc()` failing within `lxml` / `libxml2`.

Debugging this by determining the size of involved datastructures through means of `sys.getsizeof()` showed that:
1. The global variable `templates_to_check` grows to roughly 1.4GiB after the various calls to `upgrade_jinja_fields()` by upgrade scripts. The process uses ~1.5GiB at that point
2. At the start of function `verify_upgraded_jinja_fields()`, the process is still at ~1.5GiB. While iterating over all the templates in `templates_to_check`, no significant amount of memory is allocated on top of this *to python datastructures*. But, with each call to `is_converted_template_valid()`, the size of the process increases until it hits the RLIMIT. This function calls into `lxml` multiple times, suggesting that the memory is allocated in `malloc()` calls in the C library, evading python's accounting. Internet research shows that `lxml` has a long history of different memory leaks in C code, plus some caching mechanism *across documents* that could be responsible[^1]. More recent versions of the module seem to have been improved, but still we're stuck with old versions.

This patch solves / works around (2) by running the function body of `is_converted_template_valid()` in a forked subprocess that is killed immediately after (instead of using a process pool) and thus no additional memory is ever allocated by `lxml` in the main process, its memory use stays at ~1.5GiB during the runtime of `verify_upgraded_jinja_fields()`. We also add a final line to that function that clears the global data when it is processed, at which point the processes memory use drops to 100MiB.

This patch does *not* solve (1). Solving (2) is enough to fully process upg-2994884, but a future upgrade request may require a fix for (1), too. This will probably be a bigger patch, because I think a viable solution for this would mean to remove the global `templates_to_check` and instead store tha data in the database.

[^1]: https://benbernardblog.com/tracking-down-a-freaky-python-memory-leak-part-2/

### DIFF
--- a/src/util/jinja_to_qweb.py
+++ b/src/util/jinja_to_qweb.py
@@ -3,6 +3,7 @@ import functools
 import html
 import logging
 import re
+from multiprocessing import Process, Queue
 
 import babel
 import lxml
@@ -486,35 +487,46 @@ def verify_upgraded_jinja_fields(cr):
                 "Jinja upgrade",
                 format="html",
             )
+    # done with the global data, clear it for the rest of the upgrade
+    templates_to_check.clear()
 
 
 def is_converted_template_valid(env, template_before, template_after, model_name, record_id, engine="inline_template"):
-    render_before = None
-    with contextlib.suppress(Exception):
-        render_before = _render_template_jinja(env, template_before, model_name, record_id)
+    def callback(q):
+        render_before = None
+        with contextlib.suppress(Exception):
+            render_before = _render_template_jinja(env, template_before, model_name, record_id)
 
-    render_after = None
-    if render_before is not None:
-        try:
-            with mute_logger("odoo.addons.mail.models.mail_render_mixin"):
-                render_after = env["mail.render.mixin"]._render_template(
-                    template_after, model_name, [record_id], engine=engine
-                )[record_id]
-        except Exception:
-            pass
+        render_after = None
+        if render_before is not None:
+            try:
+                with mute_logger("odoo.addons.mail.models.mail_render_mixin"):
+                    render_after = env["mail.render.mixin"]._render_template(
+                        template_after, model_name, [record_id], engine=engine
+                    )[record_id]
+            except Exception:
+                pass
 
-    # post process qweb render to remove comments from the rendered jinja in
-    # order to avoid false negative because qweb never render comments.
-    if render_before and render_after and engine == "qweb":
-        element_before = lxml.html.fragment_fromstring(render_before, create_parent="div")
-        for comment_element in element_before.xpath("//comment()"):
-            comment_element.getparent().remove(comment_element)
+        # post process qweb render to remove comments from the rendered jinja in
+        # order to avoid false negative because qweb never render comments.
+        if render_before and render_after and engine == "qweb":
+            element_before = lxml.html.fragment_fromstring(render_before, create_parent="div")
+            for comment_element in element_before.xpath("//comment()"):
+                comment_element.getparent().remove(comment_element)
         render_before = lxml.html.tostring(element_before, encoding="unicode")
         render_after = lxml.html.tostring(
             lxml.html.fragment_fromstring(render_after, create_parent="div"), encoding="unicode"
         )
 
-    return render_before is not None and render_before == render_after
+        q.put(render_before is not None and render_before == render_after)
+
+    queue = Queue()
+    proc = Process(target=callback, args=((queue,)))
+    proc.start()
+    res = queue.get(timeout=60)
+    if proc.is_alive():
+        proc.kill()
+    return res
 
 
 # jinja render


aj-fuentes:
From discussion in DM: I agree we should move the data to the DB. One strategy is to do so only if it is too big.

Another thing: you could reuse the process for multiple calls to `is_converted_template_valid`. In other words let it process something like (randomly chosen) 100 items before recreating it. You can cache the info in `is_converted_template_valid`. To decide better it would be great if you post a summary of the value of `template_data` during the upgrade. How many entries are there? For how many tables? How many fields? Or the size of templates?

cawo-odoo:
> To decide better it would be great if you post a summary of the value of `template_data` during the upgrade. How many entries are there? For how many tables? How many fields? Or the size of templates?

```
(Pdb) len(templates_to_check)
4                                                                              )
(Pdb) sum([len(template_data) for template_data in templates_to_check.values()]) 
7364
(Pdb) sum([len(entry[5]) for template_data in templates_to_check.values() for entry in template_data])
52298
(Pdb) sum([len(converted) for template_data in templates_to_check.values() for entry in template_data for converted in entry[5].values()])
342645888
(Pdb) max([len(converted) for template_data in templates_to_check.values() for entry in template_data for converted in entry[5].values()])
335631
(Pdb)
```
I.e. in total there are 52298 templates to verify with a total string length of 342645888 and a maximum individual length of  335631. So yes, there's probably room for "chunking" to not loose too much time on forking (OTOH the whole upgrade 13.0 -> 17.0 ran in under half an hour in my test).

> I agree we should move the data to the DB

working on it right now.

src/util/jinja_to_qweb.py
```diff
@@ -246,70 +288,78 @@ def upgrade_jinja_fields(
     sql_where_qweb_fields = [field + r"~ '(\$\{|%\s*(if|for))'" for field in qweb_fields]
     sql_where_fields = " OR ".join(sql_where_inline_fields + sql_where_qweb_fields)
 
-    templates_to_check[table_name] = []
+    setup_templates_to_check(cr)
     model = model_of_table(cr, table_name)
 
     cr.commit()  # ease the processing for PG
-    cr.execute(
+    chunk_cr = db_connect(cr.dbname).cursor()
```
aj-fuentes: For this we prefer `util.named_cursor`, unless you have a different reason than being able to fetch by chunks interspersed with other queries.

cawo-odoo: The reason is that `util.named_cursor` returns a naked psycopg2 cursor, but I wanted to keep using **dict**fetchmany(), or I would have been forced to imitate it => even more code.

aj-fuentes: Did you try?
https://github.com/odoo/upgrade-util/blob/5adc59f66d3b565c2af8c5b2f0fd980712cfb55a/src/util/pg.py#L1545-L1546

src/util/jinja_to_qweb.py
```diff
@@ -246,70 +288,78 @@ def upgrade_jinja_fields(
...
-    for data in cr.dictfetchall():
-        _logger.info("process %s(%s) %s", table_name, data["id"], data[name_field])
-
-        # convert the fields
-        templates_converted = {}
+    while True:
+        data_chunk = chunk_cr.dictfetchmany(100)
+        if not data_chunk:
+            break
+        for data in data_chunk:
```
aj-fuentes: I'd suggest you move this to a generator then use the generator in the main loop. This will minimize your diff and make it easier to review and keep a better line history.
```suggestion
    datas = itertools.chain.from_iterable(itertools.takewhile(bool, map(ncr.dictfetchmany, itertools.repeat(300))))
    for data in datas:
```

src/util/jinja_to_qweb.py
```diff
@@ -486,35 +545,46 @@ def verify_upgraded_jinja_fields(cr):
...
+            render_before = lxml.html.tostring(element_before, encoding="unicode")
+            render_after = lxml.html.tostring(
+                lxml.html.fragment_fromstring(render_after, create_parent="div"), encoding="unicode"
+            )
+
+        q.put(render_before is not None and render_before == render_after)
 
-    return render_before is not None and render_before == render_after
+    queue = Queue()
+    proc = Process(target=callback, args=((queue,)))
```
aj-fuentes: `((x,)) -> (x,)`, and you can avoid the odd dangling comma with a list (optional)
```suggestion
    proc = Process(target=callback, args=[queue])
```

aj-fuentes: For the record: I'm OK with single shot forks since they are fast.

src/util/jinja_to_qweb.py
```diff
@@ -226,6 +226,48 @@ def _replace_endif(string):
     return reg.sub(r"\1</t>", string)
 
 
+def setup_templates_to_check(cr):
```
aj-fuentes, line 228: Since you are doing this (little, laser-focused, functions), you could create a little object that inherits `dict` and is a dict until too much data is added to it. At which point it flushes data to a table and uses queries for insert and lookup. Then keep the original code almost unchanged.

The thing is that we can keep using the dict approach for the vast majority of the upgrades.

EDIT: note that this is just an idea... it may be not so simple or actually worse, still worth exploring IMO.

src/util/jinja_to_qweb.py
cawo-odoo: :rofl: :man_facepalming: :clown_face: :hankey:

src/util/jinja_to_qweb.py
cawo-odoo: Even better: the named_cursor provides that generator: `iterdict()` :-)

aj-fuentes: Yeah my past me worked on that for current you https://github.com/odoo/upgrade-util/commit/7cd398ba52239450b229067324344867cbb5f471 ... while current me didn't recall all the details :D

src/util/jinja_to_qweb.py
cawo-odoo: > Since you are doing this (little, laser-focused, functions), you could create a little object that inherits `dict` and is a dict until too much data is added to it. At which point it flushes data to a table and uses queries for insert and lookup.

I think this would be a nice approach if we want to avoid the single-row inserts and instead insert 100 rows at once or so.

>  Then keep the original code almost unchanged.

The issue is that the dict contains a mix of different nested datastructures that are not well suited for insertion in a single flat table. I did not want to use serialisation (like json) to achieve that, so instead I chose to flatten the data to exactly what is needed in the validation step and what's easily inserted into a single table. So even if we do a mix of in-memory and DB table, I would rather go for the flattened data, since it also removes superfluous elements from the structure and that means I would still do the changes to the code supporting the changed structure.

cawo-odoo: Here's a version that inserts in batches. It seems to make no significant difference, though: run-time difference is in the range of seconds, hard to compare.

KangOl: > Here's a version that inserts in batches. It seems to make no significant difference, though: run-time difference is in the range of seconds, hard to compare.

I don't think it worth the added complexity. The previous patch was clear enough.

cawo-odoo: > I don't think it worth the added complexity. The previous patch was clear enough.

That's exactly what I think.

KangOl (approved):
@robodoo r+

PR #289, opened 2025-07-02
[IMP] util.update_record_from_xml

KangOl:
Set the `xml_filename` argument of the `xml_import` records.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import uuid
+from collections import OrderedDict  # used for python2 compatibility
 from contextlib import contextmanager
 from operator import itemgetter
 
@@ -12,12 +13,12 @@ from psycopg2 import sql
 from psycopg2.extras import Json, execute_values
 
 try:
-    from odoo import modules, release
+    from odoo import modules
     from odoo.tools.convert import xml_import
     from odoo.tools.misc import file_open
     from odoo.tools.translate import xml_translate
 except ImportError:
-    from openerp import modules, release
+    from openerp import modules
     from openerp.tools.convert import xml_import
     from openerp.tools.misc import file_open
 
@@ -34,7 +35,7 @@ from .helpers import (
 from .inconsistencies import break_recursive_loops
 from .indirect_references import indirect_references
 from .inherit import direct_inherit_parents, for_each_inherit
-from .misc import AUTOMATIC, chunks, parse_version, version_gte
+from .misc import AUTOMATIC, chunks, version_between, version_gte
 from .orm import env, flush
 from .pg import (
     PGRegexp,
@@ -1152,9 +1153,7 @@ def __update_record_from_xml(
     )
     xpath = "//*[self::act_window or self::menuitem or self::record or self::report or self::template][{}]".format(id_match)  # fmt: skip
 
-    # use a data tag inside openerp tag to be compatible with all supported versions
-    new_root = lxml.etree.fromstring("<openerp><data/></openerp>")
-
+    roots = OrderedDict()
     manifest = get_manifest(from_module)
     template = False
     found = False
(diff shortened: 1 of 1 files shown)


src/util/records.py
```diff
@@ -1147,6 +1148,8 @@ def add_ref(ref):
             doc = lxml.etree.parse(fp)
             for node in doc.xpath(xpath):
                 found = True
+                if xml_filename is None and module == from_module:
```
aj-fuentes: Why do we check module vs from_module? Don't we need to set the xml_filename to be the actual file always?

KangOl: It's to avoid setting the filename when we load an overridden record (initially declared in another module).
However, the models that use this `xml_filename` (views, mail templates) are not usually changed by other modules (and even less updated via `update_record_from_xml`), so I can remove this condition.

aj-fuentes: I think we should do what standard code does.

src/util/records.py
```diff
@@ -1147,6 +1148,8 @@ def add_ref(ref):
             doc = lxml.etree.parse(fp)
             for node in doc.xpath(xpath):
                 found = True
+                if xml_filename is None:
```
aj-fuentes: I wonder if we now need the `found` variable? We can keep the xml_filename as the "found" signal whenever it's set. Maybe named `found_in_xml_file`?

There is a slight subtlety. We do not break the search loop when found. Thus in practice we keep searching and adding nodes to our new root. This means that we will be lying if we keep the same filename when loading our new root. Maybe we should have a `new_root` per `found_in_xml_file`, a sort of map `filename->new_root`. We should double check the standard behavior if we really care about this.

EDIT: This is obviously a corner case since it's a mistake to have two records with same xmlid redefined in the same module in xml. But this has been actually seen before. Aligning with standard code would be "saner" even for the corner case.

KangOl:
After some tests, it doesn't work as expected in all cases.
 - `ir.ui.view`: works correctly due to the `arch` field being computed with an `inverse` function
 - `theme.ir.ui.view`: just read in the `default` function, so not updated during `write`
 - `template.reset.mixin` (`{mail,sms}.template`): set in the `create` overwrite only.

Maybe all those usages should be standardized in a common mixin which also handle the update.

aj-fuentes:
> After some tests, it doesn't work as expected in all cases.
> 
>     * `ir.ui.view`: works correctly due to the `arch` field being computed with an `inverse` function
> 
>     * `theme.ir.ui.view`: just read in the `default` function, so not updated during `write`
> 
>     * `template.reset.mixin` (`{mail,sms}.template`): set in the `create` overwrite only.

* `theme.ir.ui.view`: can't we just add `@api.depends('arch')` here https://github.com/odoo/odoo/blob/07626050bd0104fecd8799b56d30245d00da3f27/addons/website/models/theme_models.py#L56 ?
* `template.reset.mixin`: we can do in `write` the same as in `create`, https://github.com/odoo/odoo/blob/07626050bd0104fecd8799b56d30245d00da3f27/addons/mail/models/template_reset_mixin.py#L30-L36 (in master it could be extracted to a local private method...)

> Maybe all those usages should be standardized in a common mixin which also handle the update.

I think this could be an option if this becomes more widely use. For current cases it looks very localized and (if I'm not mistaked) with a simple solution. The mixin may add extra complexity that we don't need for these few cases.

KangOl:
>     * `theme.ir.ui.view`: can't we just add `@api.depends('arch')` here https://github.com/odoo/odoo/blob/07626050bd0104fecd8799b56d30245d00da3f27/addons/website/models/theme_models.py#L56 ?
> 

No, it's not enough. It should be set in `create` and `write`, like for other models.

aj-fuentes:
> > * `theme.ir.ui.view`: can't we just add `@api.depends('arch')` here https://github.com/odoo/odoo/blob/07626050bd0104fecd8799b56d30245d00da3f27/addons/website/models/theme_models.py#L56 ?
> 
> No, it's not enough. It should be set in `create` and `write`, like for other models.

What I had in mind was that the field should be a computed stored field, with "depends". I missed that it was just `default=...`

KangOl:
> EDIT: This is obviously a corner case since it's a mistake to have two records with same xmlid redefined in the same module in xml. But this has been actually seen before. Aligning with standard code would be "saner" even for the corner case.

You're right that it would be more correct to load records per file.

KangOl:

src/util/records.py
```diff
@@ -1173,15 +1172,20 @@ def add_ref(ref):
             doc = lxml.etree.parse(fp)
             for node in doc.xpath(xpath):
                 found = True
+                xml_filename = "{}/{}".format(from_module, f)
+                root = roots.get(xml_filename)
+                if root is None:
+                    # use a data tag inside openerp tag to be compatible with all supported versions
+                    root = lxml.etree.fromstring("<openerp><data/></openerp>")
```
aj-fuentes: Don't we need to set the value in the dict as well? Maybe I missed it...
```suggestion
                    root = roots[xml_filename] = lxml.etree.fromstring("<openerp><data/></openerp>")
```

KangOl: 🤦

aj-fuentes (approved):

KangOl:
@robodoo r+

PR #291, opened 2025-07-08
[FIX] util.force_upgrade_of_fresh_module

KangOl:
The purpose of this function is to cheat about upgrade status of a "to install" module.

Before saas~18.2:
As the `MigrationManager` only handle the "to upgrade" modules, we need a way to tell the ORM that, even it is in "to upgrade", the module should be loaded in "init" mode (this control the behavior of the `forcecreate` flag of the records). That was done using the global `config["init"]` dict.

Since saas~18.2:
Now, this is done the other way around. It's the `MigrationManager` that is aware of "to install" modules that upgrade scripts should be executed. This is done via the `_force_upgrade_scripts` set on the registry. The module state stay in "to install".

In either case, we should set the module version to the version of the calling script.

Note: this only affect the `init=True` calling mode (default). When called with `init=False`, we always lie about the module state and don't tell the ORM about it.

Oversight of odoo/upgrade-util#176

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -897,26 +897,26 @@ def _force_upgrade_of_fresh_module(cr, module, init, version):
     # Low level implementation
     # Force module state to be in `to upgrade`.
     # Needed for migration script execution. See http://git.io/vnF7f
-    if version_gte("saas~18.2") and init:
-        env_ = env(cr)
-        env_.registry._force_upgrade_scripts.add(module)
-        return
-
+    state = "to install" if init and version_gte("saas~18.2") else "to upgrade"
     cr.execute(
         """
             UPDATE ir_module_module
-               SET state='to upgrade',
+               SET state=%s,
                    latest_version=%s
              WHERE name=%s
                AND state='to install'
          RETURNING id
     """,
-        [version, module],
+        [state, version, module],
     )
     if init and cr.rowcount:
-        # Force module in `init` mode beside its state is forced to `to upgrade`
-        # See http://git.io/vnF7O
-        odoo.tools.config["init"][module] = "oh yeah!"
+        if version_gte("saas~18.2"):
+            env_ = env(cr)
+            env_.registry._force_upgrade_scripts.add(module)
+        else:
+            # Force module in `init` mode beside its state is forced to `to upgrade`
+            # See http://git.io/vnF7O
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only l10n_ch_hr_payroll_elm_transmission_account l10n_in_reports website_crm_partner_assign hr_homeworking+calendar in 18.0 master

aj-fuentes (approved):
LGTM

Besides the goal of this PR, there is a slight difference with what we did before: now we only add the module to `_force_upgrade_scripts` if it has state `to install` in `ir_module_module`. Before we always added it. I believe this is actually correct.

src/util/modules.py
```diff
@@ -897,26 +897,26 @@ def _force_upgrade_of_fresh_module(cr, module, init, version):
     # Low level implementation
     # Force module state to be in `to upgrade`.
     # Needed for migration script execution. See http://git.io/vnF7f
-    if version_gte("saas~18.2") and init:
-        env_ = env(cr)
-        env_.registry._force_upgrade_scripts.add(module)
-        return
-
+    state = "to install" if init and version_gte("saas~18.2") else "to upgrade"
```
HydrionBurst, line 900: NIT: shall we simply
```py
if not init:
    return
```

HydrionBurst (approved):
In general, LGTM

> Indeed. But the outcome is the same, the module kept it state. It only polluted the `_force_upgrade_scripts` set.

For now, technically, `registry._force_upgrade_scripts` has no side-effects

TBD
Is it possible we
0. load base
1. there is a module_A with `to upgrade`
2. `_force_upgrade_of_fresh_module` for module_A
3. do something else and there is a hack to change module_A.state from `to upgrade` to `to install`
4. contuine loading other modules
5. Then we may still want to execute module_A's upgrade script

src/util/modules.py
KangOl: No. As stated in the commit message:
> Note: this only affect the `init=True` calling mode (default). When called with `init=False`, we always lie about the module state and don't tell the ORM about it.

KangOl:
> do something else and there is a hack to change module_A.state from `to upgrade` to `to install`

This should never happen. I don't think of a valid case for this that couldn't be solved in another way.

KangOl:
@robodoo r+

PR #292, opened 2025-07-08
[IMP] get_module_path: remove downloaded

HydrionBurst:
tools.config.addons_data_dir is already in odoo.addons.__path__, the parameter ``downloaded`` is not necessary and removed since saas-18.5.

https://github.com/odoo/odoo/pull/217924

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -130,7 +130,8 @@ def has_enterprise():
     if os.getenv("ODOO_HAS_ENTERPRISE"):
         return True
     # XXX maybe we will need to change this for version > 9
-    return bool(get_module_path("delivery_fedex", downloaded=False, display_warning=False))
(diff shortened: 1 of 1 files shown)


src/util/misc.py
```diff
@@ -130,6 +130,8 @@ def has_enterprise():
     if os.getenv("ODOO_HAS_ENTERPRISE"):
         return True
     # XXX maybe we will need to change this for version > 9
+    if version_gte("saas~18.5"):
+        return bool(get_module_path("delivery_fedex", display_warning=False))
     return bool(get_module_path("delivery_fedex", downloaded=False, display_warning=False))
```
```suggestion
    kw = {} if version_gte("saas~18.5") else {"downloaded": False}
    return bool(get_module_path("delivery_fedex", display_warning=False, **kw))
```

PR #293, opened 2025-07-08
[IMP] tools: separate compilation/assertion code object

thle-odoo:
Adapt util.

Community PR: https://github.com/odoo/odoo/pull/217284


KangOl:
upgradeci retry with always only hr in master

PR #294, opened 2025-07-09
[IMP] snippets: prepare for changed default in python 3.14

cawo-odoo:
Cited from documentation:
> Note  The default multiprocessing start method (see Contexts and start
> methods) will change away from fork in Python 3.14. Code that requires fork
> be used for their ProcessPoolExecutor should explicitly specify that by
> passing a `mp_context=multiprocessing.get_context("fork")` parameter.[^1]

[^1]: https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor

### DIFF
--- a/src/util/snippets.py
+++ b/src/util/snippets.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 import logging
+import multiprocessing
 import re
+import sys
 from concurrent.futures import ProcessPoolExecutor
 
 from lxml import etree, html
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only sms hr_recruitment mrp account_reports crm spreadsheet_edition documents base

KangOl:
upgradeci retry with always only base mass_mailing mass_mailing_themes point_of_sale test_themes theme_common theme_treehouse web web_editor website website_blog website_enterprise website_event website_mass_mailing website_sale

PR #295, opened 2025-07-09
[IMP] snippets: preserve same file module

cawo-odoo:
In `make_pickleable_callback`, when creating a name for the module created via import_script, no longer make the name random, but derive it from the script's path deterministically via a hash function. This way, when this method is called multiple times from the same script for multiple callbacks, it will only import the script once and otherwise re-use the created module. This ensures multiple callbacks from the same script file will run within the same module, seeing e.g. the same values in modified globals, which is much more intuitive.

Also, move the function to `misc.py`. With the intention to reuse it, it's a better fit. For that, make it python2 compatible.

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -4,9 +4,12 @@ import ast
 import collections
 import datetime
 import functools
+import hashlib
+import inspect
 import logging
 import os
 import re
+import sys
 import textwrap
 import uuid
 from contextlib import contextmanager
@@ -21,7 +24,7 @@ except ImportError:
     from openerp.modules.module import get_module_path
     from openerp.tools.parse_version import parse_version
 
-from .exceptions import SleepyDeveloperError
+from .exceptions import MigrationError, SleepyDeveloperError
 
 # python3 shim
 try:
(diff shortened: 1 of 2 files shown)


src/util/snippets.py
```diff
@@ -171,10 +171,11 @@ def make_pickleable_callback(callback):
     We should then put the executed function in its own importable file.
     """
     callback_filepath = inspect.getfile(callback)
-    name = f"_upgrade_{uuid.uuid4().hex}"
-    mod = sys.modules[name] = import_script(callback_filepath, name=name)
+    name = f"_upgrade_{hashlib.sha256(callback_filepath.encode()).hexdigest()}"
+    if not sys.modules[name]:
```
aj-fuentes:
```suggestion
    if name not in sys.modules:
```

src/util/snippets.py
```diff
@@ -171,10 +171,11 @@ def make_pickleable_callback(callback):
     We should then put the executed function in its own importable file.
     """
```
aj-fuentes: Add a meta comment here to exclude this from the online docs.

src/util/misc.py
```diff
@@ -388,6 +393,30 @@ def log(chunk_num, size=chunk_size):
         log(i // chunk_size + 1, i % chunk_size)
 
 
+def make_pickleable_callback(callback):
+    """
+    .. noautodoc::
```
cawo-odoo: > Add a meta comment here to exclude this from the online docs.

This should be the one for sphinx. I didn't find a single occurrence in our code anywhere else. `pre-commit` is not amused. Is it not what you meant?  Why do we need to hide it?

src/util/misc.py
```diff
@@ -4,14 +4,19 @@
...
 import logging
 import os
 import re
+import sys
 import textwrap
 import uuid
 from contextlib import contextmanager
 from itertools import chain, islice
 
+from .exceptions import MigrationError
```
KangOl: Do it with the import of `SleepyDeveloperError`

src/util/misc.py
KangOl: It's the `:meta private: exclude from online docs` instruction.

aj-fuentes: > Is it not what you meant? 

No, it's what' @KangOl put above.

> Why do we need to hide it?

We don't want to describe online advanced (one can say obscure) functions that we mostly use internally.

You can verify the final status of your PR w.r.t the online docs by opening the `documentation` build then click in the green "View result" button
image

If you see the function documented in the docs under `Developers/Reference/Upgrade/Upgrade utils` then something is wrong.

cawo-odoo: oh. the nearest example is only 20 lines away ... I'm the SleepyDeveloper today. Thanks :)

KangOl:
upgradeci retry with always only website documents mrp spreadsheet_editions

PR #297, opened 2025-07-18
[ADD] util/bulk_update_table

Pirols:
A recurrent challenge in writing upgrade scripts is that of updating values in a table based on some form of already available mapping from the id (or another identifier) to the new value, this is often addressed with an iterative solution in the form:

```python
for key, value in mapping.items():
    cr.execute(
        """
        UPDATE table
           SET col = %s
         WHERE key_col = %s
        """,
        [value, key],
    )
```

or in a more efficient (only issuing a single query) but hacky way:

```python
cr.execute(
    """
    UPDATE table
       SET col = (%s::jsonb)->>(key_col::text)
     WHERE key_col = ANY(%s)
    """,
    [json.dumps(mapping), list(mapping)],
)
```

With the former being ineffective for big mappings and the latter often requiring some comments at review time to get it right. This commit introduces a util meant to make it easier to efficiently perform such updates.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -881,6 +881,115 @@ class TestPG(UnitTestCase):
         cr.execute(util.format_query(cr, "SELECT 1 FROM {}", TEST_TABLE_NAME))
         self.assertFalse(cr.rowcount)
 
+    def test_update_one_col_from_dict(self):
+        TEST_TABLE_NAME = "_upgrade_bulk_update_one_col_test_table"
+        N_ROWS = 10
+
+        cr = self._get_cr()
+
+        cr.execute(
+            util.format_query(
+                cr,
+                """
+                DROP TABLE IF EXISTS {table};
+
+                CREATE TABLE {table} (
+                    id SERIAL PRIMARY KEY,
+                    col1 INTEGER,
+                    col2 INTEGER
+                );
+
+                INSERT INTO {table} (col1, col2) SELECT v, v FROM GENERATE_SERIES(1, %s) as v;
+                """,
+                table=TEST_TABLE_NAME,
+            ),
+            [N_ROWS],
+        )
+        mapping = {id: id * 2 for id in range(1, N_ROWS + 1, 2)}
+        util.bulk_update_table(cr, TEST_TABLE_NAME, "col1", mapping)
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE col2 != id",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount, "unintended column 'col2' is affected")
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE col1 != id AND MOD(id, 2) = 0",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount, "unintended rows are affected")
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE col1 != 2 * id AND MOD(id, 2) = 1",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount, "partial/incorrect updates are performed")
+
+    def test_update_multiple_cols_from_dict(self):
+        TEST_TABLE_NAME = "_upgrade_bulk_update_multiple_cols_test_table"
+        N_ROWS = 10
+
+        cr = self._get_cr()
+
+        cr.execute(
+            util.format_query(
+                cr,
+                """
+                DROP TABLE IF EXISTS {table};
+
+                CREATE TABLE {table} (
+                    id SERIAL PRIMARY KEY,
+                    col1 INTEGER,
+                    col2 INTEGER,
+                    col3 INTEGER
+                );
+
+                INSERT INTO {table} (col1, col2, col3) SELECT v, v, v FROM GENERATE_SERIES(1, %s) as v;
+                """,
+                table=TEST_TABLE_NAME,
+            ),
+            [N_ROWS],
+        )
+        mapping = {id: [id * 2, id * 3] for id in range(1, N_ROWS + 1, 2)}
+        util.bulk_update_table(cr, TEST_TABLE_NAME, ["col1", "col2"], mapping)
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE col3 != id",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount, "unintended column 'col3' is affected")
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE col1 != id AND MOD(id, 2) = 0",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount, "unintended rows are affected")
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE (col1 != 2 * id OR col2 != 3 * id) AND MOD(id, 2) = 1",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount, "partial/incorrect updates are performed")
+
     def test_create_column_with_fk(self):
         cr = self.env.cr
         self.assertFalse(util.column_exists(cr, "res_partner", "_test_lang_id"))
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -32,6 +32,7 @@ except NameError:
 import psycopg2
 from psycopg2 import errorcodes, sql
 from psycopg2.extensions import quote_ident
+from psycopg2.extras import Json
 
 try:
     from odoo.modules import module as odoo_module
@@ -1621,3 +1622,76 @@ def create_id_sequence(cr, table, set_as_default=True):
                 table=table_sql,
             )
         )
+
+
+def bulk_update_table(cr, table, columns, mapping, key_col="id"):
+    """
+    Update table based on mapping.
+
+    Each `mapping` entry defines the new values for the specified `columns` for the row(s)
+    whose `key_col` value matches the key.
+
+    .. example::
+
+       .. code-block:: python
+
+          # single column update
+          util.bulk_update_table(cr, "res_users", "active", {42: False, 27: True})
+
+          # multi-column update
+          util.bulk_update_table(
+              cr,
+              "res_users",
+              ["active", "password"],
+              {
+                  "admin": [True, "1234"],
+                  "demo": [True, "5678"],
+              },
+              key_col="login",
+          )
+
+    :param str table: table to update.
+    :param str | list(str) columns: columns spec for the update. It could be a single
+                                    column name or a list of column names. The `mapping`
+                                    must match the spec.
+    :param dict mapping: values to set, which must match the spec in `columns`,
+                         following the **same** order
+    :param str key_col: column used as key to get the values from `mapping` during the
+                        update.
+
+    .. warning::
+
+       The values in the mapping will be casted to the type of the target column.
+       This function is designed to update scalar values, avoid setting arrays or json
+       data via the mapping.
+    """
+    _validate_table(table)
+    if not columns or not mapping:
+        return
+
+    assert isinstance(mapping, dict)
+    if isinstance(columns, str):
+        columns = [columns]
+    else:
+        n_columns = len(columns)
+        assert all(isinstance(value, (list, tuple)) and len(value) == n_columns for value in mapping.values())
+
+    query = format_query(
+        cr,
+        """
+        UPDATE {table} t
+           SET ({cols}) = ROW({cols_values})
+          FROM JSONB_EACH(%s) m
+         WHERE t.{key_col}::text = m.key
+        """,
+        table=table,
+        cols=ColumnList.from_unquoted(cr, columns),
+        cols_values=SQLStr(
+            ", ".join(
+                "(m.value->>{:d})::{}".format(col_idx, column_type(cr, table, col_name))
+                for col_idx, col_name in enumerate(columns)
+            )
+        ),
+        key_col=key_col,
+    )
+    cr.execute(query, [Json(mapping)])


KangOl:

src/util/pg.py
```diff
@@ -1621,3 +1622,89 @@ def create_id_sequence(cr, table, set_as_default=True):
                 table=table_sql,
             )
         )
+
+
+def update_table_from_dict(cr, table, mapping, key_col="id", bucket_size=DEFAULT_BUCKET_SIZE):
```
aj-fuentes: I think the interface should be
```suggestion
def update_table_from_dict(cr, table, columns, mapping, key_col="id", bucket_size=DEFAULT_BUCKET_SIZE):
```
From experience, when we use a dict to update, sometimes the dict becomes too big. If we make the keys repeated each time we may unintentionally make this harder. We can instead take a mapping `{id: [val1, val2,...]}` and a spec of columns to update `[col1, col2, ...]`, or `{id: val}` with just `col1`. This has the advantage that for the most common case where we update a single column the values could be scalar as well. When `columns` is a list then the values in the mapping must be a list as well.
Example usage:
```py
update_table_from_dict(cr, table, "name", {1: "foo", 2: "bar"})
update_table_from_dict(cr, table, ["name", "passwd"], {1: ["foo", "baz"], 2: ["bar", ""]})
```

src/util/pg.py
```diff
@@ -1621,3 +1622,100 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+           SET ({cols}) = ROW({cols_values})
+          FROM JSONB_EACH(%s) m
+         WHERE t.{key_col}::varchar = m.key
+        """,
+        table=table,
+        cols=ColumnList.from_unquoted(cr, columns),
+        cols_values=SQLStr(
+            ", ".join(
+                "(m.value->>{:d})::{}".format(
+                    col_idx, sql.Identifier(column_type(cr, table, col_name)).as_string(cr._cnx)
```
KangOl: The column type is not an identifier...

src/util/pg.py
```diff
@@ -1621,3 +1622,100 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+
+       As a side effect, the cursor may be committed.
+
+    :param str table: database's table to perform the update of
+    :param str | list[str] columns: table's columns to update
+    :param dict[any, any | list[any]] mapping: matches values of `key_col` to the new `columns` values
+    :param str key_col: column to match against keys of `mapping`
+    :param int bucket_size: maximum number of rows to update per single query
+    """
+    _validate_table(table)
```
KangOl, line 1668: I would also like to have a validation of the `mapping` 
```suggestion
    _validate_table(table)
    assert isinstance(mapping, dict) and all(isinstance(key, int) and isinstance(value, (list, tuple)) for key, value in mapping.items())
```

src/base/tests/test_util.py
```diff
@@ -881,6 +881,115 @@ def test_parallel_execute_retry_on_serialization_failure(self):
...
+                    id SERIAL PRIMARY KEY,
+                    col1 INTEGER,
+                    col2 INTEGER
+                );
+
+                INSERT INTO {table} (col1, col2) SELECT v, v FROM GENERATE_SERIES(1, %s) as v;
+                """
+                % N_ROWS,
+                table=TEST_TABLE_NAME,
+            )
```
```suggestion
                """,
                table=TEST_TABLE_NAME,
            ),
            [N_ROWS],
```

src/base/tests/test_util.py
```diff
@@ -881,6 +881,115 @@ def test_parallel_execute_retry_on_serialization_failure(self):
...
+        util.update_table_from_dict(cr, TEST_TABLE_NAME, "col1", mapping)
+
+        cr.execute(
+            util.format_query(
+                cr,
+                "SELECT id FROM {table} WHERE col2 != id",
+                table=TEST_TABLE_NAME,
+            )
+        )
+        self.assertFalse(cr.rowcount)  # otherwise unintended column is affected
```
```suggestion
        self.assertFalse(cr.rowcount, "unintended column 'col2' is affected")
```

src/util/pg.py
Pirols: I know. It's an improper use to make it safe to inject. Alternative proposals? Can we just trust the output of our util?

KangOl: Yes, `column_type` can be trusted. You can inject it [AsIs](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.AsIs) (or with `SQLStr`)

src/util/pg.py
Pirols: Do we want to keep `key_col` and limit its potential targets to integer columns or remove it entirely?

src/util/pg.py
aj-fuentes: We could modify `util.column_type` to return `SQLStr(nfo[0]) ...`
https://github.com/odoo/upgrade-util/blob/6c9a30c2e0125194154de24b0a1380b4db836652/src/util/pg.py#L543

src/util/pg.py
aj-fuentes: I think `key_col` should always be an `int`, but I like to have it variable. It would allow to update a table based in a FK instead of a PK. To decide you can grep the repo and see if we currently use the dict trick for other columns than `id`... I think we did but my memory could betray me.

src/util/pg.py
```diff
@@ -1621,3 +1622,86 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+        cols_values=SQLStr(
+            ", ".join("(m.value->>{:d})::{}".format(col_idx, column_type(cr, table, col_name)))
+            for col_idx, col_name in enumerate(columns)
+        ),
+        key_col=key_col,
+    )
+
+    if len(mapping) <= 1.1 * bucket_size:
+        cr.execute(query, [json.dumps(mapping)])
+    else:
```
aj-fuentes: I think we could remove this part to keep it simple. It's a niche case?
```suggestion
```

src/util/pg.py
```diff
@@ -1621,3 +1622,86 @@ def create_id_sequence(cr, table, set_as_default=True):
                 table=table_sql,
             )
         )
+
+
+def update_table_from_dict(cr, table, columns, mapping, key_col="id", bucket_size=DEFAULT_BUCKET_SIZE):
```
aj-fuentes: Same here. We could remove the bucket_size parameter. In most cases when we **need** to use a dict update is because some process was necessarily done in Python. Thus it should be the caller responsibility to chunk it anyway. Moreover this would allow using the new util from a callback in a process pool for example. I mean, it's still usable but spanning a parallel run from a child process would be a bit messy.

src/util/pg.py
Pirols: From a relatively quick search I found 4 usages of the json trick on columns other than `id`:

* varchar column (2 usages): https://github.com/odoo/upgrade/blob/dad3dd7a1554c526d37fd5f61e6f4a9e18667e56/migrations/sale_subscription/saas~15.3.1.1/pre-migrate.py#L588-L610 and https://github.com/odoo/upgrade/blob/dad3dd7a1554c526d37fd5f61e6f4a9e18667e56/migrations/sale_subscription/saas~15.3.1.1/pre-migrate.py#L612-L634
* varchar column: https://github.com/odoo/upgrade/blob/dad3dd7a1554c526d37fd5f61e6f4a9e18667e56/migrations/account/saas~17.4.1.2/pre-migrate.py#L51-L63
* integer column: https://github.com/odoo/upgrade/blob/dad3dd7a1554c526d37fd5f61e6f4a9e18667e56/migrations/l10n_in_hr_payroll/saas~17.5.1.0/post-migrate.py#L21-L43

src/util/pg.py
Pirols: Forgot to answer, but it's a better interface indeed. :)

src/util/pg.py
Pirols: You're right, better keep such a util orthogonal to any parallelisation.

src/util/pg.py
aj-fuentes: OK, then maybe it's worth taking the type from the `key_col` in the table directly.

Pirols: That should not be necessary since `JSONB_EACH` returns the key as a text[^1]. 
Unless we want to validate the type of the keys of `mapping`. In which case I would need to define (or reuse, if already available) a mapping from pg types to python types. Wdyt?

[^1]: https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING-TABLE

aj-fuentes: Well I just noticed you are casting the key_col to text before comparing with the actual json dict key, then it's fine. :) What I had in mind was casting the dict key to the column type. In principle my suggestion was to block/fail if we get an invalid json dict (incompatible with the table), I have no strong opinion against current approach TBH. @KangOl wdyt?

src/util/pg.py
```diff
@@ -1621,3 +1622,72 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+        cols=ColumnList.from_unquoted(cr, columns),
+        cols_values=SQLStr(
+            ", ".join(
+                "(m.value->>{:d})::{}".format(col_idx, column_type(cr, table, col_name))
+                for col_idx, col_name in enumerate(columns)
+            )
+        ),
+        key_col=key_col,
+    )
+    cr.execute(query, [json.dumps(mapping)])
```
```suggestion
    cr.execute(query, [Json(mapping)])
```
imported from `psycopg2.extras`

src/util/pg.py
```diff
@@ -1621,3 +1622,78 @@ def create_id_sequence(cr, table, set_as_default=True):
                 table=table_sql,
             )
         )
+
+
+def update_table_from_dict(cr, table, columns, mapping, key_col="id"):
```
KangOl: Nitpicking on the function name.
```suggestion
def bulk_update_table(cr, table, columns, mapping, key_col="id"):
```

Pirols: Better, indeed.

src/util/pg.py
```diff
@@ -1621,3 +1622,78 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+              {
+                  1: [2, True],
+                  2: [3, False],
+              },
+          )
+
+    :param str table: database's table to perform the update of
+    :param str | list[str] columns: table's columns to update
+    :param dict[any, any | list[any]] mapping: matches values of `key_col` to new values for each of `columns` (in the **same** order)
+    :param str key_col: `ŧable`'s column to match against keys of `mapping`
```
aj-fuentes: Why `ŧable`? Is it a table with two boards `ŧ` vs `t` :trollface: 

Now, for real, we can short a bit the docs, be more direct.
```suggestion
    :param str table: table to update
    :param str | list columns: columns spec for the update. It could be a single column
                               name or a list of column names. The `mapping` must match
                               the spec.
    :param dict mapping: values to set. The values must match the spec in `columns`,
                         with **same** order.
    :param str key_col: column used as key to get the values from `mapping` during
                        the update.
```
The typing info (a-la mypy) doesn't look nice. We could put something simpler as we do elsewhere here.
image

src/util/pg.py
```diff
@@ -1621,3 +1622,78 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+
+    .. warning::
+
+       This util has known shortcomings which may never be relevant and are thus left unaddressed.
+       The most prominent example is that of assigning an array- or a json(b)-typed value in a single column update,
+       which will fail if passed as a list/dict (as opposed to strings), due to how the query is currently designed.
```
aj-fuentes: All utils here can fail one way or another. It's OK that you thought about it. But I think your implementation is good for what we meant it to be. If somebody reports an extra corner case we can improve this further. IMHO this warning here can be removed, or kept in a simpler way (less detrimental to the implementation):
```suggestion

    .. warning::

       The values in the mapping will be casted to the type of the target column.
       This util was designed to update scalar values, avoid setting arrays or
       json data via the mapping.
```

KangOl: `This function is designed to ...`

src/util/pg.py
Pirols: > Why `ŧable`? Is it a table with two boards `ŧ` vs `t` :trollface:

You've got some eagle eyes there! :eyes:

src/util/pg.py
```diff
@@ -1621,3 +1622,83 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+          # single column update
+          util.bulk_update_table(
+              cr,
+              "account_move",
+              "always_tax_eligible",
+              {
+                  1: True,
+                  2: False,
+              },
+          )
```
aj-fuentes: The examples take a lot of real space in the docs. We can make it shorter
```suggestion
          util.bulk_update_table(cr, "res_users", "active", {1: False, 2: True})
```

src/util/pg.py
```diff
@@ -1621,3 +1622,83 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+          # multi-column update
+          util.bulk_update_table(
+              cr,
+              "account_move",
+              ["closing_return_id", "always_tax_eligible"],
+              {
+                  1: [2, True],
+                  2: [3, False],
+              },
+          )
```
aj-fuentes:
```suggestion
          util.bulk_update_table(
              cr,
              "res_users",
              ["active", "login"],
              {2: [True, "admin"], 6: [True, "demo"]},
          )
```

aj-fuentes (commented):
Some doc nits, we can make the example take less space?
image

src/util/pg.py
Pirols: Also embedded the usage of `key_col` in this one.

src/util/pg.py
```diff
@@ -1621,3 +1622,76 @@ def create_id_sequence(cr, table, set_as_default=True):
...
+          # multi-column update
+          util.bulk_update_table(
+              cr,
+              "res_users",
+              ["active", "login"],
+              {
+                  "administrator": [True, "admin"],
+                  "demouser": [True, "demo"],
+              },
+              key_col="login",
```
aj-fuentes: `key_col` is always the key of the mapping... unless I'm missing something.
```suggestion
                  2: [True, "admin"],
                  6: [True, "demo"],
              },
              key_col="id",
```

aj-fuentes: Maybe, if we want an example of non-id key:
```suggestion
              ["active", "password"],
              {
                  "admin": [True, "admin"],
                  "demo": [True, "demo"],
              },
              key_col="login",
```

Pirols: The intention is twofold: to showcase that you can use `key_col` to use any column as key. And that when you do, it is still possible to update the column of choice. Dislike the idea? Suggestions to improve its "implementation"?

aj-fuentes: I see. Although fine it's a bit confusing. In the example it's better to assume the existing value is what users get in a default install. The value set can then be different. 

IMO this blends too many options into a single example, and could be confusing. We should focus in the columns spec, that's the "weird" part of the function. I think it's fine to just put multi-column+explicit key column. Reflect this as well int the comment.

Nothing prevents updating the key column. Nothing implies that either from anything we said. IMO we don't need an example for it.

Pirols: Fair enough. :+1:

KangOl (approved):
@robodoo r+

PR #299, opened 2025-08-06
[FIX] util/misc: `split_osenv`, do not return `[""]`

aj-fuentes:
If the env var was not set we returned `[""]`. That was confusing.
It also breaks the boolean check of the result.

With this patch we just return an empty list when the variable is not set.

See also: odoo/upgrade#8174

Co-Authored-By: Christophe Simonis (chs) <chs@odoo.com>


KangOl (approved):
@robodoo r+

PR #302, opened 2025-08-19
[IMP] util.explode_query_range

KangOl:
Always return at least one query on CI. It ensure[^1] the give query will be executed and thus avoid failures when run in production.

[^1]: fine, we can't be sure the caller will actually execute them.


KangOl:

KangOl:
@robodoo r+

PR #303, opened 2025-08-21
[FIX] recompute_fields: get ids in batches

jjmaksoud:
If no ids are given to recompute, the util will fetch all ids in the target table and then recompute in chunks. Fetching all the ids itself can cause a memory error if the table is too large.
Using a named cursor with a limit of 1M records to fetch can eliminate this possibility.

opw-4909458
upg-3106002

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -14,7 +14,6 @@ import re
 from contextlib import contextmanager
 from functools import wraps
 from itertools import chain
-from operator import itemgetter
 from textwrap import dedent
 
 try:
@@ -43,7 +42,7 @@ from .const import BIG_TABLE_THRESHOLD
 from .exceptions import MigrationError
 from .helpers import table_of_model
 from .misc import chunks, log_progress, version_between, version_gte
-from .pg import column_exists, get_columns
+from .pg import column_exists, format_query, get_columns, named_cursor
 
 # python3 shims
 try:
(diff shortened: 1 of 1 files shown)


jjmaksoud:
upgradeci retry with always only account

src/util/orm.py
```diff
@@ -281,35 +281,47 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+        ncr = named_cursor(cr, MAX_SIZE)
+        ncr.execute(format_query(cr, "SELECT id FROM {t}", t=table_of_model(cr, model)))
+        res = ncr.fetchmany(MAX_SIZE)
+        while res:
+            yield tuple(map(itemgetter(0), res))
+            res = ncr.fetchmany(MAX_SIZE)
+        ncr.close()
+
+    for _ids in get_record_ids():
+        # will be checked with the first batch of ids only
```
aj-fuentes: I think we can do a `select count(id) ...` in advance instead of this.

src/util/orm.py
```diff
@@ -281,35 +281,47 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+        if ids:
+            yield ids
+            return
+        MAX_SIZE = 1000000
+        ncr = named_cursor(cr, MAX_SIZE)
+        ncr.execute(format_query(cr, "SELECT id FROM {t}", t=table_of_model(cr, model)))
+        res = ncr.fetchmany(MAX_SIZE)
+        while res:
+            yield tuple(map(itemgetter(0), res))
+            res = ncr.fetchmany(MAX_SIZE)
```
aj-fuentes: To keep the implementation the same we could hack only `ids`
```suggestion
    if ids is None:
        ncr = named_cursor(cr, itersize=1000000)
        ncr.execute(format_query(cr, "SELECT id FROM {t}", t=table_of_model(cr, model)))
        ids = (id_ for (id_,) in ncr)
```
Then the main loop `for subids in log_progress(chunks(ids, chunk_size, list), logger, qualifier=qual, size=size):` can be kept untouched?

aj-fuentes: As an extra I'd suggest we eliminate nondeterministic iteration by adding `ORDER BY`, basically `SELECT id FROM ... ORDER BY id`

KangOl:
This has the disadvantage of splitting the `log_progress`.

I would keep one loop
```python
if ids is None:
    cr.execute("SELECT COUNT(id) FROM ...")
    [count] = cr.fetchone()
else:
    count = len(ids)

if strategy == "auto":
    big_table = count > BIG_TABLE_THRESHOLD
    ...

size = (count + chunk_size - 1) / chunk_size

def get_ids():
    if ids is not None:
        for id_ in ids:
            yield id_
    MAX_SIZE = 1000000
    with named_cursor(cr, MAX_SIZE) as ncr:
        ncr.execute("SELECT id FROM ...")
        for (id_,) in ncr:
            yield id_

for subids in log_progress(chunks(get_ids(), chunk_size, list), logger, qualifier=qual, size=size):
    ...
```

src/util/orm.py
```diff
@@ -281,35 +281,47 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+        ncr.close()
+
+    for _ids in get_record_ids():
+        # will be checked with the first batch of ids only
+        if strategy == "auto":
+            big_table = len(_ids) > BIG_TABLE_THRESHOLD
+            any_tracked_field = any(getattr(Model._fields[f], _TRACKING_ATTR, False) for f in fields)
+            strategy = "commit" if big_table and any_tracked_field else "flush"
+
+        size = (len(_ids) + chunk_size - 1) / chunk_size
```
aj-fuentes: len(_ids) can be determined upfront with a `select count(id)...`

src/util/orm.py
```diff
@@ -281,35 +281,47 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
-        if strategy == "commit":
-            cr.commit()
-        else:
-            flush(records)
-        invalidate(records)
+    def get_record_ids():
+        if ids:
+            yield ids
+            return
+        MAX_SIZE = 1000000
```
cawo-odoo:
```suggestion
        MAX_SIZE = 2**20
```
For alignment with the default `chunk_size`?

jjmaksoud:
all comments applied

src/util/orm.py
```diff
@@ -281,18 +280,31 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
     assert strategy in {"flush", "commit", "auto"}
     Model = env(cr)[model] if isinstance(model, basestring) else model
     model = Model._name
+
+    def get_ids():
+        if ids is not None:
+            for id_ in ids:
+                yield id_
+        MAX_SIZE = 2**20
+        with named_cursor(cr, MAX_SIZE) as ncr:
```
aj-fuentes: You need a branch here, either use `return` or `else:`
nit: you can get rid of the MAX_SIZE var, otoh I find `itersize=` useful (explicit)
```suggestion
        else:
            with named_cursor(cr, itersize=2**20) as ncr:
```

aj-fuentes (approved):
LGTM

cawo-odoo (approved):

PR #304, opened 2025-08-22
[IMP] create_m2m: auto generate and return table name

Pirols:
In order to limit chances for typos, while keeping the util backwards compatible (except for the special `m2m` "auto" value), allow users to provide "auto" as a special flag to ask the util to automatically generate the m2m table name, following the same rule as the ORM[^1].

[^1]: Example from 18.0
https://github.com/odoo/odoo/blob/f42fe80576c8b419b5215ff17859e0dfebcd996a/odoo/fields.py#L4918-L4923

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1045,6 +1045,18 @@ class TestPG(UnitTestCase):
         self.assertTrue(s(ulc.using(alias="x")), ', "x"."a", "x"."A"')
         self.assertIs(ulc, ulc.using(leading_comma=True))
 
+    def test_create_m2m(self):
+        cr = self.env.cr
+
+        m2m_name = "random_table_name"
+        created_m2m = util.create_m2m(cr, m2m_name, "res_users", "res_groups")
+        self.assertEqual(m2m_name, created_m2m)
+        self.assertTrue(util.table_exists(cr, created_m2m))
+
+        auto_generated_m2m_table_name = util.create_m2m(cr, util.AUTO, "res_users", "res_groups")
+        self.assertEqual("res_groups_res_users_rel", auto_generated_m2m_table_name)
+        self.assertTrue(util.table_exists(cr, auto_generated_m2m_table_name))
+
 
 class TestORM(UnitTestCase):
     def test_create_cron(self):
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -44,7 +44,7 @@ from .domains import _adapt_one_domain, _replace_path, _valid_path_to, adapt_dom
 from .exceptions import SleepyDeveloperError
 from .helpers import _dashboard_actions, _validate_model, resolve_model_fields_path, table_of_model
 from .inherit import for_each_inherit
-from .misc import log_progress, safe_eval, version_gte
+from .misc import AUTO, log_progress, safe_eval, version_gte
 from .orm import env, invalidate
 from .pg import (
     SQLStr,
@@ -756,15 +756,12 @@ def convert_m2o_field_to_m2m(cr, model, field, new_name=None, m2m_table=None, co
     table1 = table_of_model(cr, model)
     table2, _, _ = target_of(cr, table1, field)
 
-    if m2m_table is None:
-        m2m_table = "{}_{}_rel".format(*sorted([table1, table2]))
-
     if col1 is None:
         col1 = "{}_id".format(table1)
     if col2 is None:
         col2 = "{}_id".format(table2)
 
-    create_m2m(cr, m2m_table, table1, table2, col1, col2)
+    m2m_table = create_m2m(cr, m2m_table or AUTO, table1, table2, col1, col2)
 
     dedup = SQLStr("ON CONFLICT DO NOTHING")
     if cr._cnx.server_version < 90500:
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -60,6 +60,9 @@ class Sentinel:
         return self.name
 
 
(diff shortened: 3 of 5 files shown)


aj-fuentes (approved):
LGTM

Thanks!

I'll let @KangOl double check this. Maybe we can do something better than then `"auto"` magic value. As we discussed in DM I don't see any other option that is also _clear_ and backwards compatible.

src/util/pg.py
```diff
@@ -1339,14 +1339,33 @@ def drop_depending_views(cr, table, column):
...
+    :param str col2: column referencing `fk2`, defaults to `"{fk2}_id"`
+    :return: the name of the table just created/fixed-up
+    :rtype: str
+    """
     if col1 is None:
         col1 = "%s_id" % fk1
     if col2 is None:
         col2 = "%s_id" % fk2
 
+    if m2m == "auto":
```
KangOl: I would reuse the `AUTOMATIC` sentinel (which should then be moved to `misc.py`)
```suggestion
    if m2m is AUTOMATIC:
```

src/util/pg.py
```diff
@@ -1339,14 +1339,33 @@ def drop_depending_views(cr, table, column):
...
+        m2m = "{}_{}_rel".format(*sorted([fk1, fk2]))
+
     if table_exists(cr, m2m):
         fixup_m2m(cr, m2m, fk1, fk2, col1, col2)
-        return
+        return m2m
 
     cr.execute(
         """
```
KangOl, line 1373: Maybe take the opportunity to improve the code and use `format_query`.

src/base/tests/test_util.py
```diff
@@ -1045,6 +1045,18 @@ def test_ColumnList(self):
...
 
+    def test_create_m2m(self):
+        cr = self.env.cr
+
+        m2m_name = "random_table_name"
+        created_m2m = util.create_m2m(cr, m2m_name, "res_users", "res_groups")
+        self.assertEqual(m2m_name, created_m2m)
+        self.assertTrue(util.table_exists(cr, created_m2m))
+
+        auto_generated_m2m_table_name = util.create_m2m(cr, "auto", "res_users", "res_groups")
```
```suggestion
        auto_generated_m2m_table_name = util.create_m2m(cr, util.AUTOMATIC, "res_users", "res_groups")
```

src/util/pg.py
Pirols: Nice, done.

src/base/tests/test_util.py
Pirols: :facepalm:

src/util/pg.py
```diff
@@ -1339,26 +1339,55 @@ def drop_depending_views(cr, table, column):
...
 def create_m2m(cr, m2m, fk1, fk2, col1=None, col2=None):
+    """
+    Ensure a m2m table exists or is created.
+
+    This function creates the table associated to a m2m field.
+    If the table already exists, :func:`~odoo.upgrade.util.pg.fixup_m2m` is run on it.
+    The table name can be generated automatically, applying the same logic of the ORM.
+    In order to do so, use the value "auto" for the `m2m` parameter.
+
+    :param str m2m: table name, if `AUTOMATIC` it is generated from `fk1` and `fk2`
```
```suggestion
    :param str m2m: table name, if :const:`~odoo.upgrade.util.misc.AUTOMATIC` it is generated from `fk1` and `fk2`
```

KangOl:
upgradeci retry with always only base helpdesk

KangOl:
> [!WARNING]  
> Some CI checks were not successful.

Can you please rebase all your branches to ensure they are up-to-date?

src/util/pg.py
```diff
@@ -1339,26 +1339,56 @@ def drop_depending_views(cr, table, column):
...
     cr.execute(
-        """
-        CREATE TABLE {m2m}(
-            {col1} integer NOT NULL REFERENCES {fk1}(id) ON DELETE CASCADE,
-            {col2} integer NOT NULL REFERENCES {fk2}(id) ON DELETE CASCADE,
-            PRIMARY KEY ({col1}, {col2})
-        );
-        CREATE INDEX ON {m2m}({col2}, {col1});
-    """.format(**locals())
+        format_query(
```
aj-fuentes: Note that `query = """...`
then `cr.execute(format_query(query...))` would keep the history better than indenting the whole thing.

src/util/fields.py
```diff
@@ -733,7 +733,7 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
         convert_field_to_html(cr, inh.model, field, skip_inherit=skip_inherit)
 
 
-def convert_m2o_field_to_m2m(cr, model, field, new_name=None, m2m_table=None, col1=None, col2=None):
+def convert_m2o_field_to_m2m(cr, model, field, new_name=None, m2m_table=AUTO, col1=None, col2=None):
```
KangOl: Not needed. Keep `None` as default value.

src/util/misc.py
```diff
@@ -60,6 +60,10 @@ def __repr__(self):
         return self.name
 
 
+AUTOMATIC = Sentinel("AUTOMATIC")
+AUTO = AUTOMATIC
```
```suggestion
AUTOMATIC = AUTO = Sentinel("AUTOMATIC")
```

KangOl: If we want to promote the usage of `AUTO`, we need to also change the name (for the doc)
```suggestion
AUTO = AUTOMATIC = Sentinel("AUTO")
```

KangOl (approved):
@robodoo r+

KangOl:
@robodoo rebase-ff

PR #305, opened 2025-08-22
[RFC][ADD] util/convert_m2o_to_m2m

Pirols:
TODO

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -1585,3 +1585,59 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
         ) % {"src_model": src_model, "dst_model": dst_model, "actions": ", ".join(action_names)}
 
         add_to_migration_reports(message=msg, category="Server Actions")
+
+
+def convert_m2o_to_m2m(cr, origin_model, m2o_field, m2m_field):
+    """
+    TODO
+    """
+    _validate_model(origin_model)
+
+    if not table_exists(cr, m2m_field):
+        raise SleepyDeveloperError("m2m table should already exist warning")
+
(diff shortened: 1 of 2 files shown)


src/util/fields.py
```diff
@@ -1585,3 +1585,59 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
...
+
+
+def convert_m2o_to_m2m(cr, origin_model, m2o_field, m2m_field):
+    """
+    TODO
+    """
+    _validate_model(origin_model)
+
+    if not table_exists(cr, m2m_field):
+        raise SleepyDeveloperError("m2m table should already exist warning")
```
aj-fuentes, line 1597: IMO this util should create the m2m the table (auto key?). 

Still I don't get what you are checking here. Why should there be a table with the name of the m2m_field? Maybe you wanted to check a _column_?

If we auto-create the m2m table, we will allow simpler code for fields that are introduced as m2o and later changed to m2m in the **same** major version --think 18.2 and 18.4 for example.

Pirols: Yes, you're right. I did not want to duplicate the logic in `create_m2m` and/or provide the same kind of arguments here. But it is necessary. I guess I can close this in favor of https://github.com/odoo/upgrade-util/pull/306

Pirols:
In favor of https://github.com/odoo/upgrade-util/pull/306

src/util/fields.py
aj-fuentes: I think you can just move your commit there... not sure if that's what you intend to do :)

EDIT (by Pirols): Communication hiccup solved in pm: this PR is not #304 and #306 is not #305.

PR #306, opened 2025-08-22
[ADD] util.convert_many2one_field_to_many2many

KangOl:
TODO:
 - write doc
 - tests


KangOl:
upgradeci skip

KangOl:

src/util/fields.py
```diff
@@ -731,6 +733,79 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+
+    if col1 is None:
+        col1 = "{}_id".format(table1)
+    if col2 is None:
+        col2 = "{}_id".format(table2)
+
+    create_m2m(cr, m2m_table, table1, table2, col1, col2)
+
+    dedup = SQLStr("ON CONFLICT DO NOTHING")
+    if cr._cnx.server_version < 90500:
```
aj-fuentes, line 770: We probably never going to need this (minimal PG for valid targets it's been 10 for a while) but I guess it doesn't hurt.

src/util/fields.py
```diff
@@ -731,6 +733,79 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+        )
+
+    fill_query = format_query(
+        cr,
+        """
+            INSERT INTO {m2m}({col1}, {col2})
+                 SELECT id, {field}
+                   FROM {table}
+                  WHERE {field} IS NOT NULL
+                {dedup}
```
aj-fuentes, line 791: ~~Unless the the original table is really broken I don't see how we can get a duplicated `id`s. IMO we can let the upgrade fail in such cases since anyway the source table is broken.~~

EDIT: well maybe the table already exists (from DM)

src/util/pg.py
```diff
@@ -1339,14 +1340,33 @@ def drop_depending_views(cr, table, column):
...
+    :return: the name of the table just created/fixed-up
+    :rtype: str
+    """
     if col1 is None:
         col1 = "%s_id" % fk1
     if col2 is None:
         col2 = "%s_id" % fk2
 
+    if m2m == "auto":
+        m2m = _get_default_m2m_table_name(fk1, fk2)
```
KangOl: I don't really like this. I would prefer to keep this argument explicit for this low-level function.

This doesn't look necessary now that we have the `convert_many2one_field_to_many2many` function.

src/util/fields.py
```diff
@@ -733,6 +733,10 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
         convert_field_to_html(cr, inh.model, field, skip_inherit=skip_inherit)
 
 
+def _get_default_m2m_table_name(table1, table2):
```
KangOl: Do we really need such a helper?
If we do, it should be defined in `pg.py`.

Pirols: In fact, we don't, I just realized that if we merge #304, we can just use the new `auto` value for `m2m`.

src/util/fields.py
KangOl: Indeed, but it's to be consistent with 

https://github.com/odoo/upgrade-util/blob/eadd4dc45ef6fbea97002585f9d2920e87677434/src/util/pg.py#L904-L908

src/util/fields.py
```diff
@@ -731,6 +733,83 @@ def convert_field_to_html(cr, model, field, skip_inherit=()):
...
+    return "{}_{}_rel".format(*sorted([table1, table2]))
+
+
+def convert_many2one_field_to_many2many(cr, model, field, new_name=None, m2m_table=None, col1=None, col2=None):
```
Pirols:
```suggestion
def convert_m2o_field_to_m2m(cr, model, field, new_name=None, m2m_table=None, col1=None, col2=None):
```

In all other utils (create_m2m, fixup_m2m and a dozen more) we use the shorthand form (`m2{m,o}` instead of `many2{many,one}`, shouldn't we keep it consistent here as well?

aj-fuentes:
I like https://github.com/odoo/upgrade-util/pull/304 and the "auto" solution (as an independet commit.) I think it's useful: one less thing to typo-check. It's easy to miss those long table names, and if devs just copy and paste it for the insert query the typo can still be there.

I also think that the solution here was fine. It doesn't need the extra fixups. IMO we don't need `_get_default_m2m_table_name` since it's a simple code anyway and we can just inline it in both places.

In short, I'd vote for merge both PRs as they were.

src/util/fields.py
aj-fuentes: I understood the intent. But even the part you linked is obsolete now IMO. (No need to remove it, but IMO the chances of actually relying on it are minimal today). 

Unless we rewrite older scripts to use the new util here, I see close to zero chances of this code being called on top of an older PG version.

Pirols:
> In short, I'd vote for merge both PRs as they were.

I agree, if anything I'd use the `"auto"` value in the new util. :)

src/util/fields.py
KangOl: You're right.

KangOl:
@robodoo r+

PR #309, opened 2025-08-27
[FIX] orm: patch manual field names

jjmaksoud:
The check for manual field names moved from being a model method to a util function. It needs to be patched starting version `saas~18.4` to allow loading custom module fields as manual for tests during the upgrade process.

Should be merged only after https://github.com/odoo/odoo/pull/224524 and https://github.com/odoo/enterprise/pull/93335

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -726,8 +726,11 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
 
         patches.append(patch.object(BaseModel, "_add_magic_fields", _add_magic_fields))
 
-    if version_gte("saas~16.4"):
-        # 3.5.3 allow loading manual fields
+    # 3.5.3 allow loading manual fields
+    if version_gte("saas~18.4"):
+        patches.append(patch("odoo.fields.is_manual_field_name", lambda name: True))
+        patches.append(patch("odoo.orm.model_classes.is_manual_field_name", lambda name: True))
+    elif version_gte("saas~16.4"):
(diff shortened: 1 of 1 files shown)


src/util/orm.py
```diff
@@ -726,10 +726,15 @@ def _add_magic_fields(self):
 
         patches.append(patch.object(BaseModel, "_add_magic_fields", _add_magic_fields))
 
-    if version_gte("saas~16.4"):
-        # 3.5.3 allow loading manual fields
+    # 3.5.3 allow loading manual fields
+
+    if version_between("saas~16.4", "saas~18.4"):
```
aj-fuentes: This includes 18.4, I'd suggest you invert the checks
```
if version >= 18.4:
  ...
elif version >= 16.4:
  ...
```
Note that there are some errors in the CI.

jjmaksoud: The ci errors are expected as I did not bundle the branches for runbot, and matt is running upgrades to master where the fw is not created

aj-fuentes: OK, we will need some dance to merge the PRs in the right order.
cc: @KangOl

jjmaksoud: community should be deployed first, even if we dont merge a util fix for a while the outcome will be as it is now. But if we merge a patch before the community deployment we will get the error that is currently in the ci

KangOl:

src/util/orm.py
aj-fuentes: Do not forget to put the check for the versions in the right order.
https://github.com/odoo/upgrade-util/blob/5efae69424301785b96c451da7b26132e93f9a51/src/util/misc.py#L123

PR #310, opened 2025-08-29
[FIX] util/pg: drop indexes when removing constraints

aj-fuentes:
If a constraint with the same name was found. Attempt as well to drop
any existing index with the same name. Otherwise we may leave behind
constraints implemented as unique indexes.

See odoo/upgrade/8318


aj-fuentes:
upgradeci retry with always base account_edi_proxy_client

KangOl:
upgradeci retry with always only base account_edi_proxy_client

KangOl:
And it also won't work for partial indexes.

PR #312, opened 2025-09-03
[FIX] util/report: Make 'name' argument optional

depr-odoo:
This [commit](https://github.com/odoo/upgrade-util/pull/215) ensure that if the name argument is not provided, it will automatically generate a default value for name based on the model and id.

When we call method get_anchor_link_to_record without name, it fails with

```
TypeError: get_anchor_link_to_record() missing 1 required positional argument: 'name'
```

Updated get_anchor_link_to_record to default 'name' to None if not provided.

Also added a check to ensure 'id' is not a tuple before proceeding with its [use](https://github.com/odoo/upgrade-util/blob/master/src/util/report.py#L328-L330).


src/util/report.py
```diff
@@ -319,9 +319,11 @@ def ref(xid):
                 _logger.warning("Cannot unfold chat window", exc_info=True)
 
 
-def get_anchor_link_to_record(model, id, name, action_id=None):
+def get_anchor_link_to_record(model, id, name=None, action_id=None):
```
Pirols, line 322:
```suggestion
def get_anchor_link_to_record(model, id, name, action_id=None):
```

We handle cases when `name` turns out to be Falsey, but I don't think we want the argument to be optional.

cc: @aj-fuentes

src/util/report.py
```diff
@@ -319,9 +319,11 @@ def ref(xid):
...
     _validate_model(model)
     if not name:
+        if isinstance(id, tuple):
```
Pirols: **If** we want this, we should also check for lists:

```suggestion
        if isinstance(id, (tuple, list)):
```

Pirols (commented):
Left a couple of comments. Still, I think that providing a list/tuple of ids is a misuse of the function. So I'd expect the error to need correcting on the user-side. Otherwise, can you refer to where you encountered a relevant issue?

src/util/report.py
aj-fuentes: Ah, yes, indeed. I think that was the intent. But I think we can add the default to None... I don't see any downside. @KangOl wdyt?

src/util/report.py
aj-fuentes: I think this is not what we want. `id` must be an int.

KangOl: Why this change?

aj-fuentes: @depr-odoo remove this please.

depr-odoo: Hello @aj-fuentes,
it's done

aj-fuentes (approved):
LGTM

cc: @KangOl

PR #313, opened 2025-09-04
[IMP] util/fields: Convert to company dependent without company field

filisbits:
The current implementation of the function `_convert_field_to_company_dependent` requires a field that stores the company value, but this field is not always available.

This change aims to allow that case by passing a falsy value as company_field, then it creates a json with the current value for all companies as the new value in the column

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1156,6 +1156,65 @@ class TestField(UnitTestCase):
 
         self.assertEqual(new_default, "en_US")
 
+    @unittest.skipIf(not util.version_gte("saas~17.5"), "Company dependent fields are stored as jsonb since saas~17.5")
+    def test_convert_field_to_company_dependent(self):
+        cr = self.env.cr
+
+        partner_model = self.env["ir.model"].search([("model", "=", "res.partner")])
+        self.env["ir.model.fields"].create(
+            [
+                {
+                    "name": "x_test_cd_1",
+                    "ttype": "char",
+                    "model_id": partner_model.id,
+                },
+                {
+                    "name": "x_test_cd_2",
+                    "ttype": "char",
+                    "model_id": partner_model.id,
+                },
+            ]
+        )
+
+        c1 = self.env["res.company"].create({"name": "Flancrest"})
+        c2 = self.env["res.company"].create({"name": "Flancrest2"})
+
+        test_partners = self.env["res.partner"].create(
+            [
+                {"name": "Homer", "x_test_cd_1": "A", "x_test_cd_2": "A", "company_id": c1.id},
+                {"name": "Marjorie", "x_test_cd_1": "B", "x_test_cd_2": "B"},
+                {"name": "Bartholomew"},
+            ]
+        )
+        test_partners.invalidate_recordset(["x_test_cd_1", "x_test_cd_2"])
+
+        # Using company_id as default, only records with company set are updated
+        util.make_field_company_dependent(cr, "res.partner", "x_test_cd_1", "char")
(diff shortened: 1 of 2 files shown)


src/util/fields.py
```diff
@@ -836,9 +836,17 @@ def _convert_field_to_company_dependent(
...
-    where_condition += format_query(cr, " AND {} IS NOT NULL", company_field)
 
-    using = format_query(cr, "jsonb_build_object({}, {{0}})", company_field)
+    if company_field:
+        where_condition += format_query(cr, " AND {} IS NOT NULL", company_field)
+        using = format_query(cr, "jsonb_build_object({}, {{0}})", company_field)
+    else:
+        using = format_query(
+            cr,
+            "(SELECT jsonb_object_agg(c.id, x.{{0}}) FROM {} AS x, res_company AS c WHERE x.id = {}.id GROUP BY x.id)",
```
aj-fuentes: Can't we just select from company directly?
```suggestion
            "(SELECT jsonb_object_agg(c.id::text, x.{{0}}) FROM res_company c)",
```

aj-fuentes:
I rebased and pushed a fixup for the test to avoid reading the whole JSON data back

src/base/tests/test_util.py
```diff
@@ -1156,6 +1156,61 @@ def test_change_field_selection_with_default(self):
...
+            ]
+        )
+        test_partners.invalidate_recordset(["x_test_cd_1", "x_test_cd_2"])
+
+        # Using company_id as default, only records with company set are updated
+        util.make_field_company_dependent(cr, "res.partner", "x_test_cd_1", "char")
+        util.make_field_company_dependent(cr, "res.partner", "x_test_cd_2", "char", company_field=False)
+
+        # make the ORM re-read the info about these manual fields from the DB
+        self.registry._setup_models__(self.env.cr, ["res.partner"])
```
aj-fuentes: With the help of @HydrionBurst this was the missing part to make the ORM reload the manual fields with the right metadata.

src/util/fields.py
```diff
@@ -881,9 +881,16 @@ def _convert_field_to_company_dependent(
...
-    using = format_query(cr, "jsonb_build_object({}, {{0}})", company_field)
+    if company_field:
+        where_condition += format_query(cr, " AND {} IS NOT NULL", company_field)
+        using = format_query(cr, "jsonb_build_object({}, {{0}})", company_field)
+    else:
+        using = format_query(
+            cr,
+            "(SELECT jsonb_object_agg(c.id::text, {}.{{0}}) FROM res_company c)",
+            table,
+        )
```
HydrionBurst, line 893: The company dependent field has a btree not null index, since it optimizes for spasing columns.
So if the original value is equal to the fallback, we shouldn't build the jsonb data.

src/base/tests/test_util.py
```diff
@@ -1156,6 +1156,64 @@ def test_change_field_selection_with_default(self):
...
+        )
+        test_partners.invalidate_recordset(["x_test_cd_1", "x_test_cd_2"])
+
+        # Using company_id as default, only records with company set are updated
+        util.make_field_company_dependent(cr, "res.partner", "x_test_cd_1", "char")
+        util.make_field_company_dependent(cr, "res.partner", "x_test_cd_2", "char", company_field=False)
+
+        # make the ORM re-read the info about these manual fields from the DB
+        if hasattr(self.registry, "_setup_models__"):
+            self.registry._setup_models__(self.env.cr)
```
```suggestion
            args = (["res.partner"],) if util.version_gte("saas~18.4") else ()
            self.registry._setup_models__(self.env.cr, *args)
```

aj-fuentes: Done, a variant of this...

KangOl (approved):
@robodoo r+

PR #315, opened 2025-09-09
[IMP] util/modules.py

KangOl:


src/util/modules.py
```diff
@@ -395,10 +409,12 @@ def merge_module(cr, old, into, update_dependers=True):
     if old not in mod_ids:
         # this can happen in case of temp modules added after a release if the database does not
         # know about this module, i.e: account_full_reconcile in 9.0
-        # `into` should be know. Let it crash if not
         _logger.log(NEARLYWARN, "Unknown module %s. Skip merge into %s.", old, into)
         return
 
+    if into not in mod_ids:
+        raise UnknownModuleError(into)
+
```
aj-fuentes, line 432: Don't we need to update `old` from `ENVIRON["__modules_auto_discovery_force_installs"]` in `merge_module` as well?
I'm afraid `_trigger_auto_discovery` will raise `UnknownModuleError` otherwise.

src/util/modules.py
```diff
@@ -510,6 +540,18 @@ def _up(table, old, new):
...
+        if version_gte("saas~14.5"):
+            # We must delay until the modules actually exists. They are added by the auto discovery process.
+            ENVIRON["__modules_auto_discovery_force_installs"].add(module)
+            return None
+        else:
+            raise
+
+
+def _force_install_module(cr, module, if_installed=None, reason="it has been explicitly asked for"):
     """
```
aj-fuentes: We need to move the documentation to the public function.

src/util/modules.py
```diff
@@ -518,9 +560,18 @@ def force_install_module(cr, module, if_installed=None, reason="it has been expl
                                            already installed
     :return str: the *new* state of the module
     """
```
aj-fuentes:
```
"""Strict implementation of force_install: raise errors for missing modules"""
```

src/util/modules.py
```diff
@@ -392,13 +406,29 @@ def merge_module(cr, old, into, update_dependers=True):
...
+    fi = ENVIRON["__modules_auto_discovery_force_installs"]
+    if old in fi:
+        fi.remove(old)
+        fi.add(into)
+    fu = ENVIRON["__modules_auto_discovery_force_upgrades"]
+    if old in fu:
+        if into not in fu:
+            fu[into] = fu.pop(old)
+        else:
+            init = fu[old][0] or fu[into][0]
```
aj-fuentes: since we use indexing, a comment helps:
```suggestion
            init = fu[old][0] or fu[into][0]  # keep init flag
```

src/util/modules.py
```diff
@@ -392,13 +406,29 @@ def merge_module(cr, old, into, update_dependers=True):
...
+            version = fu[old][1] if parse_version(fu[old][1]) < parse_version(fu[into][1]) else fu[into][1]
```
aj-fuentes, line 420:
```suggestion
            # keep the min version, this controls which upgrade scripts run
            version = fu[old][1] if parse_version(fu[old][1]) < parse_version(fu[into][1]) else fu[into][1]
```

src/util/modules.py
```diff
@@ -518,9 +549,31 @@ def force_install_module(cr, module, if_installed=None, reason="it has been expl
...
     subquery = ""
     params = [module]
     if if_installed:
+        try:
+            _assert_modules_exists(cr, *if_installed)
+        except UnknownModuleError as e:
+            # Warn about unknown modules (can help track typos in the call).
+            # we don't care about missing modules, as by definition, they are not installed.
+            _logger.log(
+                NEARLYWARN, "Unknown 'if_installed' modules: %s; consider them as uninstalled.", ", ".join(e.args)
```
aj-fuentes, line 575: Shouldn't we block the CI on this?

KangOl: I don't think so. We may have cases where such modules are added in saas versions and thus not available in the .0 version. Or modules added after the release.

PR #316, opened 2025-09-10
[IMP] util.change_field_selection_values

aj-fuentes:
Handle company_dependent fields.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -41,7 +41,7 @@ except ImportError:
 from . import json
 from .const import ENVIRON
 from .domains import _adapt_one_domain, _replace_path, _valid_path_to, adapt_domains
-from .exceptions import SleepyDeveloperError
+from .exceptions import SleepyDeveloperError, UpgradeError
 from .helpers import _dashboard_actions, _validate_model, resolve_model_fields_path, table_of_model
 from .inherit import for_each_inherit
 from .misc import AUTO, log_progress, safe_eval, version_gte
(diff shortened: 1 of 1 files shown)


src/util/fields.py
```diff
@@ -1122,13 +1122,62 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
+        data = {
+            "keys": tuple(mapping),
+            "json": json.dumps(mapping),
+            "match": "exists($.* ? (@ like_regex {})".format(str(Json(like_regex))[1:-1]),
+        }
         queries = [
-            cr.mogrify(q, [json.dumps(mapping), tuple(mapping)]).decode()
-            for q in explode_query_range(cr, query, table=table)
+            cr.mogrify(q, data).decode()
+            for q in explode_query_range(cr, format_query(cr, query, table=table, column=field), table=table)
```
```suggestion
            for q in explode_query_range(cr, format_query(cr, query, table=table, column=field), table=table, alias="t")
```

src/util/fields.py
```diff
@@ -1122,13 +1122,62 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
+        if ctype == "varchar":
+            query = "UPDATE {table} t SET {column} = %(json)s::jsonb->>t.{column} WHERE t.{column} IN %(keys)s"
+        elif ctype == "jsonb":
+            # company dependent selection field
+            query = """
+                UPDATE {table}
+                   SET
+                    WITH upd AS (
+                         SELECT t.id,
+                                jsonb_object_agg(v.key, %s::jsonb->>{field}) AS value
```
```suggestion
                                jsonb_object_agg(v.key, %(json)s::jsonb->>v.value) AS value
```
Actually I think what you need to do here is to replace values by the ones in the mapping.

src/util/fields.py
```diff
@@ -1122,13 +1122,62 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
     if column_exists(cr, table, field):
-        query = "UPDATE {table} SET {field}= %s::jsonb->>{field} WHERE {field} IN %s".format(table=table, field=field)
+        ctype = column_type(cr, table, field)
+        if ctype == "varchar":
+            query = "UPDATE {table} t SET {column} = %(json)s::jsonb->>t.{column} WHERE t.{column} IN %(keys)s"
+        elif ctype == "jsonb":
+            # company dependent selection field
+            query = """
+                UPDATE {table}
+                   SET
```
aj-fuentes: typo?
```suggestion
```

src/util/fields.py
```diff
@@ -1122,13 +1122,62 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
+                    WITH upd AS (
+                         SELECT t.id,
+                                jsonb_object_agg(v.key, %s::jsonb->>{field}) AS value
+                           FROM {table} t
+                           JOIN LATERAL jsonb_each_text(t.{column}) v
+                             ON true
```
aj-fuentes: This should be equivalent
```suggestion
                           FROM {table} t,
                                jsonb_each_text(t.{column}) v
```

src/util/fields.py
```diff
@@ -1122,13 +1122,62 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
+                          WHERE jsonb_path_match(t.{column}, %(match)s)
```
aj-fuentes: Can't we just do a simple check with keys?
```suggestion
                          WHERE v.value IN %(keys)s
```

KangOl:
upgradeci retry with always only stock_account

src/util/fields.py
KangOl: No, we need to keep other values. This (like the new version) is to filter out the records that should not be modified.

src/util/fields.py
```diff
@@ -1122,13 +1122,58 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
+        if ctype == "varchar":
+            query = "UPDATE {table} t SET {column} = %(json)s::jsonb->>t.{column} WHERE t.{column} = ANY(%(keys)s)"
+        elif ctype == "jsonb":
+            # company dependent selection field
+            query = """
+                WITH upd AS (
+                     SELECT t.id,
+                            jsonb_object_agg(v.key, COALESCE(%(json)s::jsonb->>v.value, v.value)) AS value
+                       FROM {table} t
+                       JOIN LATERAL jsonb_each_text(t.{column}) v
```
aj-fuentes, line 1135: > LATERAL can also precede a function-call FROM item, but in this case it is a noise word, because the function expression can refer to earlier FROM items in any case.

From PG [docs](https://www.postgresql.org/docs/current/sql-select.html#SQL-FROM)

PR #317, opened 2025-09-10
[IMP] util/orm: `recompute_fields` add query parameter

aj-fuentes:
In most cases when we pass a list of `ids` to `recompute_fields` we just run a query to get them. The main downside of this approach is when the list is too big: it can cause MemoryErrors. Instead we can let `recompute_fields` use a query to get the ids and take the necessary precautions to not fetch millions of entries all at once. As a nice side effect code becomes easier to review since we don't need to check the fetching of ids is done correctly.

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -251,7 +251,7 @@ def no_selection_cache_validation(f=None):
 
 
 @no_selection_cache_validation
-def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256, strategy="auto"):
+def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256, strategy="auto", query=None):
     """
     Recompute field values.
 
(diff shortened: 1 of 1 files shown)


aj-fuentes:
upgradeci skip

aj-fuentes:
upgradeci retry with always only base account crm delivery helpdesk_sale_timesheet helpdesk_timesheet hr_attendance hr_expense hr_holidays hr_recruitment_extract l10n_in_sale l10n_it_edi l10n_mx_edi_stock loyalty mrp point_of_sale pos_sale project_enterprise purchase resource sale sale_project_forecast sale_renting sale_stock sale_subscription sale_timesheet_margin stock

src/util/orm.py
```diff
@@ -271,40 +271,52 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
-    def get_ids():
-        if ids is not None:
-            for id_ in ids:
-                yield id_
+    if ids is None:
+        query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
+        cr.execute(query)
+        count = cr.rowcount
+        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
+            ids_ = [id_ for (id_,) in cr.fetchall()]
```
```suggestion
            ids_ = (id_ for (id_,) in cr.fetchall())
```

src/util/orm.py
```diff
@@ -271,40 +271,52 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
     :param logger: logger used to report the progress
     :type logger: :class:`logging.Logger`
     :param int chunk_size: number of records per chunk - used to split the processing
     :param str strategy: strategy used to process the re-computation
+    :param str query: query to get the IDs of records to recompute, it is an error to set
+                      both `ids` and `query`
     """
-    assert strategy in {"flush", "commit", "auto"}
+    assert strategy in {"flush", "commit", "auto"}, "Invalid strategy {!r}".format(strategy)
+    assert ids is None or query is None, "Cannot set both `ids` and `query`"
```
```suggestion
    if strategy not in {"flush", "commit", "auto"}:
        raise ValueError("Invalid strategy {!r}".format(strategy))
    if ids is not None and query is not None:
        raise ValueError("Cannot set both `ids` and `query`")
```

src/util/orm.py
```diff
@@ -271,40 +271,54 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
 
-    def get_ids():
-        if ids is not None:
-            for id_ in ids:
-                yield id_
+    if ids is None:
+        query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
+        cr.execute(query)
+        count = cr.rowcount
+        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
```
jjmaksoud, line 294: at least 2 full chunks?

aj-fuentes: Yes. That's intended. We could say 1.1x, or 1.5x, I believe it doesn't matter. I chose 2x because it's simpler to express. In practice we haven't seen any issue with this for less than 10 million entries.

PR #320, opened 2025-09-12
[IMP] orm: parallel iter_browse

cawo-odoo:
###  [IMP] orm: add optional parallelism to iter_browse.__attr__()

In some cases, e.g. if it is known that calling a certain method on the model will only trigger inserts or it is clear that updates will be disjunct, such method calls can be done in parallel.

###  [IMP] orm: add optional parallelism to iter_browse.create()

Like the same support added to `__attr__` in the parent commit, this can only be used by callers when it is known that database modifications will be distinct, not causing concurrency issues or side-effects on the results.

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -9,12 +9,21 @@ cases totally different alternatives to the ORM's own functions are provided. Th
 on this module work along the ORM of *all* supported versions.
 """
 
+import collections
 import logging
+import multiprocessing
+import os
 import re
+import sys
 from contextlib import contextmanager
 from functools import wraps
-from itertools import chain
+from itertools import chain, repeat
 from textwrap import dedent
+
+try:
+    from concurrent.futures import ProcessPoolExecutor
+except ImportError:
+    ProcessPoolExecutor = None
 
 try:
     from unittest.mock import patch
@@ -27,9 +36,9 @@ try:
     except ImportError:
         from odoo import SUPERUSER_ID
     from odoo import fields as ofields
-    from odoo import modules, release
+    from odoo import modules, release, sql_db
 except ImportError:
-    from openerp import SUPERUSER_ID, modules, release
+    from openerp import SUPERUSER_ID, modules, release, sql_db
 
     try:
         from openerp import fields as ofields
@@ -41,8 +50,8 @@ except ImportError:
 from .const import BIG_TABLE_THRESHOLD
 from .exceptions import MigrationError
 from .helpers import table_of_model
-from .misc import chunks, log_progress, version_between, version_gte
-from .pg import SQLStr, column_exists, format_query, get_columns, named_cursor, query_ids
+from .misc import chunks, log_progress, str2bool, version_between, version_gte
+from .pg import SQLStr, column_exists, format_query, get_columns, get_max_workers, named_cursor, query_ids
 
 # python3 shims
 try:
@@ -51,6 +60,10 @@ except NameError:
     basestring = str
 
 _logger = logging.getLogger(__name__)
+
+UPG_PARALLEL_ITER_BROWSE = str2bool(os.environ.get("UPG_PARALLEL_ITER_BROWSE", "0"))
+# FIXME: for CI! Remove before merge
+UPG_PARALLEL_ITER_BROWSE = True
 
 
 def env(cr):
@@ -348,6 +361,21 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
         invalidate(records)
 
 
+def _mp_iter_browse_cb(ids_or_values, params):
+    me = _mp_iter_browse_cb
+    # init upon first call. Done here instead of initializer callback, because py3.6 doesn't have it
+    if not hasattr(me, "env"):
+        sql_db._Pool = None  # children cannot borrow from copies of the same pool, it will cause protocol error
+        me.env = env(sql_db.db_connect(params["dbname"]).cursor())
+        me.env.clear()
+    # process
+    if params["mode"] == "browse":
+        getattr(
+            me.env[params["model_name"]].with_context(params["context"]).browse(ids_or_values), params["attr_name"]
+        )(*params["args"], **params["kwargs"])
+    me.env.cr.commit()
+
+
 class iter_browse(object):
     """
     Iterate over recordsets.
@@ -413,7 +441,25 @@ class iter_browse(object):
         self._yield_chunks = kw.pop("yield_chunks", False)
         self._logger = kw.pop("logger", _logger)
         self._strategy = kw.pop("strategy", "flush")
-        assert self._strategy in {"flush", "commit"}
+        assert self._strategy in {"flush", "commit", "multiprocessing"}
+        if self._strategy == "multiprocessing":
+            if not ProcessPoolExecutor:
+                raise ValueError("multiprocessing strategy can not be used in scripts run by python2")
+            if not UPG_PARALLEL_ITER_BROWSE:
+                self._strategy = "commit"  # downgrade
+                if self._size > 100000:
+                    _logger.warning(
+                        "Browsing %d %s, which may take a long time. "
+                        "This can be sped up by setting the env variable UPG_PARALLEL_ITER_BROWSE to 1. "
+                        "If you do, be sure to examine the results carefully.",
+                        self._size,
+                        self._model._name,
+                    )
+                else:
+                    _logger.info(
+                        "Caller requested multiprocessing strategy, but UPG_PARALLEL_ITER_BROWSE env var is not set. "
+                        "Downgrading strategy to commit.",
+                    )
         if kw:
             raise TypeError("Unknown arguments: %s" % ", ".join(kw))
 
@@ -442,7 +488,7 @@ class iter_browse(object):
         return self._model.browse(*args)
 
     def _end(self):
-        if self._strategy == "commit":
+        if self._strategy in ["commit", "multiprocessing"]:
             self._model.env.cr.commit()
         else:
             flush(self._model)
@@ -485,8 +531,41 @@ class iter_browse(object):
             args = self._cr_uid + args
             return [getattr(chnk, attr)(*args, **kwargs) for chnk in chain(it, self._end())]
 
+        def caller_multiprocessing(*args, **kwargs):
+            params = {
+                "dbname": self._model.env.cr.dbname,
+                "model_name": self._model._name,
+                # convert to dict for pickle. Will still break if any value in the context is not pickleable
+                "context": dict(self._model.env.context),
+                "attr_name": attr,
+                "args": self._cr_uid + args,
+                "kwargs": kwargs,
+                "mode": "browse",
+            }
+            self._model.env.cr.commit()
+            extrakwargs = {"mp_context": multiprocessing.get_context("fork")} if sys.version_info >= (3, 7) else {}
+            with ProcessPoolExecutor(max_workers=get_max_workers(), **extrakwargs) as executor:
+                batch_len = min(get_max_workers() * 10, 1000000 // self._chunk_size)
+                ipc_chunksize = (batch_len + get_max_workers() - 1) // get_max_workers()
+                batch = []
+
+                def submit_batch():
+                    collections.deque(
+                        executor.map(_mp_iter_browse_cb, batch, repeat(params), chunksize=ipc_chunksize), maxlen=0
+                    )
+                    del batch[:]
+
+                for chunk in it:
+                    batch.append(chunk.ids)
+                    if len(batch) % batch_len == 0:
+                        submit_batch()
+                submit_batch()
+            next(self._end(), None)
+            # do not return results in // mode, we expect it to be used for huge numbers of
+            # records and thus would risk MemoryError, also we cannot know if what attr returns is pickleable
+
         self._it = None
-        return caller
+        return caller_multiprocessing if self._strategy == "multiprocessing" else caller
 
     def _values_iter(self, query):
         cr = self._model.env.cr
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -373,7 +373,12 @@ def _mp_iter_browse_cb(ids_or_values, params):
         getattr(
             me.env[params["model_name"]].with_context(params["context"]).browse(ids_or_values), params["attr_name"]
         )(*params["args"], **params["kwargs"])
+    if params["mode"] == "create":
+        new_ids = me.env[params["model_name"]].with_context(params["context"]).create(ids_or_values).ids
     me.env.cr.commit()
+    if params["mode"] == "create":
+        return new_ids
+    return None
 
 
 class iter_browse(object):
@@ -615,6 +620,12 @@ class iter_browse(object):
             qualifier = "env[%r].create([:%d])" % (self._model._name, self._chunk_size)
             it = log_progress(it, self._logger, qualifier=qualifier, size=sz)
 
+        if self._strategy == "multiprocessing":
+            return self._create_multiprocessing(it, size, multi)
+
+        return self._create(it, multi)
+
+    def _create(self, it, multi):
         ids = []
         self._patch = no_selection_cache_validation()
         for sub_values in it:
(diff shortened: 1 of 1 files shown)


src/util/orm.py
```diff
@@ -384,18 +414,48 @@ class iter_browse(object):
...
+            try:
+                self._size = len(ids)
+            except TypeError:
+                raise ValueError("When passing ids as a generator, the size kwarg is mandatory")
         self._chunk_size = kw.pop("chunk_size", 200)  # keyword-only argument
+        self._multiprocessing = kw.pop("multiprocessing", False)
         self._logger = kw.pop("logger", _logger)
-        self._strategy = kw.pop("strategy", "flush")
+        self._strategy = kw.pop("strategy", "commit" if self._multiprocessing and UPG_PARALLEL_ITER_BROWSE else "flush")
         assert self._strategy in {"flush", "commit"}
```
KangOl: What about merging the `multiprocessing` option to the `strategy` argument?
```suggestion
        assert self._strategy in {"flush", "commit", "multiprocessing"}
```

cawo-odoo: This would obviously be more elegant. :-)

src/util/orm.py
```diff
@@ -338,6 +343,31 @@ def get_ids():
...
+    if _mp_iter_browse_cb.mode == "browse":
+        getattr(
+            _mp_iter_browse_cb.env[_mp_iter_browse_cb.model_name].browse(ids_or_values), _mp_iter_browse_cb.attr_name
+        )(*_mp_iter_browse_cb.args, **_mp_iter_browse_cb.kwargs)
+    if _mp_iter_browse_cb.mode == "create":
+        _mp_iter_browse_cb.env[_mp_iter_browse_cb.model_name].create(ids_or_values)
+    _mp_iter_browse_cb.env.cr.commit()
+
+
+def _init_mp_iter_browse_cb(dbname, model_name, attr_name, args, kwargs, mode):
```
aj-fuentes: This could be internal to iter_browse, inline even.
To avoid any kind of leak from one run to another, I'd suggest a cleanup that removes all values set here in `_end` of iter_browse. The same for `create` once implemented.
I'd do something like `_init_mp_iter_browse_cb.runtime_data = (dbname, ...)` to centralize the set/unset to one single attribute.

src/util/orm.py
```diff
@@ -384,18 +414,48 @@ class iter_browse(object):
...
         self._chunk_size = kw.pop("chunk_size", 200)  # keyword-only argument
+        self._multiprocessing = kw.pop("multiprocessing", False)
         self._logger = kw.pop("logger", _logger)
-        self._strategy = kw.pop("strategy", "flush")
+        self._strategy = kw.pop("strategy", "commit" if self._multiprocessing and UPG_PARALLEL_ITER_BROWSE else "flush")
         assert self._strategy in {"flush", "commit"}
+        if self._multiprocessing:
+            if self._strategy == "flush":
+                raise ValueError("With multiprocessing, strategy must be 'commit'")
+            if self._size > 100000 and self._logger and not UPG_PARALLEL_ITER_BROWSE:
```
aj-fuentes:
```suggestion
            if self._size > 100000 and not UPG_PARALLEL_ITER_BROWSE:
```
Can't we just use `util._logger` always?

src/util/orm.py
```diff
@@ -391,11 +420,28 @@ def __init__(self, model, *args, **kw):
         self._model = model
         self._cr_uid = args[:-1]
         ids = args[-1]
-        self._size = len(ids)
+        self._size = kw.pop("size", None)
```
aj-fuentes, line 443: I had in mind to add the `query` parameter as we did in `recompute_fields`. Most usage of this method runs a query first to return ids, fetch all, unpack the ids, then call `iter_browse`. We can hide all that _inside_ this util. Your linked PR is an example of what we could avoid.

EDIT: so, basically, why not adding it here already? :)

src/util/orm.py
cawo-odoo: > Can't we just use `util._logger` always?

I thought so, yes, but then chose to follow what is done in `__iter__`, `__getattr__`  and `create`. I can remove the `if` on all of them if nobody sees a problem with that.

src/util/orm.py
cawo-odoo: another great idea :)

cawo-odoo: The only problem with that is that I want to avoid the `fetchall()`, meaning I need a `named_cursor` to fetch from postgres in chunks, meaning I need not only the id-selecting query, but also a `count()` query, because `named_cursor` doesn't provide `rowcount` before the last tuple has been fetched.

src/util/orm.py
aj-fuentes: > > Can't we just use `util._logger` always?
> 
> I thought so, yes, but then chose to follow what is done in `__iter__`, `__getattr__` and `create`. I can remove the `if` on all of them if nobody sees a problem with that.

I think it's a different usage.  `__iter__`, `__getattr__` and `create` are part of the API available for the caller to use or NOT a specific logger. Here we are outputting a warning to the full upgrade logs, so to me it actually belongs to `util._logger`. It makes sense to always output this message for any usage of iter_browse. To keep track of optimization points.

src/util/orm.py
aj-fuentes: > The only problem with that is that I want to avoid the `fetchall()`, meaning I need a `named_cursor` to fetch from postgres in chunks, meaning I need not only the id-selecting query, but also a `count()` query, because `named_cursor` doesn't provide `rowcount` before the last tuple has been fetched.

I don't see any problem with that. You can run the query twice, first via `cr` to get the rowcount (never fetch!), then with `named_cursor`. Running the query twice should be cheap compared with what would come after. And you can opt to fetch all ids if they aren't too many.
https://github.com/odoo/upgrade-util/blob/13adede3ed38f37030db825c5655172659f94614/src/util/orm.py#L291-L304

aj-fuentes: A second option is
```
WITH query_output AS ({query})
SELECT count(*) FROM query_output
UNION ALL
SELECT * FROM query_output
```
The first item is the count, `fetchone` is harmless :)

src/util/orm.py
```diff
@@ -391,11 +412,28 @@ def __init__(self, model, *args, **kw):
...
         self._chunk_size = kw.pop("chunk_size", 200)  # keyword-only argument
         self._logger = kw.pop("logger", _logger)
         self._strategy = kw.pop("strategy", "flush")
-        assert self._strategy in {"flush", "commit"}
+        assert self._strategy in {"flush", "commit", "multiprocessing"}
+        if self._strategy == "multiprocessing":
+            if UPG_PARALLEL_ITER_BROWSE:
+                self._task_size = self._chunk_size
+                self._chunk_size = max(get_max_workers() * 10 * self._task_size, 1000000)
+            elif self._size > 100000:
```
aj-fuentes: If we downgrade multiprocessing to commit when the variable is unset we don't need to check it anymore.
```suggestion
            else:
                self._strategy = "commit"  # downgrade
                if self._size > 100000:
                   ...  current warning
                else:
                   ... log info: this call is marked as multi-process compatible
                   downgraded to commit since env var not set
```

src/util/orm.py
```diff
@@ -447,6 +487,25 @@ def __getattr__(self, attr):
 
         def caller(*args, **kwargs):
             args = self._cr_uid + args
+            if self._strategy == "multiprocessing" and UPG_PARALLEL_ITER_BROWSE:
```
aj-fuentes:
```suggestion
            if self._strategy == "multiprocessing":
```

src/util/orm.py
```diff
@@ -412,8 +450,10 @@ def _browse(self, ids):
         return self._model.browse(*args)
 
     def _end(self):
-        if self._strategy == "commit":
+        if self._strategy in ["commit", "multiprocessing"]:
             self._model.env.cr.commit()
+            if self._strategy == "multiprocessing" and UPG_PARALLEL_ITER_BROWSE:
```
aj-fuentes:
```suggestion
            if self._strategy == "multiprocessing":
```

src/util/orm.py
```diff
@@ -371,13 +393,12 @@ class iter_browse(object):
...
-    :param list(int) ids: list of IDs of the records to iterate
-    :param int chunk_size: number of records to load in each iteration chunk, `200` by
-                           default
+    :param iterable(int) ids: iterable of IDs of the records to iterate
+    :param int chunk_size: number of records to load in each iteration chunk, `200` by default
     :param logger: logger used to report the progress, by default
                    :data:`~odoo.upgrade.util.orm._logger`
     :type logger: :class:`logging.Logger`
-    :param str strategy: whether to `flush` or `commit` on each chunk, default is `flush`
+    :param str strategy: whether to `flush`, `commit` or `multiprocess` each chunk, default is `flush`
```
aj-fuentes: I think we can leave multiprocess (sic - multiprocessing?) out the online docs. Once we have tested it enough we can publish it. In any case the documentation is incomplete because it should refer to the env variable. Ideally with some warning about why this is experimental.

aj-fuentes (commented):
To simplify the code I think we should check UPG_PARALLEL_ITER_BROWSE only _once_.

src/util/orm.py
```diff
@@ -338,6 +343,23 @@ def get_ids():
         invalidate(records)
 
 
+def _mp_iter_browse_cb(ids_or_values):
```
aj-fuentes:
```suggestion
def _mp_iter_browse_cb(ids_or_values, params):
```

src/util/orm.py
```diff
@@ -338,6 +343,23 @@ def get_ids():
...
+    me = _mp_iter_browse_cb
+    # init upon first call. Done here instead of initializer callback, because py3.6 doesn't have it
+    if not hasattr(me, "env"):
+        sql_db._Pool = None  # children cannot borrow from copies of the same pool, it will cause protocol error
+        me.env = env(sql_db.db_connect(me.params["dbname"]).cursor())
```
aj-fuentes:
```suggestion
        me.env = env(sql_db.db_connect(params["dbname"]).cursor())
```

src/util/orm.py
```diff
@@ -338,6 +343,23 @@ def get_ids():
...
+        me.env.clear()
+    # process
+    if me.params["mode"] == "browse":
+        getattr(me.env[me.params["model_name"]].browse(ids_or_values), me.params["attr_name"])(
+            *me.params["args"], **me.params["kwargs"]
```
aj-fuentes:
```suggestion
    if params["mode"] == "browse":
        getattr(me.env[params["model_name"]].browse(ids_or_values), params["attr_name"])(
            *params["args"], **params["kwargs"]
```

src/util/orm.py
```diff
@@ -338,6 +343,23 @@ def get_ids():
...
+        )
+    if me.params["mode"] == "create":
+        me.env[me.params["model_name"]].create(ids_or_values)
```
aj-fuentes:
```suggestion
    if params["mode"] == "create":
        me.env[params["model_name"]].create(ids_or_values)
```

src/util/orm.py
```diff
@@ -447,6 +487,25 @@ def __getattr__(self, attr):
 
         def caller(*args, **kwargs):
             args = self._cr_uid + args
+            if self._strategy == "multiprocessing" and UPG_PARALLEL_ITER_BROWSE:
+                _mp_iter_browse_cb.params = {
```
aj-fuentes:
```suggestion
                params = {
```

src/util/orm.py
```diff
@@ -412,8 +450,10 @@ def _browse(self, ids):
         return self._model.browse(*args)
 
     def _end(self):
-        if self._strategy == "commit":
+        if self._strategy in ["commit", "multiprocessing"]:
             self._model.env.cr.commit()
+            if self._strategy == "multiprocessing" and UPG_PARALLEL_ITER_BROWSE:
+                delattr(_mp_iter_browse_cb, "params")
```
aj-fuentes:
```suggestion
```

src/util/orm.py
```diff
@@ -447,6 +487,25 @@ def __getattr__(self, attr):
...
+                    "model_name": self._model._name,
+                    "attr_name": attr,
+                    "args": args,
+                    "kwargs": kwargs,
+                    "mode": "browse",
+                }
+                with ProcessPoolExecutor(max_workers=get_max_workers()) as executor:
+                    for chunk in it:
+                        collections.deque(
+                            executor.map(_mp_iter_browse_cb, chunks(chunk._ids, self._task_size, fmt=tuple)),
```
aj-fuentes: Finally:
```suggestion
                            executor.map(_mp_iter_browse_cb, chunks(chunk._ids, self._task_size, fmt=tuple), itertools.repeat(params)),
```

aj-fuentes (commented):
Proposal: get rid of the hack injecting `params`.

Code is simpler, easier to read, and less hacky.

cawo-odoo:
> Proposal: get rid of the hack injecting `params`.
> 
> Code is simpler, easier to read, and less hacky.

It is true that it looks that way, but there's on big disadvantage: it reduces the possible use-cases, because now everything in there (including `{kw,}args`) needs to be :cucumber:able.

src/util/orm.py
```diff
@@ -447,6 +487,25 @@ def __getattr__(self, attr):
...
+            if self._strategy == "multiprocessing" and UPG_PARALLEL_ITER_BROWSE:
+                _mp_iter_browse_cb.params = {
+                    "dbname": self._model.env.cr.dbname,
+                    "model_name": self._model._name,
+                    "attr_name": attr,
+                    "args": args,
+                    "kwargs": kwargs,
+                    "mode": "browse",
+                }
+                with ProcessPoolExecutor(max_workers=get_max_workers()) as executor:
```
KangOl: Like #294 
```suggestion
                extrakwargs = {"mp_context": multiprocessing.get_context("fork")} if sys.version_info >= (3, 7) else {}
                with ProcessPoolExecutor(max_workers=get_max_workers(), **extrakwargs) as executor:
```

aj-fuentes:
> > Proposal: get rid of the hack injecting `params`.
> > Code is simpler, easier to read, and less hacky.
> 
> It is true that it looks that way, but there's one big disadvantage: it reduces the possible use-cases, because now everything in there (including `{kw,}args`) needs to be 🥒able.

Do you have a use case where we need to pass more complex objects? I think once we have it we can make this hacky.

src/util/orm.py
cawo-odoo: :man_facepalming: thank you :)

src/util/orm.py
```diff
@@ -447,6 +487,25 @@ def __getattr__(self, attr):
...
         def caller(*args, **kwargs):
             args = self._cr_uid + args
+            if self._strategy == "multiprocessing" and UPG_PARALLEL_ITER_BROWSE:
+                _mp_iter_browse_cb.params = {
+                    "dbname": self._model.env.cr.dbname,
+                    "model_name": self._model._name,
+                    "attr_name": attr,
+                    "args": args,
+                    "kwargs": kwargs,
+                    "mode": "browse",
```
KangOl: You also need to pass the context
```suggestion
                    "mode": "browse",
                    "context": self._model.context,
```

cawo-odoo:
> you're tying the code to the _internal implementation_ of the standard Python library.

What do you mean? The fact that it forks? This whole thing is tied to that, as are all of our other uses of ProcessPoolExecutor, that's why https://github.com/odoo/upgrade-util/pull/320#discussion_r2348147163 is important.

I also wanted simplicity: in the callback signature, to avoid as many serialisation issues as we can. OTOH, your suggestion is more readable and easier to understand. I think both are valid goals and they are only in conflict here, because `ProcessPoolExecutor.map` (not `map`) is *not* a simple interface, it just pretends to be. I made this point on the last PR already and it didn't change, there are all kinds of quirks in this thing.

Anyway. It is true that input to this code - if used - has to be chosen very carefully for all kinds of reasons. So maybe it is a good idea to restrict, in this way, the arguments that can be passed to the model attr being called, for the purpose of limiting the use-cases to simpler ones that hopefully have a lower risk of dangerous use.

aj-fuentes:
> What do you mean? The fact that it forks?

No, the fact that `map` would use pickle to send the parameters. 

Once more, I'm OK with this. You could probably include some comments about why you chose to do this. I'd say something like: "we want to avoid the overhead of serializing the same parameters for all ids via pickle in the map implementation. Warning: we need to clean them up in _end to avoid any potential conflicts".

Note that we need to keep in mind --some testing would be nice-- what happens for multiple calls to iter_browse with multiprocessing within the same run (upgrade step). I don't see anything problematic at the moment.

> it just pretends to be

Yes. We should honor that --if we can--, that's the point.

src/util/orm.py
```diff
@@ -9,11 +9,17 @@
...
 """
 
+import collections
 import logging
+import multiprocessing
+import os
 import re
+import sys
+import uuid
+from concurrent.futures import ProcessPoolExecutor
```
KangOl: Won't work in python2

src/util/orm.py
```diff
@@ -476,31 +616,72 @@ def create(self, values, **kw):
         if self._size:
             raise ValueError("`create` can only called on empty `browse_record` objects.")
 
-        ids = []
-        size = len(values)
-        it = chunks(values, self._chunk_size, fmt=list)
+        if self._strategy == "multiprocessing" and not multi:
```
aj-fuentes: As discussed in DM: we should have two private implementations. The current one under `_create` and the new one in `_create_multiplrocessing` we call each depending on the options.

src/util/orm.py
```diff
@@ -476,31 +616,72 @@ def create(self, values, **kw):
...
+            raise ValueError("The multiprocessing strategy only supports the multi version of `create`")
+
+        if isinstance(values, SQLStr):
```
aj-fuentes: Why don't we add a `query` parameter making `values=None` by default. To make this more explicit.
```py
def create(self, values=None, query=None, **kw):
```

In any case if we keep hacking `values` for queries I wouldn't use `SQLStr`, just check against plain `str`. Normal values of `create` are never strings so this won't risk any confusion. I still believe being explicit is better though.

cawo-odoo: addressed in https://github.com/odoo/upgrade-util/pull/350

src/util/orm.py
cawo-odoo: done

cawo-odoo:
added fixups to get rid of the scattered `_superchunk_size` logic and replaced it with local "batching" logic around `executor.map` calls. I'll fix py2 compat when #350 has finally been merged.

PR #321, opened 2025-09-15
[FIX] util/fields: handle o2m when removing relation field

aj-fuentes:
`relation_field_id` is an ON DELETE CASCADE FK. This triggers the
removal of entries in `ir_model_fields` when we remove a field
referenced by an o2m field. This action leaves many (metadata) leftovers
for manual fields. To avoid potential issues we call `remove_field` on
such fields. That way the full cleanup (domains, contexts, filters,
etc.) happens.

We also now warn about compute implementation of manual fields if we
detect that the implementation may refer to a field being removed from
the depends.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -278,10 +278,10 @@ def remove_field(cr, model, fieldname, cascade=False, drop_column=True, skip_inh
     # remove this field from dependencies of other fields
     if column_exists(cr, "ir_model_fields", "depends"):
         cr.execute(
-            "SELECT id,model,depends FROM ir_model_fields WHERE state='manual' AND depends ~ %s",
+            "SELECT id,model,depends,COALESCE(compute,''),name FROM ir_model_fields WHERE state='manual' AND depends ~ %s",
             [r"\m{}\M".format(fieldname)],
         )
-        for id, from_model, deps in cr.fetchall():
+        for id, from_model, deps, compute_code, dep_field_name in cr.fetchall():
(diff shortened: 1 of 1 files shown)


src/util/fields.py
```diff
@@ -393,6 +422,11 @@ def adapter(leaf, is_or, negated):
     if drop_column and stored:
         remove_column(cr, table, fieldname, cascade=cascade)
 
+    # Remove related one2many fields that are cascade removed due to the removal of their relation field
+    for rel_model, rel_field in o2m_to_remove:
+        _logger.info("Cascade removing one2many field %s.%s", rel_model, rel_field)
+        remove_field(cr, rel_model, rel_field)
```
KangOl: Shouldn't it be done *before* the deletion (in cascade) of the `ir_model_fields` row?

aj-fuentes: Logically, yes. Practically, I don't like to have a field half-removed while triggering the removal of _another_ field. I prefer to ensure the current field is fully removed then trigger extra removals at the very end.

KangOl: I understand. Then, maybe the deletion of the `ir_model_fields` entry can be done *after*. This can be done after the handling of the translations, defaults and attachments.

aj-fuentes: That may work. Checking...

aj-fuentes: Pushed the new version

aj-fuentes:

PR #322, opened 2025-09-19
[FIX] orm: in recompute_fields, avoid memory error from postgresql

cawo-odoo:
improves 13adede3ed38f37030db825c5655172659f94614

Queries run through client-side cursors will make postgresql materialze the whole of the result immediately (which is actually, why `cr.rowcount` is always available right after `execute` in this case). With server side cursors (named cursors) on the other hand, tuples are materialized when they are fetched. This is why running the `query` for ids through the client-side cursor just to be able to access `cr.rowcount`, can cause an out-of-memory exception from PostgreSQL.

We fix this by wrapping the query in a `CREATE TABLE AS` statement that inserts returned ids into a temporary table. We then use a named_cursor to fetch ids from this table in chunks, server-side.

Another approach would have been to just wrap the query in a `SELECT count(*)` query and run this once to get the `count`. The approach using `CREATE TABLE AS` has been chosen over that solution to support queries that include DML statements (e.g. `UPDATE ... RETURNING`) that affect the results of the compute, as it allows us to run the query on the main (client) cursor, while still using a named_cursor for fetching the ids memory-efficiently.

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -42,7 +42,7 @@ from .const import BIG_TABLE_THRESHOLD
 from .exceptions import MigrationError
 from .helpers import table_of_model
 from .misc import chunks, log_progress, version_between, version_gte
-from .pg import column_exists, format_query, get_columns, named_cursor
+from .pg import SQLStr, column_exists, format_query, get_columns, named_cursor
 
 # python3 shims
 try:
@@ -278,7 +278,8 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
     :param int chunk_size: number of records per chunk - used to split the processing
     :param str strategy: strategy used to process the re-computation
     :param str query: query to get the IDs of records to recompute, it is an error to set
-                      both `ids` and `query`
+                      both `ids` and `query`. Note that the processing will always happen
+                      in ascending order. If that is unwanted, you must use `ids` instead.
     """
     if strategy not in {"flush", "commit", "auto"}:
         raise ValueError("Invalid strategy {!r}".format(strategy))
@@ -288,25 +289,26 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
     model = Model._name
 
     if ids is None:
-        query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
-        cr.execute(query)
+        query = format_query(cr, "SELECT id FROM {}", table_of_model(cr, model)) if query is None else SQLStr(query)
+        cr.execute(
+            format_query(cr, "CREATE UNLOGGED TABLE _upgrade_rf(id) AS (WITH query AS ({}) SELECT * FROM query)", query)
+        )
         count = cr.rowcount
-        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
-            ids_ = (id_ for (id_,) in cr.fetchall())
-        else:
+        cr.execute("ALTER TABLE _upgrade_rf ADD CONSTRAINT pk_upgrade_rf_id PRIMARY KEY (id)")
 
-            def get_ids():
-                with named_cursor(cr, itersize=2**20) as ncr:
-                    ncr.execute(query)
-                    for (id_,) in ncr:
-                        yield id_
+        def get_ids():
+            with named_cursor(cr, itersize=2**20) as ncr:
+                ncr.execute("SELECT id FROM _upgrade_rf ORDER BY id")
+                for (id_,) in ncr:
+                    yield id_
 
-            ids_ = get_ids()
+        ids_ = get_ids()
     else:
         count = len(ids)
         ids_ = ids
 
     if not count:
+        cr.execute("DROP TABLE IF EXISTS _upgrade_rf")
         return
 
     _logger.info("Computing fields %s of %r on %d records", fields, model, count)
@@ -336,6 +338,7 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
         else:
             flush(records)
         invalidate(records)
+    cr.execute("DROP TABLE IF EXISTS _upgrade_rf")
 
 
 class iter_browse(object):


src/util/orm.py
```diff
@@ -289,14 +289,16 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
 
     if ids is None:
         query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
-        cr.execute(query)
-        count = cr.rowcount
-        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
+        count_query = "SELECT count(*) FROM ({query})".format(query=query)
+        cr.execute(count_query)
+        count = cr.fetchone()[0]
```
KangOl: Actually, `SELECT count(*)` itself can be slow.
We can just estimate the number of row with `EXPLAIN`
```suggestion
        count_query = format_query(cr, "EXPLAIN (FORMAT JSON) {}", query)
        cr.execute(count_query)
        plan = cr.fetchone()[0]
        count = plan[0]['Plan']['Plan Rows']  # estimated count, actually
```

---
Source: https://wiki.postgresql.org/wiki/Count_estimate

src/util/orm.py
```diff
@@ -289,14 +289,16 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+        itersize = 100000
```
KangOl: Maybe we can reuse BIG_TABLE_THRESHOLD
```suggestion
        itersize = BIG_TABLE_THRESHOLD
```
Or a factor of it
```suggestion
        itersize = 2 * BIG_TABLE_THRESHOLD
```

src/util/orm.py
cawo-odoo: It's a nice idea to use the table stats, but I'm unsure we want the logging to be off potentially:
https://github.com/odoo/upgrade-util/pull/322/files#diff-e83e04eeae2ce53657ba28fd8de2126f5f3ef6422a73066af36dcf9191de1278R321-R323

Also, a parallel seq scan that takes 20s should be dwarfed by the overall runtime of this.

cawo-odoo:
Tests with the currently effected DB look good.

src/util/orm.py
```diff
@@ -289,14 +289,19 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
 
     if ids is None:
         query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
-        cr.execute(query)
-        count = cr.rowcount
-        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
+        if "RETURNING" not in query.upper().strip():  # query must be DQL
```
```suggestion
        count = None
        if "RETURNING" not in query.upper():  # query must be DQL
```

src/util/orm.py
```diff
@@ -289,14 +289,19 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+            count_query = "SELECT count(*) FROM ({query}) AS count".format(query=query)
+            cr.execute(count_query)
+            count = cr.fetchone()[0]
+        itersize = BIG_TABLE_THRESHOLD
+        if not count or count < 2 * itersize:  # avoid the overhead of a named cursor unless we have at least two chunks
```
```suggestion
        if count is None or count < 2 * itersize:  # avoid the overhead of a named cursor unless we have at least two chunks
```

src/util/orm.py
```diff
@@ -289,14 +289,19 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
 
     if ids is None:
         query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
-        cr.execute(query)
-        count = cr.rowcount
-        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
+        if "RETURNING" not in query.upper().strip():  # query must be DQL
+            count_query = "SELECT count(*) FROM ({query}) AS count".format(query=query)
```
```suggestion
            count_query = format_query(cr, "SELECT count(*) FROM ({}) AS count", SQLStr(query))
```

src/util/orm.py
aj-fuentes: Why? The query could be `select id from table`

Why don't we use a `named_cursor` always? TBH I wasn't sure about any overhead. In principle a recompute would be way slower than the expected overhead.

In summary, we can either use a CTE approach
`WITH query AS ({query}) SELECT count(*) FROM query` for the count if we want to fetch it first to decide. Or `WITH query AS ({query}) SELECT count(*) FROM query union select * from query` if we default to a named cursor. Note that in the later still if there aren't too many ids we can still fetch them all and close the named cursor.

aj-fuentes: Actually I think the only safe way is to use a temp table to store the ids.
A delete or update query cannot be run twice in the general case. Also an update could affect how the actual compute of the fields go. Thus, we should fully run the query first then fetch the ids.

cawo-odoo: ah, dammit

KangOl: It's to overcome queries like https://runbot.odoo.com/runbot/build/89536248

It would indeed work with a CTE, but the query can't be done twice.

aj-fuentes: > It's to overcome queries like https://runbot.odoo.com/runbot/build/89536248

That part I got it. But we don't need a RETURNING always, that's what I didn't get.

> It would indeed work with a CTE, but the query can't be done twice.

Indeed. Thus to me we need a temp table, a-la-rrr style.

EDIT: we cannot run the query in a named_cursor since our main cursor won't see the change, hence computes may be wrong.

cawo-odoo: > Why? The query could be `select id from table`

well why yes, that's what we need to be sure of, that there's no DML in the query, not in the main part, nor ina n eventual CTE, but since we now that only queries that return tuples will be passed to this function (anything else doesn't make sense at all), we can make sure of that by looking for RETURNING, no?

> Why don't we use a `named_cursor` always? TBH I wasn't sure about any overhead. In principle a recompute would be way slower than the expected overhead.

That would be totally fine with me. It does not remove the necessity to get the row count up-front though, because we cannot pass a generator to `log_progress` and not give it a `size`.

> In summary, we can either use a CTE approach `WITH query AS ({query}) SELECT count(*) FROM query`

What would be the advantage? PG seems to be doing the exact same thing as with the sub-query.

> for the count if we want to fetch it first to decide. Or `WITH query AS ({query}) SELECT count(*) FROM query union select * from query` if we default to a named cursor. Note that in the later still if there aren't too many ids we can still fetch them all and close the named cursor.

This would still not work when `query` contains a DML statement.

cawo-odoo: > EDIT: we cannot run the query in a named_cursor since our main cursor won't see the change, hence computes may be wrong.

yes :(

cawo-odoo: > That part I got it. But we don't need a RETURNING always, that's what I didn't get.

But we do, don't we? I mean passing a query to this function only makes sense if it returns ids. Meaning, it is either a DQL statement suitable for use in both `count_query` and a `named_cursor`, or it is a DML statement, in which case we can safely assume it contains a `RETURNING` keyword to return the ids, no? (EDIT: and in the latter case, we have to use the main cursor)

src/util/orm.py
cawo-odoo: I don't think we want to enter the `named_cursor` branch if `count == 0`, do we?

src/util/orm.py
aj-fuentes: Ah, I got the logic in the `if` backwards. It kind of makes sense now, if you detect a `RETURNING` you skip it. This is still brittle:
```
WITH _ignored AS (DELETE FROM table WHERE cond) SELECT id FROM table WHERE NOT cond
```
That's a valid query that updates stuff without any returning.

KangOl: So we have two options:
 - always use a named_cursor()
 - use a temporary table

aj-fuentes: > So we have two options:
>  * always use a named_cursor()
>  * use a temporary table

I think using a named cursor is not an option. Imagine that there is an update/delete in the `query` that affects how the _compute_ works. If we compute on the main cursor it won't see the changes done in the named cursor. To me the only full solution is to fully execute the original query and save its output in a temp table. Fetching ids (with named cursor if too many) from it.

KangOl:
upgradeci retry with always only account hr_recruitment_extract hr_expense sale_stock sale_subscription crm l10n_it_edi sale purchase mrp

src/util/orm.py
```diff
@@ -288,25 +288,30 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
     if ids is None:
-        query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
+        query = format_query(
+            cr,
+            "WITH query AS ({}) SELECT * INTO UNLOGGED _upgrade_rf FROM query",
+            SQLStr(
+                format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
+            ),
+        )
         cr.execute(query)
```
aj-fuentes: I think it reads clearer if we se the query with some default first then execute the formatted wrapper.

I believe that [`CREATE TABLE AS`](https://www.postgresql.org/docs/12/sql-createtableas.html) is the right tool here.
```suggestion
        query = format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else SQLStr(query)
        cr.execute(
            format_query(cr, "CREATE UNLOGGED TABLE _upgrade_rf(id) AS ({})", query)
        )
```
Specifying the columns (id in this case) will fail early for bad queries.

Example (working well even with side-effect queries)
```
test_16=> CREATE UNLOGGED TABLE _upgrade_rf(id) AS (with x as (DELETE FROM res_users WHERE id=7) SELECT id from res_users WHERE id<7)
SELECT 6
```

src/util/orm.py
```diff
@@ -288,25 +288,30 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+        count = cr.fetchone()[0]
 
-            def get_ids():
-                with named_cursor(cr, itersize=2**20) as ncr:
-                    ncr.execute(query)
-                    for (id_,) in ncr:
-                        yield id_
+        def get_ids():
+            with named_cursor(cr, itersize=2**20) as ncr:
+                ncr.execute("SELECT * FROM _upgrade_rf")
```
aj-fuentes: Make it fail if the query is wrong. Avoid id-injection to the ORM.
```suggestion
                ncr.execute("SELECT id::int FROM _upgrade_rf")
```

KangOl: I won't cast it to `int`. On some (really) huge tables, it's not excluded that they has to promote the id to `bigint`.

aj-fuentes: This was an optional stuff. But note that we could get `select name from table` as query. Whether this is an issue for the ORM or not it would depend on the running version... The error later could be harder to understand than a type cast error.

src/util/orm.py
```diff
@@ -288,25 +288,30 @@ def recompute_fields(cr, model, fields, ids=None, logger=_logger, chunk_size=256
...
+                format_query(cr, "SELECT id FROM {} ORDER BY id", table_of_model(cr, model)) if query is None else query
+            ),
+        )
         cr.execute(query)
-        count = cr.rowcount
-        if count < 2**21:  # avoid the overhead of a named cursor unless we have at least two chunks
-            ids_ = (id_ for (id_,) in cr.fetchall())
-        else:
+        cr.execute("SELECT count(*) FROM _upgrade_rf")
+        count = cr.fetchone()[0]
```
aj-fuentes: Can't we get back to `rowcount` if we use CREATE TABLE AS? Is there any special case that would break it? I don't see any.
```suggestion
        count = cr.rowcount
```

src/util/orm.py
cawo-odoo: OK, I made it like that. We also need to move the "ORDER by" to the right query and for that, we should have an index on _upgrade_rf(id)

src/util/orm.py
cawo-odoo: > This was an optional stuff. But note that we could get `select name from table` as query. Whether this is an issue for the ORM or not it would depend on the running version... The error later could be harder to understand than a type cast error.

If we really want to do that, I would make it such that it runs a normal `create table` first. In there, we can specify the column type and create the PK all in one go.

src/util/orm.py
aj-fuentes: We don't _need_ order by. We can assume the order is intentional in any explicit query (that's the logic I followed at least).

src/util/orm.py
cawo-odoo: create table as doesn't return any rows, I don't think cr.rowcount will be set

cawo-odoo: > create table as doesn't return any rows, I don't think cr.rowcount will be set

no, I'm wrong. we can use it

src/util/orm.py
aj-fuentes: We don't _need_ to :) I just wanted a quick/simple way to avoid the issue. We could cast to bigint as well, but it may have some perf impact.

src/util/orm.py
cawo-odoo: But if we don't `order by`, compute order will be undeterministic, no? Smells of hard to debug side-effects.

aj-fuentes: We order in the default query. It's up to review to ensure the `query` parameter has the intended order. If we override the order we may break some specific order the author intended.

cawo-odoo: I think I wasn't clear enough about what I meant. I moved the order by here:
https://github.com/odoo/upgrade-util/pull/322/files#diff-e83e04eeae2ce53657ba28fd8de2126f5f3ef6422a73066af36dcf9191de1278R298

Because we no longer fetch the ids from the input (or default) query. We fetch them from the temp table. If we do not specify order by on that query, it will be undeterministic. I.e. any order specified on the query passed by the caller is lost in any case. If the caller needs a special order, he cannot use this interface.

aj-fuentes: It's true that nothings prevent the results from being in some different order. I guess we can order always. In practice IDK if we can trust the order will be respected as long as the temp table is untouched.

cawo-odoo: > In practice IDK if we can trust the order will be respected as long as the temp table is untouched.

I'm pretty sure that at least if PG decides to do a // seq scan, any such assumption will be wrong.

aj-fuentes: This is an issue then. We would need to order always to be consistent. ~~That includes ordering `ids` if it is a list.~~
EDIT: or maybe not. We can document that `ids` order is respected, but `query` is always ordered by `id` (ascending). That way if certain order must be respected the caller needs to use the `ids=` version.

cawo-odoo: > We can document that `ids` order is respected, but `query` is always ordered by `id` (ascending). That way if certain order must be respected the caller needs to use the `ids=` version.

That's what I was thinking

aj-fuentes:
I think your last patch is leaving behind the table _upgrade_rf when count is zero.

cawo-odoo:
Updated commit message. From my side this is ready and it is still the same what Alvaro approved already.

KangOl (approved):
@robodoo r+

PR #323, opened 2025-09-23
[FIX] fields: update default selection

jjmaksoud:
When calling `change_field_selection_values`, any default value is not updated. This fix updates the value in the default values table.

https://runbot.odoo.com/odoo/error/230875
TBG-2171

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1131,6 +1131,31 @@ class TestField(UnitTestCase):
         initial_repartition[False] += initial_repartition.pop(None, 0)
         self.assertEqual(back_repartition, initial_repartition)
 
+    def test_change_field_selection_with_default(self):
+        cr = self.env.cr
+        lang = self.env["res.lang"].create({"name": "Elvish", "code": "el_VISH", "active": True})
+        if util.table_exists(cr, "ir_default"):
+            self.env["ir.default"].set("res.partner", "lang", "el_VISH")
+        else:
+            self.env["ir.values"].set_default("res.partner", "lang", "el_VISH")
+        util.flush(lang)
+        partner = self.env["res.partner"].create({"name": "Gandalf"})
+        self.assertEqual(partner.lang, "el_VISH")
+
+        util.invalidate(partner)
+        util.change_field_selection_values(cr, "res.partner", "lang", {"el_VISH": "en_US"})
+
+        self.assertEqual(partner.lang, "en_US")
+
+        if util.table_exists(cr, "ir_default"):
+            new_default = (getattr(self.env["ir.default"], "get", None) or self.env["ir.default"]._get)(
+                "res.partner", "lang"
+            )
+        else:
+            new_default = self.env["ir.values"].get_default("res.partner", "lang")
+
+        self.assertEqual(new_default, "en_US")
+
 
 class TestHelpers(UnitTestCase):
     def test_model_table_conversion(self):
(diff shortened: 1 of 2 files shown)


src/util/fields.py
```diff
@@ -1187,6 +1187,39 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
...
+        dumped_map = {pickle.dumps(k): pickle.dumps(v) for k, v in mapping.items()}
+    else:
+        query = """
+            UPDATE ir_default d
+               SET json_value = (%(json)s::jsonb->>d.json_value)
+              FROM ir_model_fields f
+             WHERE d.field_id = f.id
+               AND f.model = %(model)s
+               AND f.name = %(name)s
+               AND d.json_value IN %(keys)s
```
aj-fuentes, line 1214: Does this work? I'd expect json_value to have `"key"` with double quotes.

jjmaksoud: That's why I used `dumped_map = {json.dumps(k): json.dumps(v) for k, v in mapping.items()}`
I tested it for the error in the CI linked above, and the default value was updated

src/util/fields.py
```diff
@@ -1187,6 +1187,39 @@ def change_field_selection_values(cr, model, field, mapping, skip_inherit=()):
             [model, field, [k for k in mapping if k not in mapping.values()]],
         )
 
+    if table_exists(cr, "ir_values"):
+        try:
```
aj-fuentes: Move the imports to top-level

src/base/tests/test_util.py
```diff
@@ -1131,6 +1131,25 @@ def test_invert_boolean_field(self):
...
         self.assertEqual(back_repartition, initial_repartition)
 
+    def test_change_field_selection_with_default(self):
+        cr = self.env.cr
+        lang = self.env["res.lang"].create({"name": "Elvish", "code": "el_VISH", "active": True})
+        if util.table_exists(cr, "ir_default"):
+            self.env["ir.default"].set("res.partner", "lang", "el_VISH")
+        else:
+            self.env["ir.values"].set_default("res.partner", "lang", "el_VISH")
+        util.flush(lang)
```
jjmaksoud, line 1141: needed in versions <= 17.0 for the default to take effect

src/base/tests/test_util.py
```diff
@@ -1131,6 +1131,25 @@ def test_invert_boolean_field(self):
...
+        partner = self.env["res.partner"].create({"name": "Gandalf"})
+        self.assertEqual(partner.lang, "el_VISH")
+
+        util.invalidate(partner)
+        util.clear_cache(partner)
```
jjmaksoud: needed to update the cached get_default

jjmaksoud:
upgradeci retry with always only stock_account account_peppol sale sale_subscription hr in all verions

aj-fuentes:
all **verions** certainly not XD

src/base/tests/test_util.py
aj-fuentes, line 1141: I think you can just call 
```
util.change_field_selection_values(cr, "res.partner", "lang", {"el_VISH": "en_US"})
```
then check that the default was set `self.env["ir.default"]._get("res.partner", "lang")` or the alternative for `ir.values`.

We don't need to include the full logic checking if the default is set for new partners. It's testing more than what we need.

EDIT: you still probably need to flush and invalidate.

jjmaksoud: Do I keep the check that the value is changed for existing partners? there's no other test validating it

aj-fuentes: Yes. It's good to have it covered.

jjmaksoud:
upgradeci retry with always all versions

aj-fuentes (approved):

PR #331, opened 2025-10-02
[FIX] modules: defer new dependency installs

jjmaksoud:
When updating module dependencies from graph with auto discovery, all dependency diffs should be saved before attempting the installation of the modules that are new dependencies for installed modules, or force installed modules. This is to avoid auto installing unwanted modules based on their old dependencies.

**case:**
Install website_sale and l10n_mx_edi,web_map (dependecies of l10n_mx_edi_stock without delivery) in version 16.0
upgrade to 17.0
The module delivery is a new dependency for website_sale
The module l10n_mx_edi_stock has a dependency change from delivery to stock_delivery

Before the fix:
The dependency change for website_sale will install delivery before the dependency change for l10n_mx_edi_stock is persisted, leading to assume that its autoinstall condition is met
When l10n_mx_edi_stock is force installed, it installs stock with it, and many other modules because of that

After the fix:
all module dependencies are updated first,
the installation of delivery will not install l10n_mx_edi_stock or stock

### DIFF
--- a/src/util/modules.py
+++ b/src/util/modules.py
@@ -549,20 +549,25 @@ def force_install_module(cr, module, if_installed=None, reason="it has been expl
                                            already installed
     :return str: the *new* state of the module
     """
-    try:
-        return _force_install_module(cr, module, if_installed, reason)
-    except UnknownModuleError:
-        if version_gte("saas~14.5"):
-            # We must delay until the modules actually exists. They are added by the auto discovery process.
+    if version_gte("saas~14.5"):
+        if if_installed:
+            _assert_modules_exists(cr, *if_installed)
+        if not if_installed or modules_installed(cr, *if_installed):
             ENVIRON["__modules_auto_discovery_force_installs"].add(module)
-            return None
-        else:
-            raise
+        return None
+    else:
+        return _force_install_module(cr, module, if_installed, reason)
 
 
 def _force_install_module(cr, module, if_installed=None, reason="it has been explicitly asked for"):
     """Strict implementation of force_install: raise errors for missing modules."""
     _assert_modules_exists(cr, module)
+
+    cr.execute("SELECT state FROM ir_module_module WHERE name = %s", [module])
+    mod_state = cr.fetchone()[0]
+    if mod_state in INSTALLED_MODULE_STATES:
+        return mod_state
+
     subquery = ""
     params = [module]
     if if_installed:
@@ -675,7 +680,7 @@ def _force_install_module(cr, module, if_installed=None, reason="it has been exp
             [toinstall, list(INSTALLED_MODULE_STATES)],
         )
         for (mod,) in cr.fetchall():
-            force_install_module(
+            _force_install_module(
                 cr,
                 mod,
                 reason=(
@@ -716,13 +721,14 @@ def new_module_dep(cr, module, new_dep):
         [new_dep, module, new_dep],
     )
 
-    # Update new_dep state depending on module state
-    cr.execute("SELECT state FROM ir_module_module WHERE name = %s", [module])
-    mod_state = (cr.fetchone() or ["n/a"])[0]
-    if mod_state in INSTALLED_MODULE_STATES:
-        # Module was installed, need to install all its deps, recursively,
-        # to make sure the new dep is installed
-        force_install_module(cr, new_dep, reason="it's a new dependency of {!r}".format(module))
+    if not version_gte("saas~14.5"):
+        # Update new_dep state depending on module state
+        cr.execute("SELECT state FROM ir_module_module WHERE name = %s", [module])
+        mod_state = (cr.fetchone() or ["n/a"])[0]
+        if mod_state in INSTALLED_MODULE_STATES:
+            # Module was installed, need to install all its deps, recursively,
+            # to make sure the new dep is installed
+            _force_install_module(cr, new_dep, reason="it's a new dependency of {!r}".format(module))
 
 
 def remove_module_deps(cr, module, old_deps):
@@ -1025,8 +1031,25 @@ def _trigger_auto_discovery(cr):
             _set_module_countries(cr, module, countries)
             module_auto_install(cr, module, auto_install)
 
-        if module in force_installs:
-            _force_install_module(cr, module)
+    cr.execute(
+        """
+        SELECT d.name, STRING_AGG(m.name, ', ' ORDER BY m.name)
+          FROM ir_module_module_dependency d
+          JOIN ir_module_module m
+            ON d.module_id = m.id
+          JOIN ir_module_module dm
+            ON dm.name = d.name
+         WHERE m.state IN %(installed_state)s
+           AND dm.state NOT IN %(installed_state)s
+      GROUP BY d.name
+        """,
+        {"installed_state": INSTALLED_MODULE_STATES},
+    )
+    for new_dep, modules in cr.fetchall():
+        _force_install_module(cr, new_dep, reason="it's a new dependency of {}".format(modules))
+
+    for module in force_installs:
+        _force_install_module(cr, module)
 
     for module, (init, version) in ENVIRON["__modules_auto_discovery_force_upgrades"].items():
         _force_upgrade_of_fresh_module(cr, module, init, version)


src/util/modules.py
```diff
@@ -540,6 +540,18 @@ def _up(table, old, new):
         force_install_module(cr, into, reason="installed {!r} module has been merged into it".format(old))
 
 
+def _force_install_new_dependency(cr, module, reason="it is a new dependency of an installed module"):
```
aj-fuentes: Why not inlining this right in the only usage we have for it? Everything is clearer in that context.

jjmaksoud: moved

jjmaksoud:

src/util/modules.py
```diff
@@ -549,20 +549,34 @@ def force_install_module(cr, module, if_installed=None, reason="it has been expl
...
+                """
+                SELECT 1
+                  FROM ir_module_module
+                 WHERE name IN %s
+                   AND state IN %s
+                """,
+                [tuple(if_installed), INSTALLED_MODULE_STATES],
+            )
+            should_install = cr.rowcount == len(if_installed)
+        if not if_installed or should_install:
```
KangOl:
```suggestion
        if not if_installed or modules_installed(cr, *if_installed):
```

src/util/modules.py
```diff
@@ -1025,8 +1029,25 @@ def _trigger_auto_discovery(cr):
             _set_module_countries(cr, module, countries)
             module_auto_install(cr, module, auto_install)
 
-        if module in force_installs:
-            _force_install_module(cr, module)
+    cr.execute(
```
aj-fuentes, line 1034: @KangOl and I discussed about this some time ago. There is an important detail here: the data in those tables is wrong unless all modules we need were already loaded. It's at load/upgrade time that the new dependencies are updated. I don't recall the details and it isn't perfectly clear to me how it works in the example above. Maybe it's OK because delivery is updated before l10n_mx*. What I recall is that it looked that the only full solution was to save a snapshot of dependencies, update all dependencies (at once before loading base), and perform operations according to the diff. 

I think everything triggers from (all links in 16 -- similar in later versions):
https://github.com/odoo/odoo/blob/b2f10d3578c69e971601d6e12448adc221dc9b3a/odoo/modules/loading.py#L452
that then goes into
https://github.com/odoo/odoo/blob/b2f10d3578c69e971601d6e12448adc221dc9b3a/odoo/addons/base/models/ir_module.py#L760
that loads the info from the manifests into these tables
https://github.com/odoo/odoo/blob/d42e871821aa07a9b450eb24afca00c0ac2eb470/odoo/addons/base/models/ir_module.py#L811-L822

Note that the first link is after loading base, and post scripts are run from the load_module_graph of base
https://github.com/odoo/odoo/blob/b2f10d3578c69e971601d6e12448adc221dc9b3a/odoo/modules/loading.py#L434-L436
https://github.com/odoo/odoo/blob/b2f10d3578c69e971601d6e12448adc221dc9b3a/odoo/modules/loading.py#L238

In summary: we may need to replicate the update of metadata from manifests in the addons paths right here if we want the information to be fully correct at post- of base.

jjmaksoud: This part of `_trigger_auto_discovery` does update the dependencies from new manifests
https://github.com/odoo/upgrade-util/blob/5082db91ac38356ae1fcfbbdf2111ee16b4a58d5/src/util/modules.py#L1004-L1023

After this part the dependencies are updated and correct in the database.

The problem was that adding new dependencies can trigger an install, which could happen before all modules deps are updated in that loop. I also changed how `force_install_module` tries to install the module in new versions and made it defer the install until dependencies are updated to avoid installing anything pre- base

aj-fuentes: > This part of _trigger_auto_discovery does update the dependencies from new manifests

Indeed. I forgot about that part. It looks good then.

src/util/modules.py
```diff
@@ -1025,8 +1029,25 @@ def _trigger_auto_discovery(cr):
...
+        """
+        SELECT d.name, ARRAY_AGG(m.name)
```
KangOl:
```suggestion
        SELECT d.name, string_agg(m.name, ', ' ORDER BY m.name)
```

src/util/modules.py
```diff
@@ -1025,8 +1029,25 @@ def _trigger_auto_discovery(cr):
...
+          JOIN ir_module_module dm
+            ON dm.name = d.name
+         WHERE m.state IN %(installed_state)s
+           AND dm.state NOT IN %(installed_state)s
+      GROUP BY d.name
+        """,
+        {"installed_state": INSTALLED_MODULE_STATES},
+    )
+    for new_dep, modules in cr.fetchall():
+        _force_install_module(cr, new_dep, reason="it's a new dependency of {!r}".format(modules))
```
KangOl:
```suggestion
        _force_install_module(cr, new_dep, reason="it's a new dependency of {}".format(modules))
```

src/util/modules.py
```diff
@@ -549,20 +549,23 @@ def force_install_module(cr, module, if_installed=None, reason="it has been expl
                                            already installed
     :return str: the *new* state of the module
     """
-    try:
-        return _force_install_module(cr, module, if_installed, reason)
-    except UnknownModuleError:
-        if version_gte("saas~14.5"):
-            # We must delay until the modules actually exists. They are added by the auto discovery process.
+    if version_gte("saas~14.5"):
+        if not if_installed or modules_installed(cr, *if_installed):
```
aj-fuentes, line 555: We have a similar issue here. The modules in `if_installed` may not be discovered yet. Continuing the same discussion I referred before (with CHS), I think we can switch `__modules_auto_discovery_force_installs` into a dict: `module: if_installs`. And do the check post base load. A strict check, such that if a module in if_installed does not exist we get an error in the logs. This will detect issues in CI tests at the source.

jjmaksoud: normally the modules in `if_installed` are old modules, and we want to check their state before any installation starts. Scenario to avoid: a module is split and we force install a part if the old one is installed, but we do not want to force install the new module if the original module is force installed during the upgrade.
We only use this arg once in the upgrade repo but I see it also in specific fixes which means it can be used with custom modules, so I dont know how strict we want the check to be

aj-fuentes: Indeed we can leave it as is. In any case current code is the product of the limitations you are fixing here. Thus some alternative solutions were found (like manually calling deps diffs) that would be obsolete afterwards. We can review those later.

aj-fuentes: Note that if_installed check could check the actual `installed` state, vs the `to install` transient state.

KangOl: Given (our) low usage of the `if_installed` feature, a hard-fail in this case is OK.

KangOl, line 555:
```suggestion
        if if_installed:
            _assert_modules_exists(cr, *if_installed)
        if not if_installed or modules_installed(cr, *if_installed):
```

jjmaksoud: Added the validation `_assert_modules_exists`, but kept all installed states for not. I think `to upgrade` is also valid for custom modules if the upgrade is jumping multiple versions and the check is not done in the first upgrade step so the `installed` is already `to upgrade` in pre- base

KangOl (approved):
@robodoo r+

miikanissi:
This fix seems to be causing another issue. Since this defers the force installation of a new module until all deps are updated, what happens if you're trying to move a model from `old_module` into a `new_module`? Previously you could install `new_module` in the pre script of `old_module`, then move a model into the `new_module` with `move_model()`. After this fix it's not possible. 

case (note: we are not merging the modules, just moving a model):
- have model `model.example` in `old_module`
- move model `model.example` to `new_module`
- pre-script in `old_module` to force install `new_module`
**option 1:**
- pre-script in `old_module` to move model from `old_module` to `new_module` fails since `new_module` is not yet installed
**option 2:**
- skip moving the model in pre-script
- `old_module` gets upgraded and the `model.example` table is dropped as it's not part of `old_module` anymore
- `new_module` is installed and the `model.example` table is recreated, but it's empty

miikanissi:
> > Previously you could install new_module in the pre script of old_module, then move a model into the new_module with move_model()
> 
> That was not totally correct. It "worked" but you shouldn't install a module in a pre- script of _another_ module. It should be done before `base` module is loaded. You can achieve that with `--pre-upgrade-scripts` parameter of `odoo-bin` --in case you get the DB from us after a major upgrade, since it will have a fully upgraded `base` already.
> 
> The solution for your issue may be to adapt what you were doing before to `force_install(cr, "new_module")` and `force_upgrade_of_fresh_module(cr, "new_module")`. Then _steal_ the model from `old_module` in a pre- script of `new_module`, if `new_module` is loaded first, otherwise do the move in `old_module`.
> 
> All _module_ operations --install, remove, merge, force upgrade, etc-- must be done before `base` is loaded. You should adapt any other code you have that doesn't respect this or you will risk having more issues like the one you described.

Thanks for the detailed explanation. This makes sense and I was aware it wasn't exactly best practice. For our situation the limitation has to do with odoo.sh outside of a major upgrade as there isn't an explicit way to call `--pre-upgrade-scripts`. Cheers!

aj-fuentes:
We still detected an issue that is being fixed in https://github.com/odoo/upgrade-util/pull/333

Thanks for reporting the issue.

PR #332, opened 2025-10-02
[FIX] util/fields: check new table first

aj-fuentes:
Some DBs may have still `ir_values` from previous upgrades. Check instead
for the presence of the new `ir_default` table.

upg-3172950


jjmaksoud (approved):

KangOl:

PR #333, opened 2025-10-06
[FIX] util/modules: trigger autodiscovery outside major upgrade

aj-fuentes:
Outside a major upgrade `base` is already fully loaded. New modules are yet not known for the ORM. We thus need to trigger the discovery.

During a major upgrade we also want to block force installing modules outside `base` pre/post.


src/util/modules.py
```diff
@@ -553,6 +555,15 @@ def force_install_module(cr, module, if_installed=None, reason="it has been expl
...
         if not if_installed or modules_installed(cr, *if_installed):
+            cr.execute("SELECT 1 FROM ir_module_module WHERE name = 'base' AND state != 'to upgrade'")
+            if cr.rowcount:  # we are outside a major upgrade
+                if not ENVIRON.get(
+                    "AUTO_DISCOVERY_RAN"
+                ):  # run the autodiscovery once to allow to force sintall _new_ modules
+                    _trigger_auto_discovery(cr)
+                elif ENVIRON.get("AUTO_DISCOVERY_UPGRADE"):  # in a major upgrade this is an error
+                    raise MigrationError("`force_install_module` can only be called from pre/post of `base`")
+                _force_install_module(cr, module, reason="it has been explicitly asked for outside a major upgrade")
```
KangOl:
```suggestion
                return _force_install_module(cr, module, reason="it has been explicitly asked for outside a major upgrade")
```

KangOl:
```suggestion
                return _force_install_module(cr, module, reason="{} (done outside of a major upgrade)".format(reason))
```

KangOl:

PR #343, opened 2025-10-20
[FIX] pg: rename m2m field meta data

jjmaksoud:
When renaming a model and updating related m2m fields, the relation table and its columns are renamed but only manual fields meta data is updated, assuming that base fields will be updated when the module loads.
The custom modules meta data should also be updated if the tables were updated.

upg-3176133

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -45,7 +45,7 @@ except ImportError:
 
 from .exceptions import MigrationError, SleepyDeveloperError
 from .helpers import _validate_table, model_of_table
-from .misc import AUTO, Sentinel, log_progress, on_CI, version_gte
+from .misc import AUTO, Sentinel, get_modules, log_progress, on_CI, version_gte
 
 _logger = logging.getLogger(__name__)
 
@@ -1619,6 +1619,7 @@ def update_m2m_tables(cr, old_table, new_table, ignored_m2ms=()):
     if old_table == new_table or not version_gte("10.0"):
         return
     ignored_m2ms = set(ignored_m2ms)
+    standard_modules = set(get_modules()) - {"studio_customization", "__cloc_exclude__"}
     for orig_m2m_table, _, _, other_table in get_m2m_on(cr, new_table):
         if orig_m2m_table in ignored_m2ms:
             continue
@@ -1640,12 +1641,18 @@ def update_m2m_tables(cr, old_table, new_table, ignored_m2ms=()):
             rename_table(cr, orig_m2m_table, m2m_table, remove_constraints=False)
             cr.execute(
                 """
-                UPDATE ir_model_fields
+                UPDATE ir_model_fields f
                    SET relation_table = %s
-                 WHERE relation_table = %s
-                   AND state = 'manual'
+                  FROM ir_model_data d
+                 WHERE f.relation_table = %s
+                   AND (
+                       f.state = 'manual'
+                    OR (d.model = 'ir.model.fields'
+                       AND d.res_id = f.id
+                       AND d.module NOT IN %s)
+                       )
                 """,
-                [m2m_table, orig_m2m_table],
+                [m2m_table, orig_m2m_table, tuple(standard_modules)],
             )
             _logger.info("Renamed m2m table %s to %s", orig_m2m_table, m2m_table)
         else:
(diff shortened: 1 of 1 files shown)


aj-fuentes (commented):
Better check if there isn't any xmlid for the field under the available modules (standard_modules usual trick)

src/util/pg.py
```diff
@@ -1474,12 +1477,18 @@ def update_m2m_tables(cr, old_table, new_table, ignored_m2ms=()):
...
-                 WHERE relation_table = %s
-                   AND state = 'manual'
+                  FROM ir_model_data d
+                 WHERE f.relation_table = %s
+                   AND (
+                       f.state = 'manual'
+                    OR d.model = 'ir.model.fields'
+                   AND d.res_id = f.id
+                   AND d.module NOT IN %s
+                       )
```
aj-fuentes:
```suggestion
                   AND (  f.state = 'manual'
                       OR (d.model = 'ir.model.fields'
                           AND d.res_id = f.id
                           AND d.module NOT IN %s)
                       )
```
Even with formatted this query looks hard to understand. Wouldn't it be better if we just do a cte?
```
WITH to_upd (
   SELECT f.id
     FROM ir_model_fields f
     JOIN ir_model_data d
       ON d.model = 'ir.model.fields'
      AND d.res_id = f.id
    WHERE f.relation_table = %s
       AND (  f.state = 'manual'
           OR d.module NOT IN %s)
) UPDATE ...
```

src/util/pg.py
```diff
@@ -36,9 +36,11 @@
 from psycopg2.extras import Json
 
 try:
+    from odoo.modules import get_modules
```
KangOl: Import it from `.misc`
```suggestion
```

src/util/pg.py
```diff
@@ -36,9 +36,11 @@
...
     from odoo.modules import module as odoo_module
     from odoo.sql_db import db_connect
 except ImportError:
+    from openerp.modules import get_modules
```
```suggestion
```

src/util/pg.py
```suggestion
from .misc import AUTO, Sentinel, get_modules, log_progress, on_CI, version_gte
```

src/util/pg.py
```diff
@@ -1594,12 +1597,18 @@ def update_m2m_tables(cr, old_table, new_table, ignored_m2ms=()):
...
                    SET relation_table = %s
-                 WHERE relation_table = %s
-                   AND state = 'manual'
+                  FROM ir_model_data d
+                 WHERE f.relation_table = %s
+                   AND (
+                       f.state = 'manual'
+                    OR d.model = 'ir.model.fields'
+                   AND d.res_id = f.id
+                   AND d.module NOT IN %s
```
KangOl: Force parentheses
```suggestion
                    OR (d.model = 'ir.model.fields'
                        AND d.res_id = f.id
                        AND d.module NOT IN %s)
```

KangOl (changes requested):

src/util/pg.py
```diff
@@ -1678,23 +1658,35 @@ def update_m2m_tables(cr, old_table, new_table, ignored_m2ms=()):
...
-                   AND column1 = %s
-                   AND state = 'manual'
+                  FROM ir_model_data d
+                 WHERE f.relation_table = %s
+                   AND f.column1 = %s
+                   AND (
+                       f.state = 'manual'
+                    OR d.model = 'ir.model.fields'
+                   AND d.res_id = f.id
+                   AND d.module NOT IN %s
```
KangOl, line 1670: Visibly, I had to be explicit.

src/util/pg.py
```diff
@@ -1719,12 +1720,18 @@ def update_m2m_tables(cr, old_table, new_table, ignored_m2ms=()):
...
-                 WHERE relation_table = %s
-                   AND state = 'manual'
+                  FROM ir_model_data d
+                 WHERE f.relation_table = %s
+                   AND (
+                       f.state = 'manual'
+                    OR (d.model = 'ir.model.fields'
+                       AND d.res_id = f.id
+                       AND d.module NOT IN %s)
+                       )
```
KangOl, line 1732: This will update row multiple times.
Better use a CTE.
```suggestion
                WITH _f AS (
                    SELECT f.id
                      FROM ir_model_fields f
                 LEFT JOIN ir_model_data d
                        ON d.model = 'ir.model.fields'
                       AND d.res_id = f.id
                       AND d.module NOT IN %s
                     WHERE f.relation_table = %s
                       AND (d.id IS NOT NULL OR f.state = 'manual')
                  GROUP BY f.id
                )
                UPDATE ir_model_fields
                   SET relation_table = %s
                  FROM _f
                 WHERE _f.if = ir_model_fields.id
```

PR #350, opened 2025-11-10
[IMP] orm: iter_browse accept generator or query as ids or values

cawo-odoo:
### [IMP] orm: iter_browse accept generator or query as ids

This allows the caller to be memory efficient on huge numbers of ids, allowing for even more millions of records to be browsed.

### [IMP] orm: iter_browse.create() accept generator or query as values

Done to be able to create millions of records memory-efficiently.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -701,6 +701,38 @@ class TestIterBrowse(UnitTestCase):
         expected = (len(ids) + chunk_size - 1) // chunk_size
         self.assertEqual(write.call_count, expected)
 
+    def test_iter_browse_ids_gen(self):
+        cr = self.env.cr
+        cr.execute("SELECT id FROM res_country")
+        ids = (c for (c,) in cr.fetchall())
+        ids_len = cr.rowcount
+        chunk_size = 10
+
+        res_chunks = list(
+            util.iter_browse(
+                self.env["res.country"], ids, size=ids_len, logger=None, chunk_size=chunk_size, yield_chunks=True
+            )
+        )
+        no_chunks = (ids_len + chunk_size - 1) // chunk_size
+        self.assertEqual(len(res_chunks), no_chunks)
+        self.assertEqual(len(res_chunks[0]), chunk_size)
+
+    def test_iter_browse_ids_query(self):
+        cr = self.env.cr
+        query = "SELECT id FROM res_country"
+        cr.execute(query)
+        ids_len = cr.rowcount
+        chunk_size = 10
+
+        res_chunks = list(
+            util.iter_browse(
+                self.env["res.country"], None, query=query, logger=None, chunk_size=chunk_size, yield_chunks=True
+            )
+        )
+        no_chunks = (ids_len + chunk_size - 1) // chunk_size
+        self.assertEqual(len(res_chunks), no_chunks)
+        self.assertEqual(len(res_chunks[0]), chunk_size)
+
     def test_iter_browse_create_non_empty(self):
         RP = self.env["res.partner"]
         with self.assertRaises(ValueError):
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -748,6 +748,33 @@ class TestIterBrowse(UnitTestCase):
         records = ib.create([{"name": name} for name in names], multi=multi)
         self.assertEqual([t.name for t in records], names)
 
+    @parametrize([(True,), (False,)])
+    def test_iter_browse_create_val_gen(self, multi):
+        chunk_size = 2
+        RP = self.env["res.partner"]
+
+        names = [f"Name {i}" for i in range(7)]
+        ib = util.iter_browse(RP, [], chunk_size=chunk_size)
+        records = ib.create(({"name": name} for name in names), size=7, multi=multi)
+        self.assertEqual([t.name for t in records], names)
+
+    @parametrize([(True,), (False,)])
+    def test_iter_browse_create_val_query(self, multi):
+        with db_connect(self.env.cr.dbname).cursor() as cr, mute_logger("odoo.sql_db"):
+            env = api.Environment(cr, SUPERUSER_ID, {})
+            chunk_size = 2
+            RP = env["res.partner"]
+
+            query = "SELECT 'Name ' || i AS name FROM GENERATE_SERIES(0, 6) AS i"
+            ib = util.iter_browse(RP, [], chunk_size=chunk_size)
+            records = ib.create(query=query, multi=multi)
+            names = []
+            for r in records:
+                names.append(r.name)
+                r.unlink()
+
+        self.assertEqual(names, [f"Name {i}" for i in range(7)])
+
     def test_iter_browse_iter_twice(self):
         cr = self.env.cr
         cr.execute("SELECT id FROM res_country")
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -381,7 +381,11 @@ class iter_browse(object):
 
     :param model: the model to iterate
     :type model: :class:`odoo.model.Model`
-    :param list(int) ids: list of IDs of the records to iterate
+    :param iterable(int) ids: iterable of IDs of the records to iterate
+    :param int size: the length of `ids`. Can be used to enable progress logging when `ids` has no `__length__`
+    :param str query: alternative to ids, SQL query that can produce them.
+                      Can also be a DML statement with a RETURNING clause.
+                      See :func:`~odoo.upgrade.util.pg.query_ids`
     :param int chunk_size: number of records to load in each iteration chunk, `200` by
                            default
     :param bool yield_chunks: when iterating, yield records in chunks of `chunk_size` instead of one by one.
@@ -403,7 +407,8 @@ class iter_browse(object):
         self._model = model
         self._cr_uid = args[:-1]
         ids = args[-1]
-        self._size = len(ids)
+        self._size = kw.pop("size", None)
+        query = kw.pop("query", None)
         self._chunk_size = kw.pop("chunk_size", 200)  # keyword-only argument
         self._yield_chunks = kw.pop("yield_chunks", False)
         self._logger = kw.pop("logger", _logger)
@@ -411,6 +416,18 @@ class iter_browse(object):
         assert self._strategy in {"flush", "commit"}
         if kw:
             raise TypeError("Unknown arguments: %s" % ", ".join(kw))
+
+        if not (ids is None) ^ (query is None):
+            raise TypeError("Must be initialized using exactly one of `ids` or `query`")
+
+        if query:
+            ids = query_ids(self._model.env.cr, query, itersize=self._chunk_size)
+
+        if not self._size:
+            try:  # noqa: SIM105
+                self._size = len(ids)
+            except TypeError:
+                pass
 
         self._patch = None
         self._it = chunks(ids, self._chunk_size, fmt=self._browse)
@@ -440,9 +457,14 @@ class iter_browse(object):
             raise RuntimeError("%r ran twice" % (self,))
 
         it = self._it if self._yield_chunks else chain.from_iterable(self._it)
-        sz = (self._size + self._chunk_size - 1) // self._chunk_size if self._yield_chunks else self._size
         if self._logger:
-            it = log_progress(it, self._logger, qualifier=self._model._name, size=sz)
+            sz = (
+                (self._size + self._chunk_size - 1) // self._chunk_size
+                if self._size and self._yield_chunks
+                else self._size
+            )
+            qualifier = self._model._name + ("[:{}]".format(self._chunk_size) if self._yield_chunks else "")
+            it = log_progress(it, self._logger, qualifier=qualifier, size=sz)
         self._it = None
         return chain(it, self._end())
 
@@ -455,8 +477,8 @@ class iter_browse(object):
 
         it = self._it
         if self._logger:
-            sz = (self._size + self._chunk_size - 1) // self._chunk_size
-            qualifier = "%s[:%d]" % (self._model._name, self._chunk_size)
+            sz = (self._size + self._chunk_size - 1) // self._chunk_size if self._size else None
+            qualifier = "{}[:{}]".format(self._model._name, self._chunk_size)
             it = log_progress(it, self._logger, qualifier=qualifier, size=sz)
 
         def caller(*args, **kwargs):
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -42,7 +42,7 @@ from .const import BIG_TABLE_THRESHOLD
 from .exceptions import MigrationError
 from .helpers import table_of_model
 from .misc import chunks, log_progress, version_between, version_gte
-from .pg import SQLStr, column_exists, format_query, get_columns, query_ids
+from .pg import SQLStr, column_exists, format_query, get_columns, named_cursor, query_ids
 
 # python3 shims
 try:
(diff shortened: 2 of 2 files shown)


cawo-odoo:
rebased on https://github.com/odoo/upgrade-util/pull/355 to also make use of `query_ids()` here

cawo-odoo:
added / modified tests to cover these changes

src/util/pg.py
```diff
@@ -1932,3 +1944,85 @@ def bulk_update_table(cr, table, columns, mapping, key_col="id"):
...
+    """
+
+    def __init__(self, cr, query, itersize=None):
+        self._ncr = None
+        self._cr = cr
+        self._tmp_tbl = "_upgrade_query_ids_{}".format(uuid.uuid4().hex)
+        cr.execute(
+            format_query(
+                cr,
+                "CREATE UNLOGGED TABLE {}(id) AS (WITH query AS ({}) SELECT * FROM query)",
```
Pirols, line 1971:
```suggestion
                "CREATE UNLOGGED TABLE {}(id) AS {}",
```

What do we need the cte for here? 🤔

Pirols: Or perhaps the goal is to only fetch ids?
It'd make sense, note however that based on the query, the current implementation would not do that:

```SQL
odoo@/run/postgresql:test> CREATE UNLOGGED TABLE t(id) AS SELECT * FROM (VALUES (1, 2), (2, 3)) AS query (id, other_col) ; TABLE t
SELECT 2
+----+-----------+
| id | other_col |
|----+-----------|
| 1  | 2         |
| 2  | 3         |
+----+-----------+
```

Instead we should do something like:

```suggestion
                "CREATE UNLOGGED TABLE {}(id) AS (WITH query AS ({}) SELECT id FROM query)",
```

Or, simpler imo:

```python
cr.execute("CREATE UNLOGGED TABLE {table_name} (id INTEGER UNIQUE NOT NULL)")
try:
    cr.execute("INSERT INTO {table_name} {select_query}")
except ...
```

src/base/tests/test_util.py
```diff
@@ -1495,6 +1525,27 @@ def test_iter(self):
...
+        self.assertEqual(result, [[1, 2]])
+
+    def test_destructor(self):
+        ids = util.query_ids(self.env.cr, "SELECT id from res_users")
+        del ids
+
+    def test_pk_violation(self):
+        with db_connect(self.env.cr.dbname).cursor() as cr, mute_logger("odoo.sql_db"), self.assertRaises(
+            ValueError
+        ), util.query_ids(cr, "SELECT * FROM (VALUES (1), (1)) AS x(x)") as ids:
```
Pirols, line 1586:
```suggestion
        ), util.query_ids(cr, "SELECT * FROM (VALUES (1), (1), (NULL)) AS x(x)") as ids:
```

NULL values are also not welcome in a pkey.

Pirols (commented):
On 7317015c612b7329c3c6fe320e1d0afab8872088

src/base/tests/test_util.py
cawo-odoo: Note that this comment really belongs to https://github.com/odoo/upgrade-util/pull/355

src/util/pg.py
cawo-odoo: Also https://github.com/odoo/upgrade-util/pull/355

cawo-odoo: > What do we need the cte for here? 🤔

I see you did not "move" this question to the other PR, but to answer it: The CTE was actually introduced on the already merged PR https://github.com/odoo/upgrade-util/pull/322 and it was done to handle this: https://runbot.odoo.com/runbot/build/89536248

Pirols: Huh, I see. Thanks :+1:

cawo-odoo: > Instead we should do something like: `"CREATE UNLOGGED TABLE {}(id) AS (WITH query AS ({}) SELECT id FROM query)"`

The issue is that we do not want to force the column name. E.g. https://runbot.odoo.com/runbot/build/89536248 should work.

Pirols: > > Instead we should do something like: `"CREATE UNLOGGED TABLE {}(id) AS (WITH query AS ({}) SELECT id FROM query)"`
> 
> The issue is that we do not want to force the column name. E.g. https://runbot.odoo.com/runbot/build/89536248 should work.

How about my second proposal then?

```
cr.execute("CREATE UNLOGGED TABLE {table_name} (id INTEGER PRIMARY KEY)")  # or UNIQUE NOT NULL
try:
    cr.execute("INSERT INTO {table_name} {select_query}")
except ...
```

It will work whatever the name of the column, so long as it's only one (integer) column.

cawo-odoo: > It will work whatever the name of the column, so long as it's only one (integer) column.

True. The reason for using `CREATE TABLE AS` is performance, *because* no constraints will be involved. Considering the sole reason to have this util is for cases with many millions of ids, we want it to be as fast as possible.

cawo-odoo: Maybe we could use `information_schema` after the fact to check this programmatically. OTOH, the name of the function alone should be clear enough. I'm not sure if it is worth running an additional check every time this is called for something that should only happen at development time (of an upgrade script)  :man_shrugging:

Pirols: > > It will work whatever the name of the column, so long as it's only one (integer) column.
> 
> True. The reason for using `CREATE TABLE AS` is performance, _because_ no constraints will be involved. Considering the sole reason to have this util is for cases with many millions of ids, we want it to be as fast as possible.

Indeed, it's nearly twice as fast (0.392s vs 0.761) for one million elements. The difference is even bigger at 10 millions.

```SQL
odoo@/run/postgresql:test> CREATE UNLOGGED TABLE t (id) AS SELECT * FROM generate_series(1, 1000000)
SELECT 1000000
Time: 0.217s
odoo@/run/postgresql:test> ALTER TABLE t ADD CONSTRAINT t_pk PRIMARY KEY (id)
You're about to run a destructive command.
Do you want to proceed? [y/N]: y
Your call!
ALTER TABLE
Time: 0.175s
odoo@/run/postgresql:test> CREATE UNLOGGED TABLE t_insert (id INTEGER PRIMARY KEY)
CREATE TABLE
Time: 0.012s
odoo@/run/postgresql:test> INSERT INTO t_insert SELECT * FROM generate_series(1, 1000000)
INSERT 0 1000000
Time: 0.749s
```

> I'm not sure if it is worth running an additional check every time this is called for something that should only happen at development time (of an upgrade script) 🤷‍♂️

It would be fairly cheap to check whether the new table has more than one column, but if it's not critical we can probably skip the validation.

aj-fuentes: Please move this discussion to #355

~~EDIT: Posted a stale comment meant for another PR 🤦🏼~~
EDIT2: well actually is 355 :/ these too many already rebased PRs are a hell to review.

cawo-odoo:
Rebased on master after https://github.com/odoo/upgrade-util/pull/355 has been merged. Should now be nicer to review.

src/util/orm.py
```diff
@@ -393,8 +408,32 @@ def __init__(self, model, *args, **kw):
...
+        self._it = chunks(self._ids, self._chunk_size, fmt=self._browse)
+
+    def _values_query(self, query):
+        cr = self._model.env.cr
+        cr.execute(format_query(cr, "WITH query AS ({}) SELECT count(*) FROM query", SQLStr(query)))
+        size = cr.fetchone()[0]
+
+        def get_values():
+            with named_cursor(cr, itersize=self._chunk_size) as ncr:
+                ncr.execute(SQLStr(query))
```
aj-fuentes: Plain `query` works here.
```suggestion
                ncr.execute(query)
```

src/util/orm.py
```diff
@@ -393,8 +408,32 @@ def __init__(self, model, *args, **kw):
...
+        if not self._size:
+            try:
+                self._size = len(self._ids)
+            except TypeError:
+                raise ValueError("When passing ids as a generator, the size kwarg is mandatory")
         self._patch = None
-        self._it = chunks(ids, self._chunk_size, fmt=self._browse)
+        self._it = chunks(self._ids, self._chunk_size, fmt=self._browse)
+
+    def _values_query(self, query):
```
aj-fuentes: The name of the utility feels odd to me (subjective)
```suggestion
    def _values_iter(self, query):
        # Returns the size of query output and an iterator over the query output.
```

aj-fuentes (commented):
This could be merged.

src/util/orm.py
```diff
@@ -362,7 +362,9 @@ class iter_browse(object):
 
     :param model: the model to iterate
     :type model: :class:`odoo.model.Model`
-    :param list(int) ids: list of IDs of the records to iterate
+    :param iterable(int) ids: iterable of IDs of the records to iterate
+    :param str query: alternative to ids, SQL query that can produce them.
+                      Can also be a DML statement with a RETURNING clause.
```
aj-fuentes, line 387: This could help understanding the parameter.
```suggestion
                      Can also be a DML statement with a RETURNING clause.
                      See :func:`~odoo.upgrade.util.pg.query_ids`
```

src/util/orm.py
```diff
@@ -396,14 +399,27 @@ class iter_browse(object):
...
 
     def __init__(self, model, *args, **kw):
         assert len(args) in [1, 3]  # either (cr, uid, ids) or (ids,)
         self._model = model
         self._cr_uid = args[:-1]
-        ids = args[-1]
-        self._size = len(ids)
+        self._ids = args[-1]
+        self._size = kw.pop("size", None)
+        self._query = kw.pop("query", None)
```
KangOl: `_ids` and `_query` doesn't need to be exposed on `self` as they won't be used outside `__init__`
```suggestion
        ids = args[-1]
        self._size = kw.pop("size", None)
        query = kw.pop("query", None)
```

src/util/orm.py
```diff
@@ -412,8 +428,32 @@ def __init__(self, model, *args, **kw):
...
-        self._it = chunks(ids, self._chunk_size, fmt=self._browse)
+        self._it = chunks(self._ids, self._chunk_size, fmt=self._browse)
+
+    def _values_iter(self, query):
+        cr = self._model.env.cr
+        cr.execute(format_query(cr, "WITH query AS ({}) SELECT count(*) FROM query", SQLStr(query)))
+        size = cr.fetchone()[0]
+
+        def get_values():
+            with named_cursor(cr, itersize=self._chunk_size) as ncr:
```
KangOl: It's not OK to execute the query twice as it may be not reentrant.

```sql
   UPDATE foo
      SET bar = "default"
    WHERE bar IS NULL
RETURNING id as foo_id
```

src/base/tests/test_util.py
```diff
@@ -677,28 +677,32 @@ def test_iter_browse_iter(self):
     def test_iter_browse_iter_chunks(self):
         cr = self.env.cr
         cr.execute("SELECT id FROM res_country")
-        ids = [c for (c,) in cr.fetchall()]
+        count = cr.rowcount
+        ids = (c for (c,) in cr.fetchall())  # ids as generator
```
KangOl: Instead of hikacking existing tests, I would prefer to have dedicated unit tests to each usage of `iter_browse` (sequence/iterator/query), like for the `create` part.

KangOl (changes requested):

src/util/orm.py
cawo-odoo: This is why I documented this to not be allowed:
https://github.com/odoo/upgrade-util/pull/350/changes#diff-e83e04eeae2ce53657ba28fd8de2126f5f3ef6422a73066af36dcf9191de1278R519

EDIT: I remember there was an earlier  discussion about this and whether a similar temporary table should be used as for the ids query. The conclusion was that it is more complicated than for `ids`, because the query might return multiple columns of different type, it could be much more data and so on, and we should postpone this idea until there is an actual use-case.

KangOl: We can use a temp table.
```sql
WITH query AS ({}) SELECT * INTO UNLOGGED TABLE _upg_blah FROM query;
```
Then we just have the size with `cr.rowcount`.

cawo-odoo: Yes, I know in principle we could apply the same idea as for `query_ids` to support data modifying queries. For those there's no other alternative, so there's no question if that would be the best way for *those*. But if we did, we need to consider the effect on the probably way more common use-case of a `SELECT` query and that would be that the data returned by it will now be written *twice* (once to the temp table and once to the model table of the actual record creation). As this query returns values, it means potentially large data in which case the effect on this case would be detrimental. The big difference to the `query_ids`case is that for `query_ids`we know the return will only be one integer column (small).
All in all I think if we want to support RETURNING queries here, it must be *in alternative* to SELECT, meaning we would need to have code for both strategies (count vs. temp table) and pick strategy based on what kind of query it is. And before we actually see a case where this would be helpful, I'm reluctant to add this complexity. To me this is just a case of YAGNI (until it isn't).

aj-fuentes: I prefer the simpler approach as well.

We can do something like:
```py
if (bad_mod := next((mod for mod in ["update", "delete", "insert"] if mod in query.lower()), None)):
    raise SleepyDeveloper("Wake up: this looks like a DML query" + bad_mod)
```

In the end we always need to review the query. If the caller needs some DML stuff it's them who need to setup a temp table and then pass the select over that table to this util.

KangOl: It's part of the *public* API. I don't like leaving such a footgun in the code.

However, I may have a solution that needs to be confirmed. Stay tuned.

src/util/orm.py
```diff
@@ -412,9 +416,33 @@ def __init__(self, model, *args, **kw):
...
+                self._size = len(ids)
+            except TypeError:
+                raise ValueError("When passing ids as a generator, the size kwarg is mandatory")
         self._patch = None
         self._it = chunks(ids, self._chunk_size, fmt=self._browse)
 
+    def _values_iter(self, query):
+        cr = self._model.env.cr
+        cr.execute(format_query(cr, "WITH query AS ({}) SELECT count(*) FROM query", SQLStr(query)))
+        size = cr.fetchone()[0]
```
KangOl: We can ensure no DML are done by using a read-only transaction.
```suggestion
        cr.commit()  # starts a new transaction
        cr._cnx.readonly = True  # in readonly
        cr.execute(format_query(cr, "WITH query AS ({}) SELECT count(*) FROM query", SQLStr(query)))
        size = cr.fetchone()[0]
        cr.rollback()
        cr._cnx.readonly = False  # back to normal
```

cawo-odoo: Good idea! I changed it to make use of https://www.psycopg.org/docs/connection.html#connection.set_session and just a single comment.

KangOl: `.readonly` actually call `set_session` under the hood, as shown by the error I got when I forgot to get commit/rollback the transaction.
```
In [23]: env.cr._cnx.readonly = False
---------------------------------------------------------------------------
ProgrammingError                          Traceback (most recent call last)
Cell In[23], line 1
----> 1 env.cr._cnx.readonly = False

ProgrammingError: set_session cannot be used inside a transaction
```

src/base/tests/test_util.py
```diff
@@ -716,6 +748,29 @@ def test_iter_browse_create(self, multi):
...
+        with db_connect(self.env.cr.dbname).cursor() as cr, mute_logger("odoo.sql_db"):
+            env = api.Environment(cr, SUPERUSER_ID, {})
+            chunk_size = 2
+            RP = env["res.partner"]
+
+            names = [f"Name {i}" for i in range(7)]
+            query = "SELECT 'Name ' || i AS name FROM GENERATE_SERIES(0, 6) AS i"
+            ib = util.iter_browse(RP, [], chunk_size=chunk_size)
+            records = ib.create(query=query, multi=multi)
+            self.assertEqual([t.name for t in records], names)
```
```suggestion
            query = "SELECT 'Name ' || i AS name FROM GENERATE_SERIES(0, 6) AS i"
            ib = util.iter_browse(RP, [], chunk_size=chunk_size)
            records = ib.create(query=query, multi=multi)
            names = [t.name for t in records]
            # cleanUp
            records.unlink()
            
            self.assertEqual(names, [f"Name {i}" for i in range(7)])
```

cawo-odoo: applied it modified to only iterate the returned object once.

cawo-odoo:
All suggestions are taken. The remaining matt error seems unrelated (or rather caused by https://github.com/odoo/odoo/commit/9fca3bea4427f566ff0a33fdb6f20ab73f60e44f)

aj-fuentes:
You need to rebase this to get the fix https://github.com/odoo/upgrade-util/commit/962a5038d14fbdb010e0a624ad0df633bb800e5d

cawo-odoo:
> You need to rebase this to get the fix [962a503](https://github.com/odoo/upgrade-util/commit/962a5038d14fbdb010e0a624ad0df633bb800e5d)

Ah. I thought I did, but it fetched the wrong remote. :man_facepalming: Thanks

src/util/orm.py
```diff
@@ -466,35 +499,47 @@ def caller(*args, **kwargs):
...
-        ids = []
-        size = len(values)
+        if query:
+            size, values = self._values_iter(query)
+
+        if size is None:
+            try:
+                size = len(values)
+            except TypeError:
+                raise ValueError("When passing a generator of values, the size kwarg is mandatory")
```
aj-fuentes: Can we rethink this? Why do we insist in making `size` mandatory? Do we need it other than for logging? It's odd to run a query twice, including some connection tweaking just for this :/ Can't we just log _differently_?

Or just make it clear here with a comment why we need `size` in other place than logging.

KangOl: > Can't we just log _differently_?

Good idea: https://github.com/odoo/upgrade-util/pull/440

cawo-odoo: > Good idea: #440

Ah. I just did 80% of the same. OK, let's use yours.

src/util/orm.py
```diff
@@ -466,35 +488,55 @@ def caller(*args, **kwargs):
...
+        size = kw.pop("size", None)
         if kw:
             raise TypeError("Unknown arguments: %s" % ", ".join(kw))
 
-        if not values:
-            raise ValueError("`create` cannot be called with an empty `values` argument")
+        if not (values is None) ^ (query is None):
+            raise ValueError("`create` needs to be called using exactly one of `values` or `query` arguments")
 
         if self._size:
```
```suggestion
        if self._size != 0:
```

cawo-odoo: Yes, I was to comment about that this condition is no longer sufficient. I did not think that a `None` value is necessarily invalid for this use, but OK, I guess we can force the caller to pass in an empty iterable that has a `__length__`. I'll also adapt the error message to make it more clear.

KangOl (approved):
@robodoo rebase-ff r+

PR #360, opened 2025-12-01
[ADD] util/report: add new reporting function

diagnoza:
Add a new function that replaces `util.add_to_migration_reports`.

The new function can automatically parse the provided data as a list of records. It can also limit the number of displayed records to prevent the upgrade report from growing too large.

The structure of the note is now defined within the function itself, standardizing its appearance.

Example usage
===============
More examples will follow in `odoo/upgrade`, replacing the existing use of `util.add_to_migration_reports`.

Example 1
---------------
```
cr.execute(
        """
        SELECT p.id, p.name, p.city, p.comment, p.company_id, com.name FROM res_partner p
          JOIN res_company com
            ON p.company_id = com.id
        """
    )

summary = """
    This is a test of the utility method that is displaying a list of rows
    returned by an SQL query. The number of displayed rows is limited.
    """
row_format = "Partner with id {partner_link} works at company {company_link} in {city}, ({comment})"
columns = ("id", "name", "city", "comment", "company_id", "company_name")
links = {"company_link": ("res.company", "company_id", "company_name"), "partner_link": ("res.partner", "id", "name")}
util.add_report(header="Test of the new utility method", summary=summary, row_format=row_format, columns=columns, links=links, data=cr.fetchall())
```
image

Example 2
---------------
```
util.add_report(header="Simple note", summary="A detailed description of the change.")
```
image

Example 3
---------------
```
msg = """
This is some text in **bold**.  

This is a line  
that should break.

- Item one
- Item two
    - Sub-item

And some _italic_ text.
"""
util.add_report(header="MD note", summary=msg, report_format="md")
```
image

Example 4
---------------
```
msg = """
This is some text in **bold**.

| This is a line
| that should break.

* Item one
* Item two

  * Sub-item

And some *italic* text.
"""
util.add_report(header="RST note", summary=msg, report_format="rst")
```
image

Remaining details to discuss
===============
1. The introduction of env var to control `limit`.
2. Enforce/assert the use of tuples instead of lists. Currently, it's possible to pass a list for `columns` (instead of a tuple) or a dictionary of lists for `links` (instead of a dictionary of tuples).
3. How to handle depreciation of `add_to_migration_reports`.
4.  Indentation of the `ir.ui.view` record corresponding to the note is sometimes incorrect (previously there was the same issue):
image

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -3064,3 +3064,104 @@ class TestAssertUpdated(UnitTestCase):
         ):
             p2.city = "Niflheim"
             util.flush(p2)
+
+
+class TestReportUtils(UnitTestCase):
+    def test_report_simple(self):
+        self.assertEqual(
+            util.report_with_list(
+                summary="Simple test with minimal arguments and no data.",
+                data=[],
+                columns=("id", "name"),
+                row_format="Partner {name} has id {id}.",
+            ),
+            "<summary>Simple test with minimal arguments and no data.</summary>",
+        )
+
+    def test_report_links(self):
+        self.assertEqual(
+            util.report_with_list(
+                summary="Testing links.",
+                data=[],
+                columns=("id", "name"),
+                row_format="Partner {partner_link}.",
+                links={"partner_link": ("res.partner", "id", "name")},
+            ),
+            "<summary>Testing links.</summary>",
+        )
+
+    def test_report_data_minimal(self):
+        href = (
+            "/odoo/res.partner/1?debug=1"
+            if util.version_gte("18.0")
+            else "/web?debug=1#view_type=form&amp;model=res.partner&amp;action=&amp;id=1"
+        )
+        expected = (
+            "<summary>Test with minimal data.<details><i></i><ul>\n"
+            '<li>Partner <a target="_blank" href="{}">Partner One</a>.</li>\n'
+            "</ul></details></summary>"
+        ).format(href)
+        self.assertEqual(
+            util.report_with_list(
+                summary="Test with minimal data.",
+                data=[(1, "Partner One")],
+                columns=("id", "name"),
+                row_format="Partner {partner_link}.",
+                links={"partner_link": ("res.partner", "id", "name")},
+            ),
+            expected,
+        )
+
+    def test_report_data_limited(self):
+        if util.version_gte("18.0"):
+            href1 = "/odoo/res.partner/1?debug=1"
+            href2 = "/odoo/res.partner/2?debug=1"
+        else:
+            href1 = "/web?debug=1#view_type=form&amp;model=res.partner&amp;action=&amp;id=1"
+            href2 = "/web?debug=1#view_type=form&amp;model=res.partner&amp;action=&amp;id=2"
+        expected = (
+            "<summary>Test with limited data.<details><i>The total number of affected records is 3. This list is showing the first 2 records.</i><ul>\n"
+            '<li>Partner <a target="_blank" href="{}">Partner One</a>.</li>\n'
+            '<li>Partner <a target="_blank" href="{}">Partner Two</a>.</li>\n'
+            "</ul></details></summary>"
+        ).format(href1, href2)
+        self.assertEqual(
+            util.report_with_list(
+                summary="Test with limited data.",
+                data=[(1, "Partner One"), (2, "Partner Two"), (3, "Partner Three")],
+                columns=("id", "name"),
+                row_format="Partner {partner_link}.",
+                links={"partner_link": ("res.partner", "id", "name")},
+                total=3,
+                limit=2,
+            ),
+            expected,
+        )
+
+    def test_report_data_limitless(self):
+        if util.version_gte("18.0"):
+            href1 = "/odoo/res.partner/1?debug=1"
+            href2 = "/odoo/res.partner/2?debug=1"
+            href3 = "/odoo/res.partner/3?debug=1"
+        else:
+            href1 = "/web?debug=1#view_type=form&amp;model=res.partner&amp;action=&amp;id=1"
+            href2 = "/web?debug=1#view_type=form&amp;model=res.partner&amp;action=&amp;id=2"
+            href3 = "/web?debug=1#view_type=form&amp;model=res.partner&amp;action=&amp;id=3"
+        expected = (
+            "<summary>Test with limitless data.<details><i></i><ul>\n"
+            '<li>Partner <a target="_blank" href="{}">Partner One</a>.</li>\n'
+            '<li>Partner <a target="_blank" href="{}">Partner Two</a>.</li>\n'
+            '<li>Partner <a target="_blank" href="{}">Partner Three</a>.</li>\n'
+            "</ul></details></summary>"
+        ).format(href1, href2, href3)
+        self.assertEqual(
+            util.report_with_list(
+                summary="Test with limitless data.",
+                data=[(1, "Partner One"), (2, "Partner Two"), (3, "Partner Three")],
+                columns=("id", "name"),
+                row_format="Partner {partner_link}.",
+                links={"partner_link": ("res.partner", "id", "name")},
+                limit=None,
+            ),
+            expected,
+        )
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -73,7 +73,7 @@ from .pg import (
     target_of,
 )
 from .records import _remove_import_export_paths, ref
-from .report import add_to_migration_reports, get_anchor_link_to_record
+from .report import add_to_migration_reports, report_with_list
 
 # python3 shims
 try:
@@ -1598,11 +1598,26 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
                )
     """.format(col_prefix=col_prefix, name=get_value_or_en_translation(cr, "ir_act_server", "name"))
     cr.execute(q, {"old_pattern": p["old_pattern"], "old": p["old"], "standard_modules": tuple(standard_modules)})
-    li = ""
-    if cr.rowcount:
-        li = "".join(
-            "<li>{}</li>".format(get_anchor_link_to_record("ir.actions.server", aid, aname))
-            for aid, aname in cr.fetchall()
+
+    model_text = "All models"
+    if only_models:
+        model_text = "Models " + ", ".join("<kbd>{}</kbd>".format(m) for m in only_models)
+    total = cr.rowcount
+    if total:
+        summary = """
+            {model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>.
+            The following server actions may need an update.
+            If a server action is a standard one and you haven't made any modifications, you may ignore it.
+        """.format(model_text=model_text, old=old, new=new)
+        report_with_list(
+            summary=summary,
+            data=cr.fetchall(),
+            columns=("id", "name"),
+            row_format="{action_link}",
+            links={"action_link": ("ir.actions.server", "id", "name")},
+            total=total,
+            limit=20,
+            category="Fields renamed",
         )
 
     if column_exists(cr, "ir_model_fields", "compute"):
@@ -1621,28 +1636,24 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
                    )
             """
         cr.execute(q, {"old_pattern": p["old_pattern"], "standard_modules": tuple(standard_modules)})
-        if cr.rowcount:
-            li += "".join(
-                "<li>{}</li>".format(get_anchor_link_to_record("ir.model.fields", fid, fname))
-                for fid, fname in cr.fetchall()
+
+        total = cr.rowcount
+        if total:
+            summary = """
+                {model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>.
+                The following compute methods of other fields may need an update.
+                If a field is a standard one and you haven't made any modifications, you may ignore it.
+            """.format(model_text=model_text, old=old, new=new)
+            report_with_list(
+                summary=summary,
+                data=cr.fetchall(),
+                columns=("id", "name"),
+                row_format="{field_link}",
+                links={"field_link": ("ir.model.fields", "id", "name")},
+                total=total,
+                limit=20,
+                category="Fields renamed",
             )
-    if li:
-        model_text = "All models"
-        if only_models:
-            model_text = "Models " + ", ".join("<kbd>{}</kbd>".format(m) for m in only_models)
-        add_to_migration_reports(
-            """
-<details>
-  <summary>
-    {model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>. The following server actions and compute methods of other fields may need an update.
-    If a server action or a field is a standard one and you haven't made any modifications, you may ignore them.
-  </summary>
-  <ul>{li}</ul>
-</details>
-            """.format(**locals()),
-            category="Fields renamed",
-            format="html",
-        )
 
     # if we stay on the same model. (no usage of dotted-path) (only works for domains and related)
     if "." not in old and "." not in new:
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -107,7 +107,7 @@ ODOO_SHOWCASE_VIDEOS = {
 }
 
 
-def add_to_migration_reports(message, category="Other", format="text"):
+def report_message(message, category="Other", format="text"):
     assert format in {"text", "html", "md", "rst"}
     if format == "md":
         message = md2html(dedent(message))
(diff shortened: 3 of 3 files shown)


src/util/report.py
```diff
@@ -126,6 +126,95 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
 
+def add_report(
+    header,
+    summary,
+    data=None,
+    columns=None,
+    links=None,
+    row_format=None,
+    limit=100,
+    report_format="html",
```
aj-fuentes: This util should be always for `html`
```suggestion
```

src/util/report.py
```diff
@@ -126,6 +126,95 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+):
```
aj-fuentes: I think we should split this. Have a simpler util to generate the summary/details part, then reuse it. It's a simple thing, but we keep getting it wrong and it is tedious to review it each time.
I take the opportunity to suggest we simplify (alias) `add_to_migration_reports` into something like `report`. @KangOl may chime in here with better suggestions.
```py
def report_with_summary(cr, summary, details, category="Other"):
    msg = "<summary>{}<details>{}</details></summary>".format(summary, details)
    util.report(cr, message=msg, category=category, format="html")
```

Then keep the main part here. The report is always html.
```suggestion
def report_records_list(summary, data, columns, row_format, links=None, limit=100, total=None, category="Other"):
```
With this interface the implementation is more direct. Callers should always provide the `data`, `columns`, and `row_format`. As done here already, `links` will act as an extra set of variables that can be used in `row_format` along with `columns`.

EDIT: added the `total` parameter, the point is to avoid memory errors, we cannot force the caller to provide a whole list of records.

src/util/report.py
```diff
@@ -126,6 +126,95 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+    :param str row_format: The way a row in a list is formatted, using named placeholders, e.g.:
+                           "Partner {partner_link} that lives in {city} works at company {company_link}."
+    :param int limit: The number of records that are going to be displayed in the report.
+    :param str report_format: Format of the input "summary"; must be one of {"text", "html", "md", "rst"}
+    """
+    assert report_format in {"text", "html", "md", "rst"}
+    if report_format == "md":
+        summary = md2html(dedent(summary))
+    elif report_format == "rst":
+        summary = rst2html(dedent(summary))
```
aj-fuentes: Always html
```suggestion
```

src/util/report.py
```diff
@@ -126,6 +126,95 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+                    row_dict[link_name] = get_anchor_link_to_record(
+                        link_model,
+                        row_dict[id_column],
+                        row_dict[name_column],
+                    )
+
+            row_text = row_format.format(**row_dict)
+            rows.append("<li>{}</li>".format(row_text))
+
+        rows = "<ul>\n" + "\n".join(rows) + "\n</ul>"
```
aj-fuentes: We should validate always that the provided format works. We cannot leave this for when `data` is non-empty otherwise we risk failing in production but not in CI.
It's better if we refactor the row generation (as `<li>` items) into its own function.

General idea (using simpler variable names to keep short --still very readable):
```suggestion
    def row_li(row):
        row_dict = dict(zip(columns, row))
        row_dict.update({
            link: get_anchor_link_to_record(rec_model, row_dict[id_col], row_dict[name_col]) 
            for link, (rec_model, id_col, name_col) in links.items()
         })
         return "<li>{}</li>".format(row_format.format(**row_dict))
    if not data:
        row_li(columns) # validate the format is correct, including links
        return report_with_summary(cr, summary=summary, details="No records affected", category=category)
    rows = "<ul>\n{}\n</ul>".format([row_li(row) for row in data[:limit]])
```

src/util/report.py
```diff
@@ -126,6 +126,95 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+        </summary>
+        """.format(summary)
+
+    raw = False
+    if report_format != "text":
+        if Markup:
+            msg = Markup(msg)
+        else:
+            raw = True
+    migration_reports.setdefault(header, []).append((msg, raw))
```
aj-fuentes: Since we have a shortcut for when `data` is empty we can simplify this.
```suggestion
    total = len(data)
    disclaimer = "The total number of affected records is {}.".format(total)
    if total > limit:
        disclaimer += " This list is limited to {} records.".format(limit)
    msg = "<summary>{}<details><i>{}</i>{}</details></summary>".format(summary, disclaimer, rows)
    add_to_migration_reports(cr, message=msg, format="html", category=category)
```

aj-fuentes (commented):
I think we can do some simplification here. Some suggestions below. Note that none of the suggestions were tested, take it as a guide.

src/util/report.py
```diff
@@ -126,6 +126,95 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+                        row_dict[id_column],
+                        row_dict[name_column],
+                    )
+
+            row_text = row_format.format(**row_dict)
+            rows.append("<li>{}</li>".format(row_text))
+
+        rows = "<ul>\n" + "\n".join(rows) + "\n</ul>"
+
+        total = len(data)
```
aj-fuentes:
```suggestion
        total = len(data) if total is None else total
```

src/util/report.py
diagnoza: Should we explicitly indicate somewhere in the code that this part is added only so that our internal CI can correctly run the test?

aj-fuentes: I already put a comment in my suggestion.
> `# validate the format is correct, including links`

src/util/report.py
```diff
@@ -126,6 +126,66 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+    :param str category: Title of a report entry.
+    :param str summary: Description of a report entry.
+    :param List[Tuple] data: Any data in the form of a list of tuples, e.g. the output of cr.fetchall().
+    :param Tuple[str] columns: Arbitrary names for each column in "data". All columns must be named and
+                              the order of these names must be the same as in "data".
+    :param Dict[str, Tuple[str, str, str]] links: Optional model/record links, e.g.:
+                                                  {
+                                                    "partner_link": ("res.partner", "id", "name"),
+                                                    "company_link": ("res.company", "company_id", "company_name"),
+                                                  }
```
aj-fuentes: * Do not use typing format, follow what's done elsewhere in this repo and use the plain type built-ins
* Shorten this and move all examples to a proper example block in the documentation description before the parameters:
```
.. example::

  .. code-block:: python

    # code example here
```
* Always ensure everything renders as expected -> open the documentation CI link, then log in and go to the relevant section of the documentation. This would need a PR in `odoo/documentation` to add the automodule for `util.report` and will require revision of all other functions we document here. Target 17.0
* Document the parameters in the same order they are used in the function.
* Use lowercase after the colon (basically follow the format as in other cases here ;))

Some of the above as a suggestion:
```suggestion
    :param list(tuple) data: data to report, each entry would be a row in the report. 
                             It could be empty, in which case only the summary is redered.                                                          
    :param tuple(str) columns: names for each column in "data", can be referenced in `row_format`.
    :param dict links: optional model/record links spec, the keys can be referenced in `row_format`.
```

src/util/report.py
```diff
@@ -126,6 +126,66 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+            {
+                link: get_anchor_link_to_record(rec_model, row_dict[id_col], row_dict[name_col])
+                for link, (rec_model, id_col, name_col) in links.items()
+            }
+        )
+        return "<li>{}</li>".format(row_format.format(**row_dict))
+
+    if not data:
+        row_to_html(columns)  # Validate the format is correct, including links
+        return report_with_summary(summary=summary, details="No records to display.", category=category)
```
aj-fuentes: Maybe this is better:
```suggestion
        return report_with_summary(summary=summary, details="", category=category)
```
then adapt `report_with_summary` to not add any summary/details structure when `details` is empty, and just output summary directly.

aj-fuentes (commented):
Looks better. Please adapt the documentation to the same format we use elsewhere in this repo. Some suggestions above.

src/util/report.py
```diff
@@ -126,6 +126,79 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+
+        .. code-block:: python
+
+            util.report_with_list(summary="The following records were altered.",
+                                  data=cr.fetchall(),
+                                  columns=("id", "name", "city", "comment", "company_id", "company_name")
+                                  row_format="Partner with id {partner_link} works at company {company_link} in {city}, ({comment})",
+                                  links={"company_link": ("res.company", "company_id", "company_name"), "partner_link": ("res.partner", "id", "name")},
+                                  category="Accounting"
+                                  )
```
aj-fuentes: The usual formatting allows for shorter lines. Unless there is some rendering issue?
```suggestion
            total = cr.rowcount
            data = cr.fetchmany(20)
            util.report_with_list(
                summary="The following records were altered.",
                data=data,
                columns=("id", "name", "city", "comment", "company_id", "company_name")
                row_format="Partner with id {partner_link} works at company {company_link} in {city}, ({comment})",
                links={
                    "company_link": ("res.company", "company_id", "company_name"),
                    "partner_link": ("res.partner", "id", "name")
                },
                total=total,
                category="Accounting"
            )
```
EDIT: we can include as well in the example `total=...`

src/util/report.py
```diff
@@ -126,6 +126,79 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+        else "<summary>{}</summary>".format(summary)
+    )
+    report(message=msg, category=category, format="html")
+
+
+def report_with_list(summary, data, columns, row_format, links=None, total=None, limit=100, category="Other"):
+    """Append the upgrade report with a new entry that displays a list of records.
+
+    The entry consists of a category (title) and a summary (body).
+    The entry displays a list of records previously returned by an SQL query, or any list as long as it's passed as a Python List of Tuples.
```
aj-fuentes: Try to keep the lines at max 90 chars (similar as we do for other docstrings here)
```suggestion
    The entry displays a list of records previously returned by an SQL query, or any list
    as long as it's passed as a Python List of Tuples.
```

src/util/report.py
```diff
@@ -126,6 +126,79 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+    :param list(tuple) data: data to report, each entry would be a row in the report.
+                             It could be empty, in which case only the summary is rendered.
+    :param tuple(str) columns: names for each column in "data", can be referenced in "row_format".
+    :param str row_format: the way a row in a list is formatted, using named placeholders, e.g.:
+                           "Partner {partner_link} that lives in {city} works at company {company_link}."
+    :param dict(str, tuple(str, str, str)) links: optional model/record links spec,
+                                                  the keys can be referenced in "row_format".
+    :param int total: if the original list was limited prior to calling this method, the original, total number of
+                      records, can be provided with this parameter.
+    :param int limit: the maximum number of records that are going to be displayed in the report.
```
aj-fuentes: Some ideas to shorten this and make to the point.
Also use back-quote instead of double quotes to refer to code.
```suggestion
    :param list(tuple) data: data to report, each entry would be a row in the report.
                             It could be empty, in which case only the summary is rendered.
    :param tuple(str) columns: columns in `data`, can be referenced in `row_format`.
    :param str row_format: format for rows, can use any name in `columns` or `links`, e.g.: 
                           `"Partner {partner_link} that lives in {city} works at
                           company {company_link}."`
    :param dict(str, tuple(str, str, str)) links: optional model/record links spec,
                                                  to use in `row_format`.
    :param int total: optional total number of records.
                      Taken as `len(data)` when `None` is passed.
                      Useful when `data` was limited by the caller.
    :param int limit: maximum number of records to list in the report.
                      If `data` contains more records the `total` number would be
                      included in the report as well.
```

src/util/report.py
```diff
@@ -126,6 +126,79 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+                for link, (rec_model, id_col, name_col) in links.items()
+            }
+        )
+        return "<li>{}</li>".format(row_format.format(**row_dict))
+
+    if not data:
+        row_to_html(columns)  # Validate the format is correct, including links
+        return report_with_summary(summary=summary, details="", category=category)
+
+    total = total or len(data)
```
aj-fuentes: To avoid weird situation where the number of entries in `data` is lower than the `limit`, but we still got `total`, we should adapt the `limit`
Do not use the old-style `or` shortcut, we can be explicit here: `None` is a special value.
```suggestion
    limit = min(limit, len(data))
    total = len(data) if total is None else total
```

src/util/report.py
```diff
@@ -126,6 +126,79 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+        row_dict.update(
+            {
+                link: get_anchor_link_to_record(rec_model, row_dict[id_col], row_dict[name_col])
+                for link, (rec_model, id_col, name_col) in links.items()
+            }
+        )
+        return "<li>{}</li>".format(row_format.format(**row_dict))
+
+    if not data:
+        row_to_html(columns)  # Validate the format is correct, including links
```
KangOl, line 202: It will fail. `column` doesn't have the expected type.

(a.k.a tests are missing)

src/util/report.py
```diff
@@ -126,6 +126,79 @@ def add_to_migration_reports(message, category="Other", format="text"):
         _logger.warning("Upgrade report is growing suspiciously long: %s characters so far.", migration_reports_length)
 
 
+report = add_to_migration_reports
```
KangOl: If we want to promote the usage of this new function, I would rename the function, so it appears with the good name in documentation, and alias the old name for compatibility.
```suggestion
add_to_migration_reports = report
```

src/base/tests/test_util.py
```diff
@@ -2677,3 +2677,63 @@ def test_assert_updated_combo(self):
...
+        [
+            (
+                "Simple test with minimal arguments and no data.",
+                [],
+                ("id", "name"),
+                "Partner {name} has id {id}",
+                None,
+                None,
+                100,
+                "Other",
```
diagnoza: I had to pass the default arguments as regular ones, because of how the tested method is called in `test_report_with_list`. Simply leaving them out would obviously result in
```
  File "/home/odoo/src/upgrade-util/src/testing.py", line 174, in wrapped
    return func(self, *args)
TypeError: TestReportUtils.test_report_with_list() missing 2 required positional arguments: 'category' and 'expected'
```

Passing them as `None` also wouldn't work, because since Python differentiates between "no argument passed" and `None`, they would actually be set to `None` instead of taking the default value. And if we pass `category=None`, then `add_to_migration_reports` will break on running `len(None)`.

src/base/tests/test_util.py
```diff
@@ -2682,3 +2682,63 @@ def test_assert_updated_combo(self):
         ):
             p2.city = "Niflheim"
             util.flush(p2)
+
+
+class TestReportUtils(UnitTestCase):
+    @parametrize(
```
KangOl: In this case, using `paramatrize` make the test less readable.
Better directly write 3 test functions.

src/base/tests/test_util.py
diagnoza: No longer an issue since we don't use `paramterize`.

src/util/report.py
```diff
@@ -126,6 +126,90 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+            )
+        return "<li>{}</li>".format(row_format.format(**row_dict))
+
+    if not data:
+        row_to_html(columns)  # Validate the format is correct, including links
+        return report_with_summary(summary=summary, details="", category=category)
+
+    total = len(data) if total is None else total
+    limit = min(limit, total) if limit != -1 else total
+    disclaimer = "The total number of affected records is {}.".format(total)
```
aj-fuentes: I think we shouldn't be explicit about the total unless it is explicitly passed. The reason is that maybe the caller just passed here a subset of the involved records in `data`, we cannot be sure that's the actual total number of affected records unless explicitly set.
```suggestion
    disclaimer = "The total number of affected records is {}.".format(total) if total else ""
    total = len(data) if total is None else total
    limit = min(limit, total) if limit != -1 else total    
```

src/util/report.py
```diff
@@ -127,6 +127,90 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+            )
+        return "<li>{}</li>".format(row_format.format(**row_dict))
+
+    if not data:
+        row_to_html(columns)  # Validate the format is correct, including links
+        return report_with_summary(summary=summary, details="", category=category)
+
+    disclaimer = "The total number of affected records is {}. ".format(total) if total else ""
+    total = len(data) if total is None else total
+    limit = min(limit, total) if limit != -1 else total
```
KangOl: Why `-1` and not `None` as sentinel value for no-limit?

diagnoza: IMO it may be confusing for a potential reader of the code. I think it's not common knowledge that passing `None` disables the pickup of a default value for a given param in Python.

diagnoza: As per DM with Alvaro:

1. We shouldn't assume devs don't know the language they're using.
2. `None` is actually more descriptive than `-1` for "no limit".

src/util/report.py
aj-fuentes: I think this is still unsolved. The problem here is that if we pass `row_links`, when `data` is empty the `zip(columns, columns)` won't have the right values for `row_dict[id_col]` that must be an int. What we can do instead is to prepare a dummy row:
```suggestion
        # Validate the format is correct, including links
        dummy_row = list(columns)
        if links:
            dummy_row[columns.index(id_col)] = 1
        row_to_html(dummy_row)
```

diagnoza:
@aj-fuentes @KangOl it's ready for another look.

diagnoza:
upgradeci retry

src/util/report.py
```diff
@@ -127,6 +127,90 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+
+    def row_to_html(row):
+        row_dict = dict(zip(columns, row))
+        if links:
+            row_dict.update(
+                {
+                    link: get_anchor_link_to_record(rec_model, row_dict[id_col], row_dict[name_col])
+                    for link, (rec_model, id_col, name_col) in links.items()
+                }
+            )
```
aj-fuentes, line 198: Once we generate the link items (which are already [escaped](https://github.com/odoo/upgrade-util/blob/0e5bd206f460681c7de3203597b77ff82b2181c0/src/util/report.py#L339)) I think we need to escape the remaining values in the dict.
For example imagine that `row_format` is `"{some_field}"` and in the dict we get `{"some_field": "<script>window.alert('escaped')</script>"}`
```suggestion
            )
        row_dict = {col: html_escape(val) for col, val in row_dict.items()}
```
cc: @KangOl

src/util/fields.py
```diff
@@ -1598,11 +1598,26 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
+    model_text = "All models"
+    if only_models:
+        model_text = "Models " + ", ".join("<kbd>{}</kbd>".format(m) for m in only_models)
+    total = cr.rowcount
+    if total:
+        summary = (
+            "{model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>. "
+            "The following server actions may need an update. "
+            "If a server action is a standard one and you haven't made any modifications, you may ignore it."
+        ).format(model_text=model_text, old=old, new=new)
```
aj-fuentes: We can use triple-quote strings here, in the end we are formatting HTML.
```suggestion
        summary = """
            {model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>.
            The following server actions may need an update.
            If a server action is a standard one and you haven't made any modifications, you may ignore it.
        """.format(model_text=model_text, old=old, new=new)
```

src/util/fields.py
```diff
@@ -1621,28 +1636,24 @@ def _update_field_usage_multi(cr, models, old, new, domain_adapter=None, skip_in
...
-                "<li>{}</li>".format(get_anchor_link_to_record("ir.model.fields", fid, fname))
-                for fid, fname in cr.fetchall()
+
+        total = cr.rowcount
+        if total:
+            summary = (
+                "{model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>. "
+                "The following compute methods of other fields may need an update. "
+                "If a field is a standard one and you haven't made any modifications, you may ignore it."
+            ).format(model_text=model_text, old=old, new=new)
```
```suggestion
            summary = """
                {model_text}: the field <kbd>{old}</kbd> has been renamed to <kbd>{new}</kbd>.
                The following compute methods of other fields may need an update.
                If a field is a standard one and you haven't made any modifications, you may ignore it.
            """.format(model_text=model_text, old=old, new=new)
```

src/util/report.py
```diff
@@ -127,6 +127,91 @@ def add_to_migration_reports(message, category="Other", format="text"):
...
+    :param int limit: maximum number of records to list in the report.
+                      If `data` contains more records than `limit`, the `total`
+                      number would be included in the report as well.
+                      Set `None` for no limit.
+    :param str category: title of a report entry.
+    """
+
+    def row_to_html(row):
+        raw = dict(zip(columns, row))
+        row_dict = {col: html_escape(val) for col, val in raw.items()}
```
aj-fuentes: `val` could be non-str.
```suggestion
        row_dict = {col: html_escape(str(val)) for col, val in raw.items()}
```

aj-fuentes (approved):

src/util/report.py
```diff
@@ -107,7 +107,7 @@ def html_escape(text):
 }
 
 
-def add_to_migration_reports(message, category="Other", format="text"):
+def report(message, category="Other", format="text"):
```
KangOl: That's not retro-compatible as it now shadow the `report` package.
The following snippet will not work anymore.
```python
from odoo.upgrade import util

util.report.html_escape("whatever")
```

```suggestion
def report_message(message, category="Other", format="text"):
```

KangOl:
Then, commit message should be adapted to the new function name.

src/util/report.py
```diff
@@ -107,7 +107,7 @@ def html_escape(text):
 }
 
 
-def add_to_migration_reports(message, category="Other", format="text"):
+def report_message(message, category="Other", format="text"):
```
KangOl, line 110: Other (stupid) proposals

```suggestion
def tell_customer(message, category="Other", format="text"):
```
```suggestion
def inform_user(message, category="Other", format="text"):
```
```suggestion
def yell(message, category="Other", format="text"):
```
```suggestion
def beware(message, category="Other", format="text"):
```
```suggestion
def upgrade_awareness(message, category="Other", format="text"):
```

KangOl (approved):
@robodoo rebase-ff r+

[^source]

[^source]: https://pixelfed.social/p/headofcatdept/980370589055479753

PR #388, opened 2026-02-20
[IMP] util.update_record_from_xml: set NULL/default values

Pirols:
Sometimes records get a field removed from their XML declaration. This creates a
divergence between new dbs and upgraded ones. The former will be initialised
with `NULL`/default values, whereas the latter will retain their current one.
This is usually handled in dedicated upgrade scripts with simple queries
un/setting the necessary columns. However, in some cases, these XML changes are
backported/noticed too late and need to be addressed in multiple versions.
That may create the pressure to target future versions, which creates a hidden
coupling between the xml declaration and the query, prone to turn into a bug.
It is one such [example] that inspired this PR.

Here `update_record_from_xml` is adapted to unset fields (or set them to their
default, if any) missing from the xml declaration, if passed explicitely via the
`fields` kwarg. This enables us to leverage the foreward compatibility of
`update_record_from_xml` in the scenario described above, preventing the
associated potential bugs.

[example]: https://github.com/odoo/upgrade/pull/9512

Needs https://github.com/odoo/upgrade/pull/10885

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1838,7 +1838,7 @@ class TestRecords(UnitTestCase):
         util.flush(usd)
         util.invalidate(usd)
 
-        util.update_record_from_xml(cr, xmlid, fields=("symbol"))
+        util.update_record_from_xml(cr, xmlid, fields=["symbol"])
         util.invalidate(usd)
 
         self.assertEqual(usd.symbol, "$")
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1844,6 +1844,33 @@ class TestRecords(UnitTestCase):
         self.assertEqual(usd.symbol, "$")
         self.assertEqual(usd.name, "XXX")
 
+    def test_update_record_from_xml__fields_not_in_xml(self):
+        cr = self.env.cr
+        eur_xmlid = "base.EUR"
+
+        eur = self.env.ref(eur_xmlid)
+        eur.write({"position": "before", "rounding": 0.00001})
+        util.flush(eur)
+        util.invalidate(eur)
+
+        util.update_record_from_xml(cr, eur_xmlid, fields=["position", "rounding"])
+        util.invalidate(eur)
+
+        self.assertEqual(eur.position, "after")  # default = "after", not in <record>
+        self.assertEqual(eur.rounding, 0.01)  # default = 0.01, same value in <record>
+
+        pubuser_xmlid = "base.public_user"
+
+        pubuser = self.env.ref(pubuser_xmlid)
+        pubuser.write({"active": True})
+        util.flush(pubuser)
+        util.invalidate(pubuser)
+
+        util.update_record_from_xml(cr, pubuser_xmlid, fields=["active"])
+        util.invalidate(pubuser)
+
+        self.assertFalse(pubuser.active)  # default = True, False in <record>
+
     def test_update_record_from_xml_cache(self):
         cr = self.env.cr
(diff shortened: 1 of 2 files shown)


src/util/records.py
```diff
@@ -1147,9 +1148,13 @@ def add_ref(ref):
                 found = True
                 parent = node.getparent()
                 if node.tag == "record" and fields is not None:
+                    fields_to_find = set(fields)
```
aj-fuentes:
```suggestion
                    xml_fields = set()
```

src/util/records.py
```diff
@@ -1147,9 +1148,13 @@ def add_ref(ref):
...
                     for fn in node.xpath("./field[@name]"):
                         if fn.attrib["name"] not in fields:
                             node.remove(fn)
+                        else:
+                            fields_to_find.remove(fn.attrib["name"])
+                    node.extend(lxml.builder.E.field(name=field_name, eval="False") for field_name in fields_to_find)
```
aj-fuentes:
```suggestion
                        xml_fields.add(fn.attrib["name"])
                    # override requested fields not found in xml
                    node.extend(lxml.builder.E.field(name=field_name, eval="False") for field_name in fields if field_name not in xml_fields)
```

aj-fuentes (commented):
I think current version is correct. My suggestion is because I find a bit simpler to read "override each field not found in the xml" than keeping a list of fields to find. 

That being said what I think could be problematic here are field with defaults. We would always write NULL but maybe that's fine for the use cases we have in mind. Another option may be to pass a dict of default values per field... but that may be too much for current cases... @KangOl ?

Pirols:
> @Pirols what's the status of this PR?

@diagnoza sorry for the late reply, this has been in my queue for a long time. I have just updated it, hopefully we will merge it soon. :)

Pirols:
Hi @KangOl, I have just pushed a version that checks whether the fields are found in *any* record node.

Compared to your suggestion I have not added the code to avoid inserting an empty `<record>` node. In fact, this has always been the case and with the current patch we will never[^1] process an empty node. Instead we can fill it (i.e. the last record node, which may be empty) with the missing (but required) `fields` values (NULL or default).

I believe that skipping the addition of an empty node would make this needlessly more complex.

[^1]: well except for xmlids defined in multiple record nodes, but those are so few that I doubt it would make significant difference.

KangOl:
upgradeci retry with always only hr_payroll

KangOl:
Can you add a test?

Pirols:
> Can you add a test?

Yes, I was looking into that.

src/base/tests/test_util.py
```diff
@@ -1838,7 +1838,7 @@ def test_update_record_from_xml__fields(self):
         util.flush(usd)
         util.invalidate(usd)
 
-        util.update_record_from_xml(cr, xmlid, fields=("symbol"))
+        util.update_record_from_xml(cr, xmlid, fields=["symbol"])
```
KangOl, line 1841: How does this test passed?

KangOl: Answer:
```python
   fn.attrib["name"] in "symbol"
```

KangOl:
This is great as it allows catching invalid field names.

KangOl:
upgradeci retry with always only l10n_be_hr_payroll hr_work_entry documents_hr_recruitment

KangOl (approved):
@robodoo rebase-ff r+

[^source]

[^source]: https://mastodon.social/@kittyhelpdesk/116998154888593670

PR #395, opened 2026-03-16
[ADD] @util.profiler.profile decorator

KangOl:


nseinlet (approved):

sagu-odoo:
~~Hello @KangOl, Is there any possibility to adapt this for the test mock crawl as well? Sometimes the upgrade request fails during the mock crawl, but when we run the actual run with upgrade start, it does not fail. This makes it difficult to track and analyze memory usage on the test server.~~

~~Is there any alternative approach available that would help us investigate and identify such issues more effectively?~~ as it now available

PR #413, opened 2026-04-10
[FIX] util.change_field_selection_values

KangOl:
Correctly handle `None` as source. It should match `NULL` in SQL. As [pointed by Chong], it will be coerced to the `"null"` string when converted to json and we can't use `=` to compare to NULL in SQL.

[pointed by Chong]: https://github.com/odoo/upgrade-util/pull/400#pullrequestreview-4074684375


KangOl:
upgradeci retry with always only payment account_peppol in 18.0

aj-fuentes (approved):
@robodoo r+

PR #414, opened 2026-04-10
[FIX] util.version_gte: make sure saas versions are handled in 7.0

mohammadhossam:
The `upgrade start` command on the test upgrade platform was failing for
databases in version 7.0. The root cause is that upgrade scripts using
`version_gte("saas~X.Y")` rely on `parse_version` to correctly strip
the "saas~" token when comparing version strings. However, in Odoo 7.0,
`parse_version` does not handle that token, so it leaks into the
returned tuple and breaks the comparison, crashing the upgrade process.

Strip `saas~` from the input of `version_gte` and `version_between`
before handing it to `parse_version`, so that saas version strings are
compared consistently regardless of the running Odoo version.

See: odoo/odoo@a6f724c9196efd323675a7d4ea5a9a8a26157b7b

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -24,6 +24,14 @@ except ImportError:
     from openerp.modules.module import get_module_path
     from openerp.tools.parse_version import parse_version
 
+if release.major_version == "7.0":
+    # This version doesn't handle `saas~` part in versions
+    _parse_version = parse_version
+
+    def parse_version(version):
+        return _parse_version(version.replace("saas~", ""))
+
+
 from .exceptions import MigrationError, SleepyDeveloperError
 
 # python3 shim


KangOl:
Those functions have never been written to be used in version 7.0, only from `7.saas~1`.

mohammadhossam:
> Those functions have never been written to be used in version 7.0, only from `7.saas~1`.

Yeah, I see. But like that, it's not possible to test a database in v7.0 on the test server as I noted in the message. Here is an excerpt from the server logs when I executed `upgrade start` on a v7.0 database:

```
2026-04-10 15:09:20,019 14 ERROR odoo_7 openerp: Failed to initialize database `odoo_7`.
  File "/home/odoo/src/odoo/7.0/openerp/cli/server.py", line 97, in preload_registry
    db, registry = openerp.pooler.get_db_and_pool(dbname,update_module=update_module)
  File "/home/odoo/src/odoo/7.0/openerp/pooler.py", line 33, in get_db_and_pool
    registry = RegistryManager.get(db_name, force_demo, status, update_module)
  File "/home/odoo/src/odoo/7.0/openerp/modules/registry.py", line 203, in get
    update_module)
  File "/home/odoo/src/odoo/7.0/openerp/modules/registry.py", line 233, in new
    openerp.modules.load_modules(registry.db, force_demo, status, update_module)
  File "/home/odoo/src/odoo/7.0/openerp/modules/loading.py", line 355, in load_modules
    loaded_modules, update_module)
  File "/home/odoo/src/odoo/7.0/openerp/modules/loading.py", line 256, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks)
  File "/home/odoo/src/odoo/7.0/openerp/modules/loading.py", line 159, in load_module_graph
    load_openerp_module(package.name)
  File "/home/odoo/src/odoo/7.0/openerp/modules/module.py", line 405, in load_openerp_module
    __import__('openerp.addons.' + module_name)
  File "/home/odoo/src/odoo/7.0/openerp/modules/module.py", line 133, in load_module
    mod = imp.load_module('openerp.addons.' + module_part, f, path, descr)
  File "/home/odoo/bin/addons/runsanitizer/__init__.py", line 14, in <module>
    from openerp.addons.base.maintenance.migrations import util
  File "/home/odoo/src/odoo/7.0/openerp/addons/base/maintenance/migrations/util/__init__.py", line 3, in <module>
    from .data import *
  File "/home/odoo/src/odoo/7.0/openerp/addons/base/maintenance/migrations/util/data.py", line 4, in <module>
    from .helpers import table_of_model
  File "/home/odoo/src/odoo/7.0/openerp/addons/base/maintenance/migrations/util/helpers.py", line 10, in <module>
    from .misc import splitlines, version_between, version_gte
  File "/home/odoo/src/odoo/7.0/openerp/addons/base/maintenance/migrations/util/misc.py", line 572, in <module>
    import odoo.tools.safe_eval as _safe_eval_mod
ImportError: No module named odoo.tools.safe_eval
```

aj-fuentes:
This should be patched in the platform.

aj-fuentes:
OK, from DM the problem is this line:
https://github.com/odoo/upgrade-util/blob/3af82a0295efaeb7416cd48d8f44ec2fdbeb6c91/src/util/misc.py#L585

I think we should switch to `18.4` with a comment.

aj-fuentes:
Another option is to make `parse_version` resilient, it could check if we are running before 8.0 and strip the `saas~` prefix. A quick check via `release.serie` should be enough. This is the second time we are hit with this issue.

mohammadhossam:
> Another option is to make `parse_version` resilient, it could check if we are running before 8.0 and strip the `saas~` prefix. A quick check via `release.serie` should be enough. This is the second time we are hit with this issue.

But that function is in odoo itself, so fixing it would require fixing it on `7.0` branch which is not possible as I understand, or do you mean fixing `version_gte` instead?

aj-fuentes:
> > Another option is to make `parse_version` resilient, it could check if we are running before 8.0 and strip the `saas~` prefix. A quick check via `release.serie` should be enough. This is the second time we are hit with this issue.
> 
> But that function is in odoo itself, so fixing it would require fixing it on `7.0` branch which is not possible as I understand, or do you mean fixing `version_gte` instead?

My bad, I meant `version_gte` not `parse_version`
https://github.com/odoo/upgrade-util/blob/0a39da9240642041bfe15604b0f3247710832aa4/src/util/misc.py#L91

KangOl:
Yes, a simple `.replace("saas~", "")` could do the trick.

KangOl:
> OK, from DM the problem is this line:
> 
> https://github.com/odoo/upgrade-util/blob/3af82a0295efaeb7416cd48d8f44ec2fdbeb6c91/src/util/misc.py#L585
> 
> I think we should switch to `18.4` with a comment.

If this is the only problematic call, then this is the less intrusive patch.

src/util/misc.py
```diff
@@ -101,7 +101,9 @@ def version_gte(version):
     """
     if "-" in version:
         raise SleepyDeveloperError("version cannot contains dash")
-    return parse_version(release.serie) >= parse_version(version)
+    # Strip `saas~` to keep comparisons working on Odoo 7.0, whose `parse_version`
+    # does not handle the token.
```
aj-fuentes:
```suggestion
    # Strip `saas~` to keep comparisons working on Odoo 7.0
```

KangOl:
> > OK, from DM the problem is this line:
> > https://github.com/odoo/upgrade-util/blob/3af82a0295efaeb7416cd48d8f44ec2fdbeb6c91/src/util/misc.py#L585
> > 
> > I think we should switch to `18.4` with a comment.
> 
> If this is the only problematic call, then this is the less intrusive patch.

Self-reply. No, there are other cases:
```
;  rg '^if version_(gte|between)'
src/util/inherit.py
12:if version_gte("saas~17.1"):

src/util/modules.py
63:if version_gte("15.0"):
93:if version_gte("16.0"):

src/util/records.py
333:if version_gte("saas~14.3"):

src/util/fields.py
1061:if version_gte("saas~17.5"):
1141:if version_gte("16.0"):

src/util/misc.py
570:if version_gte("saas~18.4"):

src/util/pg.py
534:if version_gte("10.0"):  # Since at least 9.0
564:if version_gte("14.0"):  # Since at least 13.0
568:if version_gte("saas~16.1"):  # Since 16.0
571:if version_gte("saas~17.1"):  # Since 17.0
575:if version_gte("saas~18.1"):  # Since 18.0
```

KangOl:
upgradeci retry with always only mail in all versions

src/util/misc.py
```diff
@@ -120,7 +121,9 @@ def version_between(a, b):
     """
     if "-" in a + b:
         raise SleepyDeveloperError("version cannot contains dash")
-    return parse_version(a) <= parse_version(release.serie) <= parse_version(b)
+    return (
+        parse_version(a.replace("saas~", "")) <= parse_version(release.serie) <= parse_version(b.replace("saas~", ""))
+    )
```
aj-fuentes: Fewer lines, shorter diff, same result :)
```suggestion
    a, b = a.replace("saas~", ""), b.replace("saas~", "")
    return parse_version(a) <= parse_version(release.serie) <= parse_version(b)
```

mohammadhossam: Done :+1:

KangOl:
What about an ad-hoc replacement only when running in `7.0`?
```diff
diff --git src/util/misc.py src/util/misc.py
index 7393b90..78ada65 100644
--- src/util/misc.py
+++ src/util/misc.py
@@ -24,6 +24,14 @@ except ImportError:
     from openerp.modules.module import get_module_path
     from openerp.tools.parse_version import parse_version
 
+    if release.major_version == "7.0":
+        # This version doesn't handle `saas~` part in versions
+        # wrap `parse_version` to remove it
+        _parse_version = parse_version
+
+        def parse_version(version):
+            return _parse_version(version.replace("saas~", ""))
+
 from .exceptions import MigrationError, SleepyDeveloperError
 
 # python3 shim
```

aj-fuentes:
> What about an ad-hoc replacement only when running in `7.0`?

Makes sense.

mohammadhossam:
> What about an ad-hoc replacement only when running in `7.0`?
> 
> ```diff
> diff --git src/util/misc.py src/util/misc.py
> index 7393b90..78ada65 100644
> --- src/util/misc.py
> +++ src/util/misc.py
> @@ -24,6 +24,14 @@ except ImportError:
>      from openerp.modules.module import get_module_path
>      from openerp.tools.parse_version import parse_version
>  
> +    if release.major_version == "7.0":
> +        # This version doesn't handle `saas~` part in versions
> +        # wrap `parse_version` to remove it
> +        _parse_version = parse_version
> +
> +        def parse_version(version):
> +            return _parse_version(version.replace("saas~", ""))
> +
>  from .exceptions import MigrationError, SleepyDeveloperError
>  
>  # python3 shim
> ```

@KangOl Done :+1:

src/util/misc.py
```diff
@@ -24,6 +24,15 @@
     from openerp.modules.module import get_module_path
     from openerp.tools.parse_version import parse_version
 
+if release.major_version == "7.0":
+    # This version doesn't handle `saas~` part in versions
+    # wrap `parse_version` to remove its
```
aj-fuentes: nit, the code speaks for itself, it's clear that we just remove it.
```suggestion
```

aj-fuentes (approved):
@robodoo r+

PR #417, opened 2026-04-21
[IMP] util.indirect_references()

KangOl:
Yield `many2one_reference` fields, now that we have the name of the "res_model" column in the database.

See odoo/odoo@753d693fcec0074a78eb5cef1b1fbb06ea346793


KangOl:

PR #418, opened 2026-04-22
[FIX] util/orm: don't mark inherited field manual for custom

sagu-odoo:
During the upgrade process, inherited fields were incorrectly marked as `manual`, whereas they should retain their original `base` state.

like Fields created through Studio or `ir.model.fields` are expected to have `state = manual`, while inherited fields are set up by the ORM with `state = base` [1].

Marking inherited fields as `manual` makes the ORM treat them as independently created fields, which can trigger unnecessary computations, e.g  for non-stored fields and
`Many2many` fields related to `ir.attachment`.

This change ensures that:
* inherited fields retain `state = base`
* only fields created from `ir.model.fields` or coming   from custom module will have the state  `manual` not the inherited one.
* field setup remains consistent with the ORM behavior

This prevents inherited fields from being treated as independent fields and avoids unwanted computations during mock crawl without marking as inhertied field.

 
[1]: https://github.com/odoo/odoo/blob/d58f4ed332af35f6de26a93f07adf05368731e20/odoo/orm/model_classes.py#L493-L508

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -16,6 +16,8 @@ from functools import wraps
 from itertools import chain
 from textwrap import dedent
 
+from psycopg2 import sql
+
 try:
     from unittest.mock import patch
 except ImportError:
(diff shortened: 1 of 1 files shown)


KangOl:
The `MemoryError` traceback looks unrelated and more like a side-effect.

aj-fuentes:
Besides what @KangOl mentioned I want to point out that the description in the PR is quite ambiguous. Please clarify what you meant. The query you modified mark all fields that weren't loaded as manual, this is important to avoid other errors. It doesn't make sense to disable this without a clear description of the setup such that we know if it is really safe to do so. Regarding the linked patch. If you need something to "get applied" maybe there are other ways to achieve that. Hence, once more, a detailed setup is important.

sagu-odoo:
> The MemoryError traceback looks unrelated and more like a side-effect.


The issue  related to computation being triggered on inherited fields, which caused blocking especially for non-stored fields  Many2Many fields linked with `ir.attachment`.

The intention of the fix is to ensure that inherited fields are **not marked as manual fields**. Only fields created via the UI or explicitly through `ir.model.fields` should be considered manual.

There is no issue with stored field computation; the problem specifically occurs with inherited (non-stored) fields. 

This has been addressed in the following PR, which resolves the blocking of upgrade requests for almost:

* [https://github.com/odoo/odoo/pull/239400](https://github.com/odoo/odoo/pull/239400/changes)

Related fixes to address the traceback but not for 19.0 version

* [https://github.com/odoo/odoo/pull/259572/changes](https://github.com/odoo/odoo/pull/259572/changes)
* [https://github.com/odoo/odoo/pull/246333/changes](https://github.com/odoo/odoo/pull/246333/changes)

sagu-odoo:
this isssue mostly limited to custom module fields only as of now.

KangOl:

sagu-odoo:
for mentioned traceback, I think we can backport something like  this fix not exact to 19.0 https://github.com/odoo/odoo/pull/246333 . So, group by feature on ``ir.attachment`` will be disable because of [this](https://github.com/odoo/odoo/pull/246333/changes#diff-ab3dadc37163820f8e863edb1dd216f8b0e7ccf19cf4e2052577a7bae7ddd7e8R616-R617) and [this](https://github.com/odoo/odoo/blob/969f7ba1827046d15be7cc8f883fa9678a2d0dcd/odoo/orm/fields.py#L921-L925).

sagu-odoo:
Hello @aj-fuentes and @KangOl , can i get your opinion on this ?.

Thanks in advance

sagu-odoo:
Hello @aj-fuentes and @KangOl ,Gentle reminder! can i get your opinion on this ?.

Thanks in advance

sagu-odoo:
Hllo @aj-fuentes , Gentle reminder!, can i get review on this ?.

Thanks in advance.

src/util/orm.py
```diff
@@ -663,34 +665,68 @@ def custom_module_field_as_manual(env, rollback=True, do_flush=False):
...
+
+    query = format_query(
+        env.cr,
+        """
+                {cte}
+                UPDATE ir_model_fields
+                   SET state = 'manual'
+                 WHERE state = 'base'
+                   AND model = %s
+                   AND name NOT IN %s
```
KangOl: Use keyword-arguments to avoid the `if cte`
```suggestion
                   AND model = %(model)s
                   AND name NOT IN %(fields)s
```

sagu-odoo: done.

sagu-odoo:
Hello @KangOl , could you please have another look ?.

Thanks in advance

PR #419, opened 2026-04-23
[IMP] util.get_inherit_model_names

HydrionBurst:
See odoo/odoo#260982
See odoo/enterprise#114817
See odoo/upgrade#10028

### DIFF
--- a/src/util/orm.py
+++ b/src/util/orm.py
@@ -229,6 +229,25 @@ def invalidate(records, *args):
     ((env and getattr(records.env, "invalidate_all", None)) or records.invalidate_cache)(*args)
 
 
+def get_inherit_model_names(model):
+    """
+    Get the names of the models the model directly or indirectly `_inherit`
+
+    :param model: recordset
+    :return: the names of the models the model directly or indirectly `_inherit`.
+        If ``model._name`` is not in the returned set, the model is defined once.
+        ``base`` is always in the returned set for non-``base`` models.
+    :rtype: list[str]
+    """
+    inherit_model_names = set()
+    for cls in type(model).mro():
+        if getattr(cls, "pool", None) is None and getattr(cls, "_inherit", None) is not None:
+            inherit_model_names.update(cls._inherit)
+    if model._name != "base":
+        inherit_model_names.add("base")
+    return sorted(inherit_model_names)
+
+
 def no_selection_cache_validation(f=None):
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only crm in all versions

src/util/models.py
```diff
@@ -683,3 +683,22 @@ def convert_model_to_abstract(cr, model, drop_table=True, keep=()):
     for field in ["id", "__last_update", "display_name", "create_uid", "create_date", "write_uid", "write_date"]:
         if field not in keep:
             remove_field(cr, model, field)
+
+
+def get_inherit_model_names(model):
```
KangOl: Can you move this function in the `orm.py` file, below the `invalidate` function?

src/util/orm.py
```diff
@@ -229,6 +229,25 @@ def invalidate(records, *args):
...
+    """
+    Get the names of the models the model directly or indirectly `_inherit`
+
+    :param model: recordset
+    :return: the names of the models the model directly or indirectly `_inherit`.
+        If ``model._name`` is not in the returned set, the model is defined once.
+        ``base`` is always in the returned set for non-``base`` models.
+    :rtype: set(str)
+    """
+    inherit_model_names = set()
```
aj-fuentes, line 242: I think this function must return the names in a consistent order. Callers could iterate on the list. With a set this could lead to non-deterministic orders across different runs.

A simple approach would be to use `collections.OrderedDict` (for Python2 compatibility).

I have no preference with respect of the order but we should be explicit as well in the docstring. For example: "the list of model names is returned as per the MRO of the class without repetitions".

HydrionBurst: Not really needed for now.
The only place to iterate it is `any("mixin" in m and f.name in self.env[m]._fields for m in util.get_inherit_model_names(model))` which doesn't rely on the order.

I don't want to make it ordered to aovid developers expect the order will be the same as the mro order. And we can still `sorted(get_inherit_model_names(model))` when necessary

aj-fuentes: As I said, I don't care much about the order. But I do care about determinism because this could be a big headache when debugging an upgrade. So at least return `sorted(inherit_model_names)`

aj-fuentes (approved):
LGTM

KangOl (approved):
@robodoo r+

PR #420, opened 2026-04-23
[IMP] util.{parallel,explode}_execute

KangOl:
Add new `qualifier` argument which is forwarded to the `log_progress` call.


aj-fuentes (approved):

KangOl:

KangOl:
@robodoo r+

PR #422, opened 2026-04-23
[IMP] util/fields: add `dump_field_to_chatter`

aj-fuentes:
Allow to save column info to the chatter.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -48,7 +48,7 @@ from . import json
 from .const import ENVIRON
 from .domains import _adapt_one_domain, _replace_path, _valid_path_to, adapt_domains
 from .exceptions import SleepyDeveloperError, UpgradeError
-from .helpers import _dashboard_actions, _validate_model, resolve_model_fields_path, table_of_model
+from .helpers import _dashboard_actions, _validate_model, model_of_table, resolve_model_fields_path, table_of_model
 from .inherit import for_each_inherit
 from .misc import AUTO, log_progress, safe_eval, version_gte
 from .orm import env, invalidate
@@ -65,13 +65,14 @@ from .pg import (
     get_columns,
     get_value_or_en_translation,
     parallel_execute,
+    pg_html_escape,
     pg_replace,
     pg_text2html,
     remove_column,
     table_exists,
     target_of,
 )
-from .records import _remove_import_export_paths
+from .records import _remove_import_export_paths, ref
 from .report import add_to_migration_reports, get_anchor_link_to_record
 
 # python3 shims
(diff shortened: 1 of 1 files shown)


src/util/misc.py
```diff
@@ -841,3 +841,183 @@ def __repr__(self):
 
 :meta private: exclude from online docs
 """
+
+
+def dump_column_to_chatter(
+    cr,
```
KangOl: Informally, this file only contains functions that doesn't use a cursor.
I would move this function into `fields.py` (or `models.py`)

src/util/misc.py
```diff
@@ -841,3 +841,183 @@ def __repr__(self):
...
+    where=None,
+    html_escape=True,
+    message_type="notification",
+    subtype_xmlid=None,
+):
+    """
+    Post a chatter message on each record to preserve a column value before it is dropped.
+
+    This is useful when a field is being removed during a migration but clients may need to
+    recover its values. The value is posted as a ``mail_message`` on the record's chatter.
```
KangOl: Do we have cases where the column is not droppe?

src/util/misc.py
```diff
@@ -841,3 +841,183 @@ def __repr__(self):
...
+def dump_column_to_chatter(
+    cr,
+    table,
+    column,
+    res_id="id",
+    res_model=None,
+    label="%s",
+    fk_display=None,
+    where=None,
+    html_escape=True,
```
KangOl: It may be too dangerous to not do it.
```suggestion
```

src/util/misc.py
aj-fuentes: There are cases I think where the column is overridden [`phone != mobile`](https://github.com/odoo/upgrade/pull/10030/changes#diff-99f505bd1e3c400f58d05e1272e59791c794933e8f25b834ec86aa7056c31786R63). Or where the column is further used in other queries.

I actually _had_ in mind to write a `dump_column_to_chatter_query` util that returns a query ready to be used (maybe in a CTE) as part of a bigger query --and make current util use it internally. Because we have some cases already where the dump to chatter happens in a CTE on a bigger query. But this may be overkill, we perhaps can write those queries without the CTE.

src/util/misc.py
aj-fuentes: This is for the case where the column is already in html, I guess we could autodetect it via the matching field if we need to.

aj-fuentes: We already have such case see
https://github.com/odoo/upgrade/pull/10030/changes#diff-4243a885e08a944d69258e3784e6e3f3a27dcc42a7c32edb0b83ca4ac81a9262R80

src/util/misc.py
aj-fuentes: Moved to `fields.py`

KangOl:

src/util/misc.py
KangOl: Can't it be auto determined base on the type of the dumped field?

src/util/fields.py
```diff
@@ -1907,3 +1908,175 @@ def update_server_actions_fields(cr, src_model, dst_model=None, fields_mapping=N
         ) % {"src_model": src_model, "dst_model": dst_model, "actions": ", ".join(action_names)}
 
         add_to_migration_reports(message=msg, category="Server Actions")
+
+
+def dump_column_to_chatter(
+    cr,
+    table,
+    column,
```
KangOl: I would work on "field" instead of "column"
```suggestion
def dump_field_to_chatter(
    cr,
    model,
    field,
```

KangOl (approved):
@robodoo r+

PR #424, opened 2026-04-23
Delete/uninstall

daniel-pcg:


aj-fuentes:
Could you please fix whatever is making you create these PRs?

Module-specific scripts should never be put inside the upgrade-util repo. Just use any other path and pass it via `--upgrade-path=/path/to/util/repo/src,/path/to/your/extra/scripts`. This repo is only for the utility functions.

Thanks.

PR #425, opened 2026-04-27
[IMP] util/logger: add `getLogger`

aj-fuentes:
Since Odoo 16 we now have a nice `__name__` for loaded upgrade scripts. For older versions we can still generate one. In this patch we provide a new function to get a "canonical" logger for an upgrade script.

See: https://github.com/odoo/upgrade/pull/10057

### DIFF
--- a/src/util/logger.py
+++ b/src/util/logger.py
@@ -2,12 +2,40 @@
 import atexit
 import logging
 import os
+import sys
 
 from .misc import on_CI
 
 _REGISTERED = False
 
 _logger = logging.getLogger(__name__.rpartition(".")[0])
+
+
+def getLogger(name=None):
+    """
+    Return a logger named after the caller.
+
+    If `name` is a dotter module, it is used as-is. Otherwise the logger name is taken
+    from the caller's `__name__` when set to a dotted module name (Odoo 16+), or built
+    from the last three components of the caller's file path as
+    `odoo.upgrade.<module>.<version>.<script>`. `<script>` would be `name` parameter if
+    set and non-dotted.
+    """
+    if name and "." in name:
+        return logging.getLogger(name)
+    frame = sys._getframe(1)
+    script = ""
+    if name is None:
+        name = frame.f_globals.get("__name__") or ""
+    else:
+        script = name
+    if "." not in name:
+        path = frame.f_globals.get("__file__") or ""
+        parts = os.path.normpath(path).split(os.sep)
+        module, version, script = parts[-3], parts[-2], script or os.path.splitext(parts[-1])[0]
+        prefix = __name__.rsplit(".", 2)[0]
+        name = ".".join([prefix, module, version, script])
+    return logging.getLogger(name)
 
 
 class CriticalHandler(logging.Handler):
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -249,12 +249,17 @@ try:
                          .. note::
                             The script must be available in the upgrade path.
 
-        :param str or None name: name to assign to the returned module, take the name from
-                                 the imported file if `None`
+        :param str or None name: name to assign to the returned module. When `None`,
+                                 a name is derived from `path` to mirror the name set
+                                 by Odoo (>=16) when loading migration scripts:
+                                 `odoo.upgrade.<module>.<version>.<script>`.
         :return: a module created from the imported upgrade script
         """
         if not name:
-            name, _ = os.path.splitext(os.path.basename(path))
+            parts = os.path.normpath(path).split(os.sep)
+            module, version, script = parts[-3], parts[-2], os.path.splitext(parts[-1])[0]
+            prefix = __name__.rsplit(".", 2)[0]
+            name = ".".join([prefix, module, version, script])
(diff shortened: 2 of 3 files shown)


KangOl:

src/util/logger.py
```diff
@@ -10,6 +11,26 @@
...
+    `odoo.upgrade.<module>.<version>.<script>`.
+    """
+    if name is None:
+        frame = sys._getframe(1)
+        name = frame.f_globals.get("__name__") or ""
+        if "." not in name:
+            path = frame.f_globals.get("__file__") or ""
+            parts = os.path.normpath(path).split(os.sep)
+            module, version, script = parts[-3], parts[-2], os.path.splitext(parts[-1])[0]
+            name = "odoo.upgrade.{}.{}.{}".format(module, version, script)
```
KangOl: For older version, shouldn't we use `openerp.upgrade`?

KangOl: We can use `__name__`
```suggestion
            prefix = __name__.rsplit(".", 2)[0]
            name = ".".join([prefix, module, version, script])
```

aj-fuentes: > For older version, shouldn't we use `openerp.upgrade`?

We can, but, is there any advantage? Even ppl still in openerp already know what odoo means :)

aj-fuentes: > We can use `__name__`
> ```suggestion
>             name = "odoo.upgrade.{}.{}.{}".format(module, version, script)
>             prefix = __name__.rsplit(".", 2)[0]
>             name = ".".join([prefix, module, version, script])
> ```

I don't get it. The idea was to align with https://github.com/odoo/odoo/blob/4301c99fda4c2fa37a45261b1190f9984c5603f1/odoo/modules/migration.py#L180 i.e. `odoo.upgrade.{module}.{version}.{name}`

KangOl: > > For older version, shouldn't we use `openerp.upgrade`?
> 
> We can, but, is there any advantage? Even ppl still in openerp already know what odoo means :)

Yes, but the logging configuration is done on the `openerp` logger, not `odoo`.

KangOl: > > We can use `__name__`
> 
> I don't get it. The idea was to align with https://github.com/odoo/odoo/blob/4301c99fda4c2fa37a45261b1190f9984c5603f1/odoo/modules/migration.py#L180 i.e. `odoo.upgrade.{module}.{version}.{name}`

Yes. It extract the `odoo.upgrade` part from the current package. On older versions, `prefix` will be `odoo.addons.base.maintenance.migrations` (depending on where `util` is imported from)

aj-fuentes: 🤦🏼 I misread `__name__` as `name` and was utterly confused, now I got it :)

src/util/misc.py
```diff
@@ -249,12 +249,17 @@ def migrate(cr, version):
...
-                                 the imported file if `None`
+        :param str or None name: name to assign to the returned module. When `None`,
+                                 a name is derived from `path` to mirror the name set
+                                 by Odoo (>=16) when loading migration scripts:
+                                 `odoo.upgrade.<module>.<version>.<script>`.
         :return: a module created from the imported upgrade script
         """
         if not name:
-            name, _ = os.path.splitext(os.path.basename(path))
+            parts = os.path.normpath(path).split(os.sep)
```
KangOl, line 259: Needs to be done for the python2 version (even if not used alot[^1])

[^1]: *The Alot is an imaginary creature that I made up to help me deal with my compulsive need to correct other people's grammar.*

aj-fuentes: Done, with a slight detour... but it avoids the RuntimeWarning.

KangOl:
upgradeci retry with always only sale in all versions

src/util/logger.py
```diff
@@ -10,6 +11,27 @@
 _logger = logging.getLogger(__name__.rpartition(".")[0])
 
 
+def get_logger(name=None):
```
KangOl: I know it goes again our naming, but maybe we can use camelCase to mimic the function in `logging`
```suggestion
def getLogger(name=None):
```

src/util/logger.py
```diff
@@ -10,6 +11,27 @@
...
+    """
+    Return a logger named after the caller.
+
+    If `name` is given, it is used as-is. Otherwise the logger name is taken
+    from the caller's `__name__` when set to a dotted module name (Odoo 16+),
+    or built from the last three components of the caller's file path as
+    `odoo.upgrade.<module>.<version>.<script>`.
+    """
+    if name is None:
```
KangOl, line 28: Shouldn't we also prepend `prefix.module.version.` if the given name doesn't contain dots?

KangOl (changes requested):

src/util/logger.py
aj-fuentes: Renamed.

src/util/logger.py
aj-fuentes: Done.

KangOl (approved):
@robodoo rebase-ff r+

PR #428, opened 2026-05-04
[ADD] spreadsheet.multiprocess_commands

KangOl:
New function that will process commands using multiprocesses.

### DIFF
--- a/src/util/spreadsheet/misc.py
+++ b/src/util/spreadsheet/misc.py
@@ -1,69 +1,105 @@
-from .. import json, pg
+import collections
+import logging
+import multiprocessing
+import warnings
+from concurrent.futures import ProcessPoolExecutor
+from functools import partial
+
+from odoo import sql_db
+
+from .. import json
+from ..misc import log_progress, make_pickleable_callback
+from ..pg import SQLStr, format_query, get_max_workers
 
 MEMORY_CAP = 2 * 10**8  # 200MB
+COUNT_CAP = 1000
+_logger = logging.getLogger(__name__)
 
 
-def iter_commands(cr, like_all=(), like_any=()):
+def _search_ids(cr, like_all=(), like_any=()):
     if not (bool(like_all) ^ bool(like_any)):
         raise ValueError("Please specify `like_all` or `like_any`, not both")
 
-    with pg.named_cursor(cr, itersize=1) as ncr:
-        ncr.execute(
-            pg.format_query(
-                cr,
-                """
-                WITH filtered AS (
-                    SELECT id,
-                           commands,
-                           LENGTH(commands) AS commands_length
-                      FROM spreadsheet_revision
-                     WHERE commands LIKE {condition} (%s::text[])
-                ), smaller AS (
-                    SELECT id,
-                           commands,
-                           sum(commands_length) OVER (ORDER BY id) / %s AS num
-                      FROM filtered
-                     WHERE commands_length <= %s
-                )
-                SELECT array_agg(id ORDER BY id),
-                       array_agg(commands ORDER BY id),
-                       min(id) AS sort_key
+    cr.execute(
+        format_query(
+            cr,
+            """
+            WITH filtered AS (
+                SELECT id,
+                       LENGTH(commands) AS commands_length
+                  FROM spreadsheet_revision
+                 WHERE commands LIKE {condition} (%(like)s::text[])
+            ), smaller AS (
+                SELECT id,
+                       sum(commands_length) OVER (ORDER BY id) / %(mem_cap)s AS num
+                  FROM filtered
+                 WHERE commands_length <= %(mem_cap)s
+            ),
+            aggregated AS (
+                SELECT array_agg(id ORDER BY id) as arr
                   FROM smaller
               GROUP BY num
 
                  UNION ALL
 
-                SELECT ARRAY[id],
-                       ARRAY[commands],
-                       id AS sort_key
+                SELECT ARRAY[id] as arr
                   FROM filtered
-                 WHERE commands_length > %s
-              ORDER BY sort_key
-                """,
-                condition=pg.SQLStr("ALL" if like_all else "ANY"),
+                 WHERE commands_length > %(mem_cap)s
             ),
-            [list(like_any or like_all), MEMORY_CAP, MEMORY_CAP, MEMORY_CAP],
-        )
-        for ids, commands, _ in ncr:
-            for revision_id, data in zip(ids, commands):
-                data_loaded = json.loads(data)
-                if "commands" not in data_loaded:
-                    continue
-                data_old = json.dumps(data_loaded, sort_keys=True)
+            chunked AS (
+                SELECT arr[i : i + %(count_cap)s -1] as chunk
+                  FROM aggregated,
+                       LATERAL generate_series(1, cardinality(arr), %(count_cap)s) AS i
+            )
+            SELECT chunk
+              FROM chunked
+          ORDER BY chunk[1]
+            """,
+            condition=SQLStr("ALL" if like_all else "ANY"),
+        ),
+        {
+            "like": list(like_any or like_all),
+            "mem_cap": MEMORY_CAP,
+            "count_cap": COUNT_CAP,
+        },
+    )
+    return [ids for (ids,) in cr.fetchall()]
 
-                changed = yield data_loaded["commands"]
-                if changed is None:
-                    changed = data_old != json.dumps(data_loaded, sort_keys=True)
 
-                if changed:
-                    cr.execute(
-                        "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s",
-                        [json.dumps(data_loaded), revision_id],
-                    )
+def _iter_ids(cr, ids):
+    cr.execute("SELECT id, commands FROM spreadsheet_revision WHERE id=ANY(%s)", [list(ids)])
+    for revision_id, data in cr.fetchall():
+        data_loaded = json.loads(data)
+        if "commands" not in data_loaded:
+            continue
+        data_old = json.dumps(data_loaded, sort_keys=True)
+
+        changed = yield data_loaded["commands"]
+        if changed is None:
+            changed = data_old != json.dumps(data_loaded, sort_keys=True)
+
+        if changed:
+            cr.execute(
+                "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s",
+                [json.dumps(data_loaded), revision_id],
+            )
 
 
-def process_commands(cr, callback, *args, **kwargs):
-    gen = iter_commands(cr, *args, **kwargs)
+def iter_commands(cr, like_all=(), like_any=()):
+    warnings.warn(
+        "`iter_commands` is deprecated; use `multiprocess_commands` instead.",
+        category=DeprecationWarning,
+        stacklevel=2,
+    )
+    return _iter_commands(cr, like_all, like_any)
+
+
+def _iter_commands(cr, like_all=(), like_any=()):
+    for ids in _search_ids(cr, like_all, like_any):
+        yield from _iter_ids(cr, ids)
+
+
+def _transform(callback, gen):
     try:
         cmd = next(gen)
         while True:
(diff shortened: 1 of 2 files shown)


src/util/spreadsheet/misc.py
```diff
@@ -44,26 +51,41 @@ def iter_commands(cr, like_all=(), like_any=()):
...
 
-def process_commands(cr, callback, *args, **kwargs):
-    gen = iter_commands(cr, *args, **kwargs)
+        if changed:
+            queries.append(
+                cr.mogrify(
+                    "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s",
+                    [json.dumps(data_loaded), revision_id],
+                ).decode()
+            )
```
cawo-odoo: If I'm not mistaken, this means that when  `_iter_ids` is called from `_mp_callback`, this may result in 73 concurrent DB connections and potentially as many active queries. This seems excessive.

```suggestion
            qry = cr.mogrify(
                "UPDATE spreadsheet_revision SET commands=%s WHERE id=%s",
                [json.dumps(data_loaded), revision_id],
            ).decode()
            if in_forked_child:
                cr.execute(qry)
            else:
                queries.append(qry)
```

src/util/spreadsheet/misc.py
```diff
@@ -72,3 +94,29 @@ def process_commands(cr, callback, *args, **kwargs):
...
+
+    cr.commit()
+    with ProcessPoolExecutor(
+        max_workers=pg.get_max_workers(),
+        initializer=init_worker_process,
+        mp_context=multiprocessing.get_context("fork"),
+    ) as executor:
+        futures = [executor.submit(_mp_callback, cr.dbname, callback, ids) for ids in _search_ids(cr, *args, **kwargs)]
+        for future in concurrent.futures.as_completed(futures):
+            future.result()
```
cawo-odoo: In similar PRs this was suggested as the more intuitive pattern:
```suggestion
        collections.deque(
            executor.map(
                _mp_callback, itertools.repeat(cr.dbname), itertools.repeat(callback), _search_ids(cr, *args, **kwargs))
            ),
            maxlen=0,
        )
```

aj-fuentes: ... with `maxlen=0` ;)
https://docs.python.org/3/library/itertools.html#itertools-recipes

I prefer the deque consume + `map` idiom, even if at first sight is more "complex", and it works as we expect. Unrolling the loop means keeping in mind the internals of `concurrent.futures` which IMO doesn't add any clarity here. If we keep the `future.result()` we need at least a comment ` # to raise any possible exception`

src/util/spreadsheet/misc.py
aj-fuentes: I think this comment is right. We need a way to identify if we are running in MP mode. I think a simple parameter to `_iter_ids` could work here.

KangOl: Actually, using `parallel_execute` was part of my first attempt to accelerate the script.
I didn't see any improvement and then goes with multiprocess.
I can go back to the original code and execute the update at each loop.
OTOH, I didn't had any issues in my tests...

src/util/spreadsheet/misc.py
```diff
@@ -72,3 +90,40 @@ def process_commands(cr, callback, *args, **kwargs):
...
+
+
+def _mp_callback(dbname, callback, ids):
+    with sql_db.db_connect(dbname).cursor() as cr:
+        _map(callback, _iter_ids(cr, ids))
+
+
+def multiprocess_commands(cr, callback, like_all=(), like_any=()):
+    if not (bool(like_all) ^ bool(like_any)):
+        raise ValueError("Please specify `like_all` or `like_any`, not both")
```
aj-fuentes, line 129: We can avoid the duplicated validation:
```suggestion
    cr.commit()
    # _search_ids validates parameters
    gen = _search_ids(cr, like_all=like_all, like_any=like_any)
```

src/util/spreadsheet/misc.py
```diff
@@ -1,69 +1,87 @@
...
+    warnings.warn(
+        "`iter_commands` is deprecated; use `multiprocess_commands` instead.",
+        category=DeprecationWarning,
+        stacklevel=1,
+    )
+    for ids in _search_ids(cr, like_all, like_any):
+        yield from _iter_ids(cr, ids)
+
+
+def _map(callback, gen):
```
aj-fuentes: I wouldn't call this `_map`, since `list(_map(func, iter))` makes no sense.
```suggestion
def _transform(callback, gen):
```

src/util/spreadsheet/misc.py
```diff
@@ -72,3 +90,40 @@ def process_commands(cr, callback, *args, **kwargs):
...
+
+    callback = make_pickleable_callback(callback)
+
+    cr.commit()
+    with ProcessPoolExecutor(
+        max_workers=pg.get_max_workers(),
+        initializer=init_worker_process,
+        mp_context=multiprocessing.get_context("fork"),
+    ) as executor:
+        gen = _search_ids(cr, like_all=like_all, like_any=like_any)
```
aj-fuentes: Do we really need this inside the context manager? (related to the [other](https://github.com/odoo/upgrade-util/pull/428/changes#r3189445089) comment)
```suggestion
```

src/util/spreadsheet/misc.py
KangOl: It's was more to have the traceback at the right place.

src/util/spreadsheet/misc.py
```diff
@@ -72,3 +94,40 @@ def process_commands(cr, callback, *args, **kwargs):
...
+        # consume the submitted tasks
+        collections.deque(
+            executor.map(partial(_mp_callback, cr.dbname, callback), gen),
```
cawo-odoo:
```suggestion
        def gen():
            for ids in _search_ids(cr, like_all=like_all, like_any=like_any):
                for chunk in chunks(ids, 1000):
                    yield chunk

        # consume the submitted tasks
        collections.deque(
            executor.map(partial(_mp_callback, cr.dbname, callback), gen()),
```
Or sth. similar to still achieve good parallelization in cases where there is lots of ids with very short commands. Maybe add a `chunk_size` argument to `multiprocess_commands` instead of hardcoding `1000`. Of course it would be even nicer if a second cap to the `_search_ids` query in addition to `MEMORY_CAP` could be added instead of "sub-chunking" here.

aj-fuentes: I think we could put a hard cap to the number of ids per chunk returned in _search_ids. We group up to the max memory size or the max number of items. We can then pass-through the parameters --like `memory_limit=MEMORY_CAP, max_count=COUNT_CAP`-- from the outermost entry points (multiprocess_commands)

src/util/spreadsheet/misc.py
```diff
@@ -1,69 +1,103 @@
...
+            condition=pg.SQLStr("ALL" if like_all else "ANY"),
+        ),
+        {
+            "like": list(like_any or like_all),
+            "mem_cap": MEMORY_CAP,
+            "count_cap": COUNT_CAP,
+        },
+    )
+    for (ids,) in cr.fetchall():
+        yield ids
```
cawo-odoo: Don't we want progress logging?
```suggestion
    return log_progress(
        (ids for (ids,) in cr.fetchall()),
        logger,
        qualifier="chunks",
        size=cr.rowcount,
        log_hundred_percent=True
    )
```

EDIT: this is the wrong place, since `executor.map` is eager, it would log 100% very quickly.

aj-fuentes: I think we just need to wrap `executor.map` with `log_progress` below (untested :/)

KangOl: Actually, even with our big database, it now runs for a little bit more than 1 minute (for the last saas~19.3 script), so I don't think adding logging is needed.

KangOl: > I think we just need to wrap `executor.map` with `log_progress` below (untested :/)

Doesn't work :(
> TypeError: object of type 'generator' has no len()

aj-fuentes: `size=....`?

Anyway, I think logging isn't critical.

KangOl: We can't get it without `_search_ids` returning the whole result without yielding individual rows.

cawo-odoo: > Actually, even with our big database, it now runs for a little bit more than 1 minute (for the last saas~19.3 script), so I don't think adding logging is needed.

That's a great result :) We can add logging later if it will ever be useful.

src/util/spreadsheet/misc.py
```diff
@@ -1,69 +1,104 @@
...
+            condition=SQLStr("ALL" if like_all else "ANY"),
+        ),
+        {
+            "like": list(like_any or like_all),
+            "mem_cap": MEMORY_CAP,
+            "count_cap": COUNT_CAP,
+        },
+    )
+    for (ids,) in cr.fetchall():
+        yield ids
```
aj-fuentes: Why is this a generator? `cr.fetchall` already fetches all the info.
It makes sense to keep it as a generator if we use a named cursor as before 🤔 
```suggestion
    return [ids for (ids,) in cr.fetchall()]
```

src/util/spreadsheet/misc.py
```diff
@@ -72,3 +107,42 @@ def process_commands(cr, callback, *args, **kwargs):
...
+        initializer=init_worker_process,
+        mp_context=multiprocessing.get_context("fork"),
+    ) as executor:
+        gen = _search_ids(cr, like_all=like_all, like_any=like_any)
+        tasks = executor.map(partial(_mp_callback, cr.dbname, callback), gen)
+        # consume the submitted tasks
+        collections.deque(tasks, maxlen=0)
+    # Another commit() is necessary as we are in `REPEATABLE READ`, so the new transaction started by `_search_ids`, won't see the
+    # changes done by the workers. Note that creating `gen` before the commit above won't help as it's code won't be executed before
+    # the first iteration.
```
aj-fuentes: I don't understand this comment. _search_ids could return the result of fetchall eagerly.

cawo-odoo: I was thinking the same, the comment is just overly complicated. The commit is simply needed to get a new transaction in the parent process to see the changes made on transactions opened (and closed) by the child processes. It has nothing to do with when the generator code runs AFAICT.

aj-fuentes: Makes sense. I was wondering why this was different from our parallel_execute.
The key is that parallel_execute reuses the same connection
https://github.com/odoo/upgrade-util/blob/2899bb6fbad849565544d4c2cbd389f2e47d1699/src/util/pg.py#L111
But here the connection on each worker is different, we force that via `sql_db._Pool = None` (we know otherwise we get a Protocol error in case of reuse).
Finally, `cr.commit()` in Odoo just defers to `cr._cnx.commit()`
https://github.com/odoo/odoo/blob/e17bcf65dcf0d9db3f8c7decff794c4e6c44e3b8/odoo/sql_db.py#L557

So, it isn't just the transaction isolation, but all those details. I think it's word summarizing them in a comment indeed.

src/util/spreadsheet/misc.py
```diff
@@ -72,3 +108,41 @@ def process_commands(cr, callback, *args, **kwargs):
...
+        raise ValueError("Please specify `like_all` or `like_any`, not both")
+
+    def init_worker_process():
+        sql_db._Pool = None
+
+    callback = make_pickleable_callback(callback)
+
+    it = _search_ids(cr, like_all=like_all, like_any=like_any)
+
+    cr.commit()
```
aj-fuentes, line 142: Just a tiny clarification for the future...
In any case just to be sure I now fully understand this.
```suggestion
    # Commit here achieves two goals:
    # 1. Commit transaction such that the workers can see changes.
    # 2. Ensure first query after the call to `multiprocess_commands` starts a new transaction
    #    and thus can _see_ the changes from the workers.
    cr.commit()
```

aj-fuentes (approved):
LGTM

src/util/spreadsheet/misc.py
cawo-odoo: > I was wondering why this was different from our parallel_execute.
> The key is that parallel_execute reuses the same connection

Actually, that is not the reason (and not correct). While psycopg2 `Cursor`s created from the same connection share the same transaction, this is not true for Odoo `Cursor`s. Each Odoo Cursor borrows a connection from the pool exclusively and when used in a `with`, will [commit](https://github.com/odoo/odoo/blob/b575bdc8cb71bf33a94ae2090e847bc37488364a/odoo/sql_db.py#L557) the transaction on that connection when exiting the scope of `with`. This commit() will leave the connection in status [CONN_STATUS_READY](https://github.com/psycopg/psycopg2/blob/3a6d9d6ddc6b53eaa80b712f5fa6b23abbdc38db/psycopg/pqpath.c#L421), only the next command sent by a new cursor on the same connection will first send `BEGIN;`. This is why the commit() before the threads start is enough for the cursor on the main thread to see the results when it is first used after the threads all committed all their Odoo Cursors (i.e. psycopg2 Connections).

I think you must have understood this in the meanwhile and thus your last suggestion https://github.com/odoo/upgrade-util/pull/428#discussion_r3208654302 , so this message is more for me to note the full picture.

cawo-odoo (approved):
🐈‍⬛

KangOl:
@robodoo r+

PR #435, opened 2026-05-27
[IMP] util.explode_query_range

KangOl:
Randomize the generated queries.
As we split queries based on their ids, it's more likely than consecutive queries (i.e. `id BETWEEN 101 AND 200` and `id BETWEEN 201 AND 300`) will hit the same page when being executed by two concurrent PG processes, blocking themself with read and write locks. By shuffling the queries, we avoid such problem.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -154,7 +154,7 @@ else:
     _parallel_execute_threaded = _parallel_execute_serial
 
 
-def parallel_execute(cr, queries, logger=_logger, qualifier="queries"):
+def parallel_execute(cr, queries, logger=_logger, qualifier="queries", order="verbatim"):
     """
     Execute queries in parallel.
 
@@ -169,6 +169,7 @@ def parallel_execute(cr, queries, logger=_logger, qualifier="queries"):
     :param list(str) queries: list of queries to execute concurrently
     :param `~logging.Logger` logger: logger used to report the progress
     :param str qualifier: qualifier of the queries. Used in the progression log
+    :param str order: order of processing of the queries. Accepted values are "verbatim" or "shuffle".
     :return: the sum of `cr.rowcount` for each query run
     :rtype: int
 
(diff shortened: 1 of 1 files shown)


KangOl:

src/util/pg.py
```diff
@@ -379,9 +380,11 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
     parallel_filter = "{alias}.id BETWEEN %(lower-bound)s AND %(upper-bound)s".format(alias=alias)
     query = _explode_format(query.replace("%", "%%"), parallel_filter=parallel_filter)
 
-    return [
+    queries = [
```
Pirols:
```suggestion
    n_ids = len(ids) - 1
    modulo = n_ids // 3
    return [
        cr.mogrify(query, {"lower-bound": ids[i], "upper-bound": ids[i + 1] - 1}).decode() 
        for i in sorted(range(n_ids), key=lambda x: (x % modulo, x))
    ]
```

Alternative idea to make it deterministic. Though slightly less efficient.

Or if we want to go random, why don't we just shuffle the numbers?

```suggestion
    n_ids = len(ids) - 1
    return [
        cr.mogrify(query, {"lower-bound": ids[i], "upper-bound": ids[i + 1] - 1}).decode() 
        for i in random.sample(range(n_ids), n_ids)
    ]
```

src/util/pg.py
```diff
@@ -379,9 +380,11 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
...
         cr.mogrify(query, {"lower-bound": ids[i], "upper-bound": ids[i + 1] - 1}).decode() for i in range(len(ids) - 1)
     ]
+    random.shuffle(queries)
```
aj-fuentes: I think patching this method is not ideal. We should keep it returning the queries in the stable order. We should instead patch the executing part. I'd add a parameter to `parallel_execute` that signals whether the queries are chunked in consecutive chunks and so we may benefit from a "better" ("randomized" to keep it simple) dispatcher. `explode_execute` could set the param to True. We could then set it in places where we know it helps for older scripts as well. In most cases just switching to `explode_execute` would be enough.

aj-fuentes: Note that by altering the order at execution time we expose ourselves to variations in running time for big tables that are scattered on disk. I cannot say how "significant", but this makes me think that using a random strategy may haunt us later.

PR #437, opened 2026-05-29
[FIX] util: invalidate stale cache before updating

sagu-odoo:
Issue:
A cache desynchronization can happen when a record is first loaded in the ORM cache and later updated directly using `cr.execute()`.

In that case, the database value changes, but the ORM cache still keeps the old value because the record is not reloaded or invalidated.

When `update_record_from_xml()` is called again for the same record, `_update_cache()` is not triggered because the cached value is still considered identical [1] [2].

As a result:
- the ORM keeps stale values in cache
- the field is not marked dirty
- flush/commit only synchronizes the dirtly field [3]

Fix:
Invalidate the environment cache before calling
`update_record_from_xml()` again so the ORM reloads fresh values from the database before performing the write.

[1]: https://github.com/odoo/odoo/blob/8d14665af5acf1bd391d05a5048dc701986e8b15/odoo/orm/fields.py#L1629-L1630
[2]: https://github.com/odoo/odoo/blob/8d14665af5acf1bd391d05a5048dc701986e8b15/odoo/orm/fields.py#L1520
[3]: https://github.com/odoo/odoo/blob/8d14665af5acf1bd391d05a5048dc701986e8b15/odoo/orm/environments.py#L384

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1760,6 +1760,28 @@ class TestRecords(UnitTestCase):
         self.assertEqual(usd.symbol, "$")
         self.assertEqual(usd.name, "XXX")
 
+    def test_update_record_from_xml_cache(self):
+        cr = self.env.cr
+        xmlid = "base.action_attachment"
+
+        record_id = util.ref(cr, xmlid)
+        query = "SELECT name FROM ir_act_window WHERE id = %s"
+        if util.column_type(cr, "ir_act_window", "name") == "jsonb":
+            query = "SELECT name->>'en_US' FROM ir_act_window WHERE id = %s"
+
+        cr.execute(query, [record_id])
+        original = cr.fetchone()[0]
+
+        # Load, SQL update, then load shouldn't skip the last load override
+        util.update_record_from_xml(cr, xmlid)
+        cr.execute("""UPDATE ir_act_window SET name = '{"en_US": "hack"}' WHERE id = %s""", [record_id])
+        util.update_record_from_xml(cr, xmlid)
+
+        # Ensure last load restored the value
+
+        cr.execute(query, [record_id])
+        self.assertEqual(cr.fetchone()[0], original)
+
     def test_ensure_xmlid_match_record(self):
         cr = self.env.cr
         tx1 = self.env["res.currency"].create({"name": "TX1", "symbol": "TX1"})
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -1242,7 +1242,16 @@ def __update_record_from_xml(
             done_refs=done_refs,
         )
 
-    cr_or_env = env(cr) if version_gte("saas~16.2") else cr
+    new_env = env(cr)
+
+    if version_gte("saas~15.4"):
+        new_env.invalidate_all()
+    elif version_gte("11.0"):
+        new_env.cache.invalidate()
+    elif version_gte("8.0"):
+        new_env.invalidate_all()
+
+    cr_or_env = new_env if version_gte("saas~16.2") else cr
     parse_kw = {"mode": "update"} if version_between("8.0", "12.0") else {}
     for xml_filename, root in roots.items():
         kw = {"xml_filename": xml_filename} if version_gte("8.saas~6") else {}
@@ -1250,7 +1259,7 @@ def __update_record_from_xml(
         importer.parse(root, **parse_kw)
 
     if version_gte("13.0"):
-        flush(env(cr)["base"])
+        flush(new_env["base"])
 
     if noupdate:
         force_noupdate(cr, xmlid, noupdate=True)


aj-fuentes:
I was working in something very similar. https://github.com/odoo/upgrade-util/pull/438

Created the PR just to compare, I think the ideas are the same. I didn't test in older versions...

src/util/records.py
```diff
@@ -1241,16 +1241,19 @@ def add_ref(ref):
             fields=None,
             done_refs=done_refs,
         )
+    upgrade_env = env(cr) if version_gte("saas~16.2") else None
+    cr_or_env = upgrade_env or cr
+    if upgrade_env:
+        upgrade_env.invalidate_all()
```
KangOl: Actually, we should always invalidate, not only from saas~16.2

KangOl:
We need to test it in all supported versions.

A test can be added with the sequence done in #438 (call, UPDATE, call).

sagu-odoo:
> We need to test it in all supported versions.

Okay i will check and adapt accordingly. 


Thanks for the review @KangOl and @aj-fuentes .

sagu-odoo:

src/util/records.py
```diff
@@ -1242,15 +1242,33 @@ def add_ref(ref):
...
         kw = {"xml_filename": xml_filename} if version_gte("8.saas~6") else {}
         importer = xml_import(cr_or_env, from_module, idref={}, mode="update", **kw)
         importer.parse(root, **parse_kw)
 
     if version_gte("13.0"):
-        flush(env(cr)["base"])
+        if version_gte("saas~16.2"):
+            flush(cr_or_env["base"])
+        else:
+            flush(new_env["base"])
```
KangOl: You can use `new_env` in all cases.
```suggestion
        flush(new_env["base"])
```

src/base/tests/test_util.py
```diff
@@ -1760,6 +1760,26 @@ def test_update_record_from_xml__fields(self):
...
+        name_column = "name || 'copy'"
+        if util.column_type(cr, "ir_act_window", "name") == "jsonb":
+            name_column = "jsonb_build_object('en_US', name->>'en_US' || 'copy')"
+        query = "UPDATE ir_act_window SET name = {name_column} WHERE id=%s"
+        cr.execute(util.format_query(cr, query, name_column=util.SQLStr(name_column)), [record.id])
+
+        util.update_record_from_xml(cr, xmlid)
+        util.invalidate(record)
+
+        self.assertEqual(record.name, original_help)
```
KangOl: Not sure the tests actually catch the issue. Does it fail if you patch in `records.py` is reverted?

sagu-odoo: image

yes, it do fail like this.  if anything more need to check or anything that need to clear. Please do let me know.

src/base/tests/test_util.py
```diff
@@ -1760,6 +1760,26 @@ def test_update_record_from_xml__fields(self):
...
+        name_column = "name || 'copy'"
+        if util.column_type(cr, "ir_act_window", "name") == "jsonb":
+            name_column = "jsonb_build_object('en_US', name->>'en_US' || 'copy')"
+        query = "UPDATE ir_act_window SET name = {name_column} WHERE id=%s"
+        cr.execute(util.format_query(cr, query, name_column=util.SQLStr(name_column)), [record.id])
+
+        util.update_record_from_xml(cr, xmlid)
+        util.invalidate(record)
+
+        self.assertEqual(record.name, original_name)
```
aj-fuentes: This could be simpler and more in line with what we want: consistency between the operations and what's **in the DB**. Thus I'd suggest use direct SQL as much as possible.
```suggestion
        record_id = util.ref(cr, xmlid)

        cr.execute("SELECT name::text FROM ir_act_window WHERE id=%s", [record_id])
        original = cr.fetchone()[0]

        # Load, SQL update, then load shouldn't skip the last load override
        util.update_record_from_xml(cr, xmlid)
        cr.execute("""UPDATE ir_act_window SET name = '{"en_US": "hack"}'""")
        util.update_record_from_xml(cr, xmlid)

        # Ensure last load restored the value
        cr.execute("SELECT name::text FROM ir_act_window WHERE id=%s", [record_id])
        self.assertEqual(cr.fetchone()[0], original)
```

src/util/records.py
```diff
@@ -1242,15 +1242,30 @@ def add_ref(ref):
...
+    #   https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/api.py#L864-L870
+    # - 11.0 to saas~15.3: invalidate_all() was removed and replaced by
+    #   cache.invalidate().
+    #   https://github.com/odoo/odoo/commit/32a58c0db3c0d666e704a4c8af33b0be0a9608e4
+    # - saas~15.4+: invalidate_all() was reintroduced.
+    #   https://github.com/odoo/odoo/commit/32bc28aa665c479e31a40afffca2f03f6f2cc805
+    if version_between("8.0", "10.0") or version_gte("saas~15.4"):
+        new_env.invalidate_all()
+    if version_between("11.0", "saas~15.3"):
+        new_env.cache.invalidate()
```
aj-fuentes: I think you can leave the comments out. We need to do this all the time. If any put it in the commit message.
The code can be simpler.
```suggestion
    if version_gte("saas~15.4"):
        new_env.invalidate_all()
    elif version_gte("11.0"):
        new_env.cache.invalidate()
    elif version_gte("8.0"):
        new_env.invalidate_all()    
```

From local testing with a 0.0.0 upgrade script instead of a test case, it seems that this is only an issue from 13.0+. I haven't checked in detail, and I don't discard a mistake on my side. Can you experiment if the cache issue happens in an upgrade 12->12.3 or 11->12? For me locally it works fine without any patch here. The reason I ask is to be sure what changed. Also this could simplify the patch further if we are sure in older versions there is no such issue. For example:
```py
    # <short comment why we need it only from 13+>
    if version_gte("saas~15.4"):
        new_env.invalidate_all()
    elif version_gte("13.0"):
        new_env.cache.invalidate()  
```

aj-fuentes: From DM with CHS: it's probably due to the "dirty" mechanism starting from 13.0+. In any case invalidating the cache always seems fine even for older versions.

src/base/tests/test_util.py
sagu-odoo: the idea was to do same as happening in ORM but what you have suggested also simple and clean.

aj-fuentes: To me the idea of the test should be the redo what we saw in upgrade scripts that is problematic. In this case it is load, sql update, load -> desync. That's why I suggested writing this from the pov of an upgrade script (that uses `cr`) instead of overusing the ORM.

sagu-odoo:
Hello @aj-fuentes  and @KangOl , suggested changes done could you please have a look ?.

Apart from this, https://runbot.odoo.com/runbot/build/112621071 build always fail  there may be an issue with the docker image for this particular version. should i report to the runbot team again ?. If I recall correctly, this was reported some time ago already.

aj-fuentes:
The upgrade from 13 is broken in runbot since a while ago. You can report it but probably it should be removed instead. Rely more on matt for this.

aj-fuentes (approved):

KangOl (approved):
@robodoo r+

PR #439, opened 2026-06-02
[IMP] testing: make `IntegrityCase` a `TransactionCase`

aj-fuentes:
Since 19.4 test cursors are available only for `TransactionCase`, and not for `BaseCase`. With the many changes over the years it seems like we can make `IntegrityCase` inherit `TransactionCase`. This change may ease the upcoming refactoring in the test machinery in the ORM as well.

### DIFF
--- a/src/testing.py
+++ b/src/testing.py
@@ -191,7 +191,7 @@ def parametrize(argvalues):
     return decorator
 
 
-def _create_meta(sequence: int, *tags: str) -> type:
+def _create_meta(sequence: int, *tags: str, **kwargs) -> type:
     if MetaCase:
 
         class UpgradeMetaCase(MetaCase):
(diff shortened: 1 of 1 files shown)


src/testing.py
```diff
@@ -593,7 +593,10 @@ def test_prepare(self):
 
 
 # pylint: disable=inherit-non-class
-class IntegrityCase(UpgradeCommon, _create_meta(20, "integrity_case")):
+__base = TransactionCase if util.version_gte("saas~19.1") else BaseCase
+
+
+class IntegrityCase(UpgradeCommon, _create_meta(20, "integrity_case", base=__base)):
```
KangOl, line 599: Even if we don't use pylint anymore, keep the pragma comment on the class.
```suggestion
# pylint: disable=inherit-non-class
class IntegrityCase(UpgradeCommon, _create_meta(20, "integrity_case", base=__base)):
```

aj-fuentes: Done.

KangOl (approved):
@robodoo r+ priority

PR #440, opened 2026-06-03
[IMP] util.log_progress: now works with generators

KangOl:
Log the count of elements processed so for, without percentage nor ETA, obviously.

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -366,8 +366,13 @@ def chunks(iterable, size, fmt=None):
 
 def log_progress(it, logger, qualifier="elements", size=None, estimate=True, log_hundred_percent=False):
     if size is None:
-        size = len(it)
+        try:  # noqa: SIM105
+            size = len(it)
+        except TypeError:
+            # object of type 'generator' has no len()
+            pass
     t0 = t1 = datetime.datetime.now()
+    i = 0
     for i, e in enumerate(it, 1):
         yield e
         t2 = datetime.datetime.now()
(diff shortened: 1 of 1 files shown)


src/util/misc.py
```diff
@@ -366,31 +366,45 @@ def chunks(iterable, size, fmt=None):
 
 def log_progress(it, logger, qualifier="elements", size=None, estimate=True, log_hundred_percent=False):
     if size is None:
-        size = len(it)
+        try:  # noqa: SIM105
+            size = len(it)
+        except TypeError:
+            # object of type 'generator' has no len()
+            pass
```
cawo-odoo, line 373:
```suggestion
        with suppress(TypeError):
            size = len(it)
```

src/util/misc.py
```diff
@@ -366,31 +366,45 @@ def chunks(iterable, size, fmt=None):
...
         yield e
         t2 = datetime.datetime.now()
         secs_last, secs_start = (t2 - t1).total_seconds(), (t2 - t0).total_seconds()
         if secs_last > 60 or (log_hundred_percent and i == size and secs_start > 10):
             t1 = datetime.datetime.now()
             tdiff = t2 - t0
-            j = float(i)
-            if estimate:
-                tail = " (total estimated time: %s)" % (datetime.timedelta(seconds=tdiff.total_seconds() * size / j),)
+            if size is not None:
```
cawo-odoo, line 383: Don't we want to keep logging every 60s the cummulative count, just without total and percent?

KangOl: We do. It;s the `else` part of this `if`.

src/util/misc.py
KangOl: No. This needs to be python2 compatible.
And we cannot use the odoo fallback
https://github.com/odoo/upgrade-util/blob/962a5038d14fbdb010e0a624ad0df633bb800e5d/src/util/domains.py#L12-L21

aj-fuentes: > Added in version 3.4.

https://docs.python.org/3/library/contextlib.html#contextlib.suppress

:/

src/util/misc.py
cawo-odoo: oh, yes ... i misread.

cawo-odoo (approved):

src/util/misc.py
```diff
@@ -366,31 +366,45 @@ def chunks(iterable, size, fmt=None):
...
+                j = float(i)
+                if estimate:
+                    tail = " (total estimated time: %s)" % (
+                        datetime.timedelta(seconds=tdiff.total_seconds() * size / j),
+                    )
```
aj-fuentes: nit: since we are already touching it, let's get rid of the % formatting.
```suggestion
                    estimate = datetime.timedelta(seconds=tdiff.total_seconds() * size / j)
                    tail = " (total estimated time: {})".format(estimate)
```

aj-fuentes (approved):

KangOl:

KangOl:
@robodoo r+

PR #441, opened 2026-06-04
[IMP] snippets: add a wrapper method

rushil-b-patel:
Community PR: https://github.com/odoo/odoo/pull/261797
Enterprise PR: https://github.com/odoo/enterprise/pull/117088
Upgrade: https://github.com/odoo/upgrade/pull/10322

### DIFF
--- a/src/util/snippets.py
+++ b/src/util/snippets.py
@@ -210,6 +210,38 @@ def format_style(styles):
     return style
 
 
+def wrap_element(el, wrapper_tag="div", move_classes=(), move_attributes=(), strip_classes=(), wrapper_classes=()):
+    """
+    Wrap an HTML element, and move classes or attributes onto that wrapper.
+
+    :param el: the HTML element to wrap
+    :param str wrapper_tag: tag name of the new wrapping element
+    :param tuple move_classes: classes to remove from ``el`` and add to the wrapper
+    :param tuple move_attributes: attribute names to pop from ``el`` and copy onto the wrapper
+    :param tuple strip_classes: classes to remove from ``el``
+    :param tuple wrapper_classes: classes to add to the wrapper
+    :return: the new wrapping element
+    """
+    wrapper_attrs = {}
+    wrapper_class_list = move_classes + wrapper_classes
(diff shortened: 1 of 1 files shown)


src/util/snippets.py
```diff
@@ -210,6 +210,32 @@ def format_style(styles):
...
+    wrapper = el.makeelement(wrapper_tag, wrapper_attrs)
+    el.addprevious(wrapper)
+    wrapper.append(el)
+    if move_classes:
+        drop = set(move_classes)
+        el.set(
+            "class",
+            " ".join(c for c in el.get("class", "").split() if c not in drop),
+        )
+    return True
```
KangOl:
```suggestion
    return wrapper
```

src/util/snippets.py
```diff
@@ -210,6 +210,34 @@ def format_style(styles):
...
+    :param tuple move_attributes: attribute names popped from ``el`` and copied onto the wrapper
+    :return: the new wrapping element
+    """
+    wrapper_attrs = {}
+    if move_classes:
+        wrapper_attrs["class"] = " ".join(move_classes)
+        drop = set(move_classes)
+        el.set(
+            "class",
+            " ".join(c for c in el.get("class", "").split() if c not in drop),
```
aans-odoo, line 233: why use `drop` (set) we can directly use `move_classes` (tuple) here.

aans-odoo (changes requested):

src/util/snippets.py
rushil-b-patel: with `drop` (set) we get O(1) lookup while with `move_classes` (tuple) its O(n) per check.

aans-odoo: Fair point, but set construction itself is O(no_of_classes), so total is O(no_of_classes + classes_to_move) vs O(no_of_classes * classes_to_move) for the tuple. The set only wins when no_of_classes is large enough to offset the build cost. In a real-world scenario both are small numbers, so let's keep what we have.

aans-odoo (approved):

PR #442, opened 2026-06-04
[IMP] handle new `ir.access` model

KangOl:
See odoo/odoo#166359

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -233,30 +233,25 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
             ids = tuple(c for (c,) in cr.fetchall())
             cr.execute("DELETE FROM ir_model_data WHERE model = 'ir.model.constraint' AND res_id IN %s", [ids])
 
-        # Drop XML IDs of ir.rule and ir.model.access records that will be cascade-dropped,
+        # Drop XML IDs of ir.rule/ir.model.access/ir.access records that will be cascade-dropped,
         # when the ir.model record is dropped - just in case they need to be re-created
-        cr.execute(
-            """
-                DELETE
-                  FROM ir_model_data x
-                 USING ir_rule a
-                 WHERE x.res_id = a.id
-                   AND x.model = 'ir.rule'
-                   AND a.model_id = %s
-        """,
-            [mod_id],
-        )
-        cr.execute(
-            """
-                DELETE
-                  FROM ir_model_data x
-                 USING ir_model_access a
-                 WHERE x.res_id = a.id
-                   AND x.model = 'ir.model.access'
-                   AND a.model_id = %s
-        """,
-            [mod_id],
-        )
+        for cascade_model in ["ir.access", "ir.rule", "ir.model.access"]:
+            table = table_of_model(cr, cascade_model)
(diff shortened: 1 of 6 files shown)


src/util/orm.py
```diff
@@ -833,7 +838,12 @@ def _add_magic_fields(self):
                 """
             )
         if disabled_ir_rule_ids:
-            env.cr.execute("UPDATE ir_rule SET active = 't' WHERE id IN %s", (tuple(disabled_ir_rule_ids),))
+            query = format_query(
+                env.cr,
+                "UPDATE ir_rule SET active = true WHERE id = ANY(%s)",
```
aj-fuentes:
```suggestion
                "UPDATE {} SET active = true WHERE id = ANY(%s)",
```
format_query dangerously swallows parameters :/

KangOl:

rco-odoo:
@KangOl I had to modify a little bit `remove_model` to deal with the removal of models `ir.model.access` and `ir.rule` themselves.

src/util/models.py
```diff
@@ -233,30 +233,27 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
-                 USING ir_model_access a
-                 WHERE x.res_id = a.id
-                   AND x.model = 'ir.model.access'
-                   AND a.model_id = %s
-        """,
-            [mod_id],
+        cascade_models = (
+            ["ir.access"] if table_exists(cr, "ir_access") else
+            ["ir.rule", "ir.model.access"] if not version_gte("saas~19.3") else
+            []
```
aj-fuentes: Why do we consider an empty list? Not >= 19.3 seems wrong, it would keep an empty list for 19.3 :/

IMO we should set the _future_, set the _past_, and add a special case for the transition state that happens in 19.4-20.0.

What I have in mind is the possible reworks and new scripts in 19.4/5, 20.0, that may miss that that multiple tables could coexist. For example in the linked upgrade PR the script is pre-20 in base, while module operations happen in pre-10. This is just an example, further confusion could happen in other scripts as well (including symlinked, or 0.0.0, ones).

Untested, idea:
```suggestion
        cascade_models = ["ir.access"]  # future
        if not version_gte("saas~19.4"):  # < saas~19.4, past
            cascade_models = ["ir.rule", "ir.model.access"]
        elif not version_gte("saas~20.1"): # >= saas~19.4, <= 20.0, transition
            cascade_models = [mod for mod in ["ir.access", "ir.rule", "ir.model.access"] if table_exists(cr, table_of_model(mod))]
```

KangOl: To avoid the transition code, we can change the upgrade script to remove the old models in `post-` (at least the tables).

aj-fuentes: That works for current linked PRs, but what I have in mind is the upcoming 19.5/20 scripts. We will have some in pre- in base there. Or some 0.0.0. We can argue that we can deal with issues once they happen, I don't disagree, I just think that having the compatibility layer here for a while may help future us --in the 20.0 merge rush :/

rco-odoo: OMG you're right.  Thanks for the suggestion!

aj-fuentes:
@robodoo delegate=rco-odoo

PR #443, opened 2026-06-05
[FIX] util.pg.*: don't use the `_cnx` attribute.

KangOl:
`quote_ident` and `as_string` accept as (psycopg2) cursor as argument, so do it.
This allow using such functions with an Odoo TestCursor as, [as of saas-19.3], their `_cnx` attribute is not psycopg2 connection object, but a wrapper.

[as of saas-19.3]: https://github.com/odoo/odoo/commit/5a0c5b21729eade84352279f7cbd174ea67fa6aa


aj-fuentes (approved):
LGTM

KangOl:

KangOl:
@robodoo r+

PR #444, opened 2026-06-05
[IMP] util.create_column: modernize

KangOl:
- Use `format_query`
- Use `.format()` instead of %-format.


KangOl:

aj-fuentes (approved):
Can be r+ed

PR #445, opened 2026-06-05
[IMP] util/misc: add `util.once`

aj-fuentes:
This allows to run code only once in an interval. The new util can be
used as a boolan or as a decorator.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -2448,6 +2448,99 @@ def not_doing_anything_converter(el):
     return True
 
 
+class TestOnce(UnitTestCase):
+    # Each case: (source, lower, upper, steps_and_expected)
+    # source: the DB version before the upgrade starts target is inferred as the last step in steps_and_expected
+    # steps_and_expected: list of (series, expected), the steps visited in order.
+    # Exactly one step should be truthy (the first step whose dest lands in [lower, upper]);
+    # all others falsy. No-intersection cases have all falsy.
+    @parametrize(
+        [
+            # Default interval: fire on first step after source
+            ("16.0", None, None, [("17.0", True), ("18.0", False)]),
+            ("16.2", None, None, [("17.0", True), ("18.0", False)]),
+            ("saas~16.3", None, None, [("17.0", True), ("18.0", False)]),
+            ("16.0", None, None, [("17.0", True)]),
+            # Source minor to target minor across a major: first major .0 fires
+            ("18.4", None, None, [("19.0", True), ("saas~19.3", False)]),
+            ("saas~18.2", None, None, [("19.0", True)]),
+            # Same-major minor-to-minor: single step, fires
+            ("saas~19.0", None, None, [("saas~19.3", True)]),
+            ("saas~19.1", None, None, [("saas~19.3", True)]),
+            # Explicit interval covering the first step
+            ("16.0", "17.0", "18.0", [("17.0", True), ("18.0", False)]),
+            ("16.0", "17.0", None, [("17.0", True), ("18.0", False)]),
+            # Explicit interval starting mid-path
+            ("16.0", "18.0", "18.0", [("17.0", False), ("18.0", True)]),
+            ("16.0", "17.0", "18.0", [("17.0", True), ("18.0", False), ("19.0", False)]),
+            ("16.0", "17.0", "19.0", [("17.0", True), ("18.0", False), ("19.0", False)]),
+            ("16.0", "saas~17.2", None, [("17.0", False), ("18.0", True), ("19.0", False)]),
+            (
+                "16.0",
+                "saas~17.2",
+                "saas~18.4",
+                [("17.0", False), ("18.0", True), ("19.0", False), ("saas~19.3", False)],
+            ),
+            ("16.0", "17.0", "saas~18.4", [("17.0", True), ("18.0", False), ("19.0", False), ("saas~19.3", False)]),
+            ("18.4", "saas~19.0", "saas~19.3", [("19.0", True), ("saas~19.3", False)]),
+            ("saas~18.2", "saas~18.2", "saas~19.3", [("19.0", True), ("saas~19.3", False)]),
+            ("saas~16.3", "saas~16.3", "18.0", [("17.0", True), ("18.0", False)]),
+            # saas~12.3 mandatory stopover: it is an extra step between 12.0 and 13.0
+            ("10.0", None, None, [("11.0", True), ("12.0", False), ("saas~12.3", False), ("13.0", False)]),
+            ("11.0", None, None, [("12.0", True), ("saas~12.3", False), ("13.0", False)]),
+            ("11.0", "saas~12.3", "saas~12.3", [("12.0", False), ("saas~12.3", True), ("13.0", False)]),
+            ("11.0", "12.0", "saas~12.3", [("12.0", True), ("saas~12.3", False), ("13.0", False)]),
+            ("10.0", "saas~12.3", "13.0", [("11.0", False), ("12.0", False), ("saas~12.3", True), ("13.0", False)]),
+            # Interval entirely outside the upgrade path: never fire
+            ("16.0", "saas~17.2", "saas~17.4", [("17.0", False), ("18.0", False)]),
+            ("16.0", "18.0", "19.0", [("17.0", False)]),
+        ]
+    )
+    def test_once(self, source, lower, upper, steps_and_expected):
+        target = steps_and_expected[-1][0]
+        with mock.patch.object(util.misc, "_SOURCE_VERSION", source), mock.patch.object(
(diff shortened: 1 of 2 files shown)


src/util/misc.py
```diff
@@ -131,6 +132,100 @@ def version_between(a, b):
...
+        if not first_step:
+            self.check = False.__bool__
+            return
+
+        once_step = parse_version(first_step)[:2]
+
+        def check():
+            return parse_version(release.serie)[:2] == once_step
+
+        self.check = check
```
KangOl: Why is it a callable?
```suggestion
        self._value = parse_version(release.serie)[:2] == once_step
```

src/base/tests/test_util.py
```diff
@@ -2448,6 +2448,81 @@ def not_doing_anything_converter(el):
...
+                    action()
+                    self.assertEqual(calls, [1] if expected else [], "decorator: " + msg)
+
+    def test_once_no_env_vars(self):
+        # When source/target are unknown, once is always truthy regardless of series
+        with mock.patch.object(util.misc, "_SOURCE_VERSION", None), mock.patch.object(
+            util.misc, "_TARGET_VERSION", None
+        ):
+            for series in ("16.0", "17.0", "18.0", "19.0", "saas~19.4"):
+                with mock.patch.object(util.misc.release, "series", series):
```
```suggestion
                with mock.patch.object(odoo.release, "series", series):
```

src/base/tests/test_util.py
```diff
@@ -2448,6 +2448,81 @@ def not_doing_anything_converter(el):
...
+
+                    action()
+                    self.assertEqual(calls, [1] if expected else [], "decorator: " + msg)
+
+    def test_once_no_env_vars(self):
+        # When source/target are unknown, once is always truthy regardless of series
+        with mock.patch.object(util.misc, "_SOURCE_VERSION", None), mock.patch.object(
+            util.misc, "_TARGET_VERSION", None
+        ):
+            for series in ("16.0", "17.0", "18.0", "19.0", "saas~19.4"):
```
KangOl, line 2528: Could be done with `@parametrize`

KangOl (changes requested):

KangOl:

src/util/misc.py
aj-fuentes: Done.

src/util/misc.py
```diff
@@ -131,6 +132,99 @@ def version_between(a, b):
...
+
+        a = s if lower is None else parse_version(lower)
+        b = t if upper is None else parse_version(upper)
+        assert a <= b, "invalid interval"
+
+        s_maj = int(_SOURCE_VERSION.replace("saas~", "").replace("saas-", "").split(".", 1)[0])
+        t_parts = _TARGET_VERSION.replace("saas~", "").replace("saas-", "").split(".", 1)
+        t_maj = int(t_parts[0])
+        steps = ["{}.0".format(x) for x in range(s_maj + 1, t_maj + 1)]
+        if t_maj > 15 and t_parts[1] != "0":
```
KangOl: Why `> 15`?

src/base/tests/test_util.py
aj-fuentes: Conceptually this test is testing what one step after the other do. When the env variables are missing. So it's a _single_ test case. Compare with the ones generated above with `parametrize`: each has a list of steps. Same here.

src/util/misc.py
aj-fuentes: See the comment below: rolling release targets. Once could argue that I can put 19 but we support upgrading from 15.2. I put >=15 to be overly strinct --we don't expect those case-- note: we got 16.x already from saas backups from clients.

Note that if we remove the check, the condition for saas-12.3 below needs to be udpated otherwise we may need with saas-12.3 twice in the list.

aj-fuentes: IOW this is just to be strict with the first version of RR. Otherwise we can put 19

KangOl: This should be independent of what the upgrade platform accept as source and target.

src/util/misc.py
```diff
@@ -131,6 +132,99 @@ def version_between(a, b):
...
+
+    :param lower: lower bound of the version interval, inclusive. Defaults to the upgrade source
+                  version when ``None``.
+    :type lower: str or None
+    :param upper: upper bound of the version interval, inclusive. Defaults to the upgrade target
+                  version when ``None``.
+    :type upper: str or None
+    """
+
+    def __init__(self, lower, upper):
```
```suggestion
    def __init__(self, lower, upper, logger=None):
        self._logger = logger
```

src/util/misc.py
```diff
@@ -131,6 +132,99 @@ def version_between(a, b):
...
+        first_step = next((x for x in steps if a <= parse_version(x) <= b), None)
+        if not first_step:
+            self.check = False
+            return
+
+        once_step = parse_version(first_step)[:2]
+        self.check = parse_version(release.serie)[:2] == once_step
+
+    def __bool__(self):
+        return self.check
```
KangOl, line 225:
```suggestion
    def __bool__(self):
        if self._logger:
            op = "run" if self.check else "skip"
            self._logger.info("%s once in version %s", op, release.serie)
        return self.check
```

src/util/misc.py
```diff
@@ -131,6 +132,99 @@ def version_between(a, b):
...
+
+    def __call__(self, func=None):
+        assert callable(func)
+
+        @contextlib.wraps(func)
+        def wrapper(*args, **kwargs):
+            if self.check:
```
```suggestion
            if self:
```

src/util/misc.py
aj-fuentes: Note that current code is independent :)

The first RR version is 15.2. The only valid upgrade targets are major versions, 12.3 (historical), and RR versions (15.2 being the first one).

KangOl (commented):
Proposal

src/util/misc.py
aj-fuentes: I pushed a new variant that doesn't take into account any RR detail.

KangOl: But this is dependent on what we do on SaaS with the current platform. Anyone could do an upgrade with any value set to those env variables.

Even if I don't know why somebody would (beside testing), nothing forbid running an upgrade from i.e. `saas~11.3` to `saas~13.1`.

Having this arbitrary `15` bound seems off.

src/base/tests/test_util.py
```diff
@@ -2448,6 +2448,99 @@ def not_doing_anything_converter(el):
...
+            ("saas~18.2", "saas~18.2", "saas~19.3", [("19.0", True), ("saas~19.3", False)]),
+            ("saas~16.3", "saas~16.3", "18.0", [("17.0", True), ("18.0", False)]),
+            # saas~12.3 mandatory stopover: it is an extra step between 12.0 and 13.0
+            ("10.0", None, None, [("11.0", True), ("12.0", False), ("saas~12.3", False), ("13.0", False)]),
+            ("11.0", None, None, [("12.0", True), ("saas~12.3", False), ("13.0", False)]),
+            ("11.0", "saas~12.3", "saas~12.3", [("12.0", False), ("saas~12.3", True), ("13.0", False)]),
+            ("11.0", "12.0", "saas~12.3", [("12.0", True), ("saas~12.3", False), ("13.0", False)]),
+            ("10.0", "saas~12.3", "13.0", [("11.0", False), ("12.0", False), ("saas~12.3", True), ("13.0", False)]),
+            # Interval entirely outside the upgrade path: never fire
+            ("16.0", "saas~17.2", "saas~17.4", [("17.0", False), ("18.0", False)]),
```
aj-fuentes, line 2495: This case is interesting. For the implementation of `once`, we have two options:
1. Execute once among all steps that **end inside** the internal (current implementation)
2. Execute once among all steps that **intersect** the interval.

For option 2 this test should be:
```suggestion
            ("16.0", "saas~17.2", "saas~17.4", [("17.0", False), ("18.0", True)]),
```

IMO we should stick to option 1. If we need 2/ behavior in one script we can make the interval `[saas~17.2, 18.0]` instead of `[saas~17.2, saas~17.4]`. I leave this comment in case further discussion is deemed worthwhile.

KangOl: I agree.

src/util/misc.py
```diff
@@ -131,6 +132,105 @@ def version_between(a, b):
...
+
+        first_step = next((x for x in steps if a <= parse_version(x) <= b), None)
+        if not first_step:
+            self.check = False
+            return
+
+        once_step = parse_version(first_step)[:2]
+        self.check = parse_version(release.serie)[:2] == once_step
+
+    def __bool__(self):
```
KangOl, line 221: Python2 compatibility
```suggestion
    def __nonzero__(self):
        return self.__bool__()

    def __bool__(self):
```

aj-fuentes: Done

KangOl (approved):
@robodoo r+

PR #446, opened 2026-06-09
[IMP] util.indirect_references

KangOl:
Yield `many2one_reference` fields.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -463,6 +463,33 @@ class TestReplaceReferences(UnitTestCase):
         self.assertEqual(data, expected)
 
 
+class TestIndirectReferences(UnitTestCase):
+    maxDiff = None
+
+    def test_ir(self):
+        cr = self.env.cr
+        triplets = [(ir.table, ir.res_model or "", ir.res_id or "") for ir in util.indirect_references(cr)]
+
+        self.assertEqual(sorted(triplets), sorted(set(triplets)))  # ensure references are only yield once
+
+        expected = [
+            ("ir_attachment", "res_model", "res_id"),
+            ("ir_model_data", "model", "res_id"),
+        ]
+        if util.module_installed(cr, "mail"):
+            expected += [
+                ("mail_followers", "res_model", "res_id"),
+                ("mail_message", "model", "res_id"),
+            ]
+            if util.version_gte("saas~19.2"):
+                # Many2oneReference fields are only reflected in the database from saas~19.2
+                expected.append(("mail_scheduled_message", "model", "res_id"))
+
+        for table, res_model, res_id in expected:
+            with self.subTest(table=table, res_model=res_model, res_id=res_id):
+                self.assertIn((table, res_model, res_id), triplets)
+
+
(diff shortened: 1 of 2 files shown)


KangOl:

src/util/indirect_references.py
```diff
@@ -128,6 +128,41 @@ def indirect_references(cr, bound_only=False):
 
         yield ir
 
+    if ("ir_model_fields", "relation_model_field") in existing:
+        cr.execute(
+            """
+                SELECT m.model, r.name as res_model, f.name as res_id
```
aj-fuentes, line 141: I find the single line way harder to read and spot the relevant info :p
```suggestion
                SELECT m.model, 
                       r.name AS res_model,
                       f.name AS res_id
```

src/util/indirect_references.py
```diff
@@ -128,6 +128,41 @@ def indirect_references(cr, bound_only=False):
...
+                   AND f.store = true
+                   AND r.store = true  -- FIXME the "res_model" column may be a related
+                   AND m.transient = false
+                   AND m.abstract = false
+            """
+        )
+        m2o_refs = [
+            _IR(table_of_model(cr, model), res_model, res_id)
+            for model, res_model, res_id in cr.fetchall()
+        ]  # fmt: skip
```
aj-fuentes, line 158: Why this fmt comment? 🤔

src/base/tests/test_util.py
```diff
@@ -462,6 +462,31 @@ def test_replace_record_references_batch__full_unique(self):
         self.assertEqual(data, expected)
 
 
+class TestIndirectReferences(UnitTestCase):
+    def test_ir(self):
+        cr = self.env.cr
+        triplets = [(ir.table, ir.res_model or "", ir.res_id or "") for ir in util.indirect_references(cr)]
```
aj-fuentes, line 471: Fails. Below what returns the first iteration over INDIRECT_REFERCECES, and the query based output for many2one_reference. Since INDIRECT_REFERECES seems to be more complete I think we need to filter by the first three parameters only. I.ie `res_model_id` is supposed to be redundant wrt to res_model?
```
IndirectReference(table='mail_activity', res_model='res_model', res_id='res_id', res_model_id='res_model_id', set_unknown=False, company_dependent_comodel=None)
IndirectReference(table='mail_activity', res_model='res_model', res_id='res_id', res_model_id=None, set_unknown=False, company_dependent_comodel=None)
```

aj-fuentes (commented):
heck, I forgot to post this...

src/util/indirect_references.py
KangOl: Else `ruff format` will one-line it, which I find harder to read.

src/base/tests/test_util.py
KangOl: > `res_model_id` is supposed to be redundant wrt to res_model?

If they both exists, yes (they may both be defined, but exists depending on the version).

KangOl: I suppose this message was about a previous version of this test...

PR #447, opened 2026-06-10
[IMP] util.snippets: add FieldScope.WEBSITE

robinlej:
`FieldScope` (introduced with [1]) only gave the option to target all
fields (`FieldScope.ALL`), or those depending on `html_builder`
(`FieldScope.SNIPPETS`).
This is not sufficient when we want to target website without targeting
mass_mailing, which is why this commit adds a `FieldScope.WEBSITE`.

[1]: https://github.com/odoo/upgrade-util/commit/f455c099369fc0ef0ef9d84314255bb30e45a8ff

### DIFF
--- a/src/util/snippets.py
+++ b/src/util/snippets.py
@@ -92,6 +92,7 @@ def get_regex_from_snippets_list(snippets):
 class FieldScope(enum.Enum):
     ALL = 1
     SNIPPETS = 2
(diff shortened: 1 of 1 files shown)


robinlej:
Hi @KangOl , would it be ok to merge it?

KangOl (approved):
@robodoo r+

PR #451, opened 2026-06-16
[IMP] util.create_column: use FK column type

MrRose765:
Before, the column type had to be hardcoded even if the column would be used for a foreign key. This could lead to errors when the type of the related column was changed. ([example])

This commit would allow to derive the column type from the related column type if the column is used for a FK.

[example]: https://github.com/odoo/upgrade/pull/10503

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -645,9 +645,10 @@ def create_column(cr, table, column, definition, **kwargs):
 
     :param str table: table of the new column
     :param str column: name of the new column
-    :param str definition: column type of the new column
+    :param str definition: column type of the new column. Use :data:`~odoo.upgrade.util.misc.AUTO`
+                           to infer the type from the `id` column of `fk_table`.
     :param bool default: default value to set on the new column
-    :param bool fk_table: if the new column if a foreign key, name of the foreign table
+    :param str fk_table: if the new column is a foreign key, name of the foreign table
     :param str on_delete_action: `ON DELETE` clause, default `NO ACTION`, only valid if
                                   the column is a foreign key.
     :return: whether the column was created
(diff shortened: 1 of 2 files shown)


src/util/pg.py
```diff
@@ -634,7 +634,7 @@ def _normalize_pg_type(type_):
     return aliases.get(main_type.strip().lower(), main_type) + suffix
 
 
-def create_column(cr, table, column, definition, **kwargs):
+def create_column(cr, table, column, definition=None, **kwargs):
```
aj-fuentes: You need to update the docstring as well.

aj-fuentes (commented):
You need to add a test for this.

We can bikeshed if we can let `definition` be absorbed by `kwargs` or use `auto` as default instead of None. But I think the important parts are already mergeable.

cc: @KangOl

MrRose765:
Thanks for the suggestions

> We can bikeshed if we can let `definition` be absorbed by `kwargs` 

I also wanted to do that but it would require to adapt all call to `create_column()` that did not explicitely use `definition`  parameter (most of them). The current change keep the function signature as before and adds the possibility to infer the FK column type.

aj-fuentes:
> I also wanted to do that but it would require to adapt all call to `create_column()` that did not explicitely use `definition`

Yes, we cannot as this would break existing scripts.

KangOl:

KangOl:
Using `auto` is a good idea, but I would pass it explicitly.
So the code would be
```python
if definition is AUTO:
    if fk_table is no_def:
        raise ValueError("Invalid column definition")
    definition = column_type(cr, fk_table, "id")
else:
    definition = _normalize_pg_type(definition)
```

KangOl:
upgradeci retry

KangOl (approved):
@robodoo r+

PR #452, opened 2026-06-17
[FIX] util/misc: `once` check current version when missing env

aj-fuentes:
`once` is to guard code execution, the bound must be respected even if
there is no env vars.


KangOl:
upgradeci retry with always only stock

aj-fuentes:
robodoo retry

KangOl (approved):
@robodoo r+

KangOl:
@robodoo rebase-ff

PR #455, opened 2026-06-23
[IMP] util/fields: `dump_field_to_chatter` column fallback

aj-fuentes:
For the cases when `mail` is not installed.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -1935,6 +1935,7 @@ def dump_field_to_chatter(
     res_model=None,
     label="%s",
     fk_display=None,
+    fallback_column=None,
     where=None,
     html_escape=True,
     message_type="notification",
@@ -1946,14 +1947,15 @@ def dump_field_to_chatter(
     This is useful when a field is being removed during a migration but clients may need to
     recover its values. The value is posted as a ``mail_message`` on the record's chatter.
 
-    If the ``mail`` module is not installed, the function silently does nothing.
+    If the ``mail`` module is not installed (no ``mail_message`` table), the value is dumped
+    into ``fallback_column`` when one is provided, otherwise the function silently does nothing.
 
     .. example::
        .. code-block:: python
 
           # Simple varchar column
           util.dump_field_to_chatter(
-              cr, "crm_lead", "mobile", label="Previous Mobile", where="phone != mobile"
+              cr, "crm_lead", "mobile", label="Previous Mobile", where="{0}.phone != {0}.mobile"
           )
 
           # Drop team_id, save the team name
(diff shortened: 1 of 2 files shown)


src/util/fields.py
```diff
@@ -1983,13 +1991,22 @@ def dump_field_to_chatter(
...
+                column,
+                fallback_column,
+            )
+        else:
+            _logger.warning(
+                "dump_field_to_chatter: mail_message doesn't exist, cannot dump column %s.%s to chatter",
+                table,
+                column,
+            )
+            return
```
KangOl, line 2023:
```suggestion
        logger = _logger.getChild("dump_field_to_chatter")
        if fallback_column:
            logger.info(
                "mail_message doesn't exist, dump column %s.%s to fallback column %s",
                table,
                column,
                fallback_column,
            )
        else:
            logger.warning(
                "mail_message doesn't exist, cannot dump column %s.%s to chatter",
                table,
                column,
            )
            return
```

src/util/fields.py
```diff
@@ -2069,12 +2084,70 @@ def dump_field_to_chatter(
...
+            """
+            ,_upd AS (
+                SELECT t.{res_id} AS id,
+                       STRING_AGG(format(%(label)s, {col_expr}), %(filler)s ORDER BY t.id) AS value
+                  FROM batch
+                  JOIN {table} t
+                    ON batch.id = t.id
+                  {fk_join}
+                  {value_join}
+              GROUP BY t.{res_id}
```
KangOl: Does it work with the "explode" done on `t.id`?

aj-fuentes: It looks like it doesn't. I think we need to fix the `batch` CTE below. I will check.

aj-fuentes: I pushed a different version. Now with clear different queries. There is a test as well for parallel updates.

src/util/fields.py
```diff
@@ -1974,22 +1976,40 @@ def dump_field_to_chatter(
...
-    :param str where: extra SQL ``WHERE`` condition. By default, only rows where the
-                      column is not ``NULL`` (and not empty for text types) are included.
+    :param str or None fallback_column: column on the same table to dump the value into when
+                                        the ``mail`` module is not installed (no
+                                        ``mail_message`` table). The column must exist (e.g.
+                                        ``res_partner.comment``), raises otherwise. Translated
+                                        (``jsonb``) columns are written through their ``en_US``
+                                        key.
+    :param str where: extra SQL ``WHERE`` condition on the source records. It is a
+                      ``str.format`` template whose ``{0}`` is the source table alias; use it
```
aj-fuentes, line 1997: This is a slight change, since this util was introduced very recently breaking scripts should be minimal. I could introduce a static check `assert where is None or '{0}' in where` to avoid missing it.

KangOl: It may work without the table alias if there are no ambiguities.

KangOl:
upgradeci retry with always only crm hr event

aj-fuentes:
I've rebased to solve conflicts. This should be good to go.

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://masto.es/@magorelax/116913083713266782

PR #457, opened 2026-06-26
[IMP] base/0.0.0: lower logging level when a computed field respawns

xmo-odoo:
NEARLYWARN causes the entry to show up in the runbot logs, where it's a lot of spam and not a lot of utility: https://runbot.odoo.com/runbot/build/115510163

### DIFF
--- a/src/base/0.0.0/end-no-respawn-fields.py
+++ b/src/base/0.0.0/end-no-respawn-fields.py
@@ -50,11 +50,11 @@ def migrate(cr, version):
         qualifier = "field" if store else "non-stored field"
         if transient:
             qualifier = "transient " + qualifier
-        lvl = util.NEARLYWARN if transient or not store else logging.CRITICAL
+        lvl = logging.INFO if transient or not store else logging.CRITICAL
         action = ""
 
         if "{}/{}".format(model, field) in ignored_fields_respawn:
             lvl = util.NEARLYWARN
             action = "; explicitly ignored"
 
-        _logger.log(lvl, "%s %s/%s has respawn%s.", qualifier, model, field, action)
+        _logger.log(lvl, "%s %s/%s has respawned%s", qualifier, model, field, action)


src/base/0.0.0/end-no-respawn-fields.py
```diff
@@ -50,11 +50,11 @@ def migrate(cr, version):
         qualifier = "field" if store else "non-stored field"
         if transient:
             qualifier = "transient " + qualifier
-        lvl = util.NEARLYWARN if transient or not store else logging.CRITICAL
+        lvl = logging.INFO if transient or not store else logging.CRITICAL
```
aj-fuentes, line 53: I think we should make this a full warning if the model is non-transient. IMO ignoring them is hiding potential field removals that shouldn't have been done. I'm not sure if this would block runbot at current state though :/. Still we can keep the NEARLYWARN for stored fields to see what we get.
```suggestion
        lvl = logging.INFO if transient else logging.NEARLYWARN if not store else logging.CRITICAL
```

KangOl:

src/base/0.0.0/end-no-respawn-fields.py
xmo-odoo: See linked build, the not-stored logs are the vast majority of the useless spam I want to remove. In master currently there are 2 lines of transient but 17 lines of non-stored reporting.

aj-fuentes: I'm having a look.

aj-fuentes: For example this is a mistake
https://github.com/odoo/upgrade/blob/57c5f86db8f87e9299ccb6583e0ab3d6a39b2a2d/migrations/hr/saas~19.2.1.1/pre-migrate.py#L30
The field is still there in 19.2
https://github.com/odoo/odoo/blob/d09170269c10c7a802bb6d1fc13d2336beb6ccf4/addons/hr/models/res_partner_bank.py#L12

aj-fuentes: Another mistake
https://github.com/odoo/enterprise/commit/b8f6b0ef47f4e4586c89927408164169d78b23f2#diff-1ceacb976fd08d6b9b48d7337dc15ecdac8114f9c273c098df0f12ca14f60ae9R17

Some fields were brought back but this wasn't reflected in the upgrade scripts

xmo-odoo: OK in that case want me to close this and you look into fixing those migrations / warnings if you consider them valuable?

And then maybe you can upgrade to a full warning if they're all either migration errors or forgotten migrations.

aj-fuentes: https://github.com/odoo/upgrade/pull/10612

xmo-odoo:
Closing because the noisy logs are legitimate and upgrade would rather fix the incorrect or missing migrations (odoo/upgrade#10612)

src/base/0.0.0/end-no-respawn-fields.py
aj-fuentes: What I propose is you make this a full warning. I'm fixing all the stuff I'm getting. In any case this PR won't be merged until all warnings are solved. I can work on that.

aj-fuentes:
New PR https://github.com/odoo/upgrade-util/pull/458 --it may take a while until we fix all issues...

PR #458, opened 2026-06-26
[IMP] base/0.0.0: too many re-spawned fields are mistakes

aj-fuentes:
We will block this at CI level.

### DIFF
--- a/src/base/0.0.0/end-no-respawn-fields.py
+++ b/src/base/0.0.0/end-no-respawn-fields.py
@@ -50,11 +50,11 @@ def migrate(cr, version):
         qualifier = "field" if store else "non-stored field"
         if transient:
             qualifier = "transient " + qualifier
-        lvl = util.NEARLYWARN if transient or not store else logging.CRITICAL
+        lvl = util.NEARLYWARN if transient else logging.ERROR if not store else logging.CRITICAL
         action = ""
 
         if "{}/{}".format(model, field) in ignored_fields_respawn:
             lvl = util.NEARLYWARN
             action = "; explicitly ignored"
 
-        _logger.log(lvl, "%s %s/%s has respawn%s.", qualifier, model, field, action)
+        _logger.log(lvl, "%s %s/%s has respawned%s.", qualifier, model, field, action)


src/base/0.0.0/end-no-respawn-fields.py
```diff
@@ -50,11 +50,11 @@ def migrate(cr, version):
         qualifier = "field" if store else "non-stored field"
         if transient:
             qualifier = "transient " + qualifier
-        lvl = util.NEARLYWARN if transient or not store else logging.CRITICAL
+        lvl = util.NEARLYWARN if transient else logging.WARNING if not store else logging.CRITICAL
```
KangOl: If they are really a mistake...
```suggestion
        lvl = util.NEARLYWARN if transient else logging.CRITICAL
```

aj-fuentes: I think CRITICAL would be odd for production logs :/
It isn't critical in the sense that there is no data loss[^1] we just need to be sure this blocks the CI.
Anyway I have no strong opinion, as long as we block the CI :)

[^1]: well, some filters and metadata would be changed but I still believe it isn't critical.

KangOl: We can then log them as `ERROR` to not fail the upgrade.

src/base/0.0.0/end-no-respawn-fields.py
```diff
@@ -50,11 +50,11 @@ def migrate(cr, version):
         qualifier = "field" if store else "non-stored field"
         if transient:
             qualifier = "transient " + qualifier
-        lvl = util.NEARLYWARN if transient or not store else logging.CRITICAL
+        lvl = util.NEARLYWARN if transient else logging.ERROR
```
KangOl: I was more thinking of
```suggestion
        lvl = util.NEARLYWARN if transient else logging.ERROR if not store else logging.CRITICAL
```

aj-fuentes: Done

aj-fuentes:
Finally green.

KangOl (approved):
@robodoo r+

PR #459, opened 2026-06-29
[IMP] util/pg: reduce query count

aj-fuentes:
The call to `column_info` for `ir_ui_view/arch_db` runs 56K times in an
upgrade 19->19.4 with a runbot enterprise DB.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -498,8 +498,20 @@ def pg_text2html(s, wrap="p"):
     )
 
 
+_JSONB_COLUMNS = {}
+if version_gte("17.0"):  # switched to jsonb in 16.0
+    _JSONB_COLUMNS[("ir_ui_view", "arch_db")] = True
+    _JSONB_COLUMNS[("ir_act_server", "name")] = True
+    _JSONB_COLUMNS[("ir_model", "name")] = True
+
+
 def get_value_or_en_translation(cr, table, column):
-    fmt = "{}->>'en_US'" if column_type(cr, table, column) == "jsonb" else "{}"
+    is_jsonb = (
+        _JSONB_COLUMNS[(table, column)]
+        if (table, column) in _JSONB_COLUMNS
+        else column_type(cr, table, column) == "jsonb"
+    )
+    fmt = "{}->>'en_US'" if is_jsonb else "{}"
     return format_query(cr, fmt, column)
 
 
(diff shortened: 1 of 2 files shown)


src/util/pg.py
```diff
@@ -498,8 +498,18 @@ def pg_text2html(s, wrap="p"):
...
+
+
 def get_value_or_en_translation(cr, table, column):
-    fmt = "{}->>'en_US'" if column_type(cr, table, column) == "jsonb" else "{}"
+    is_jsonb = (
+        _JSONB_COLUMNS[(table, column)]
+        if (table, column) in _JSONB_COLUMNS
+        else column_type(cr, table, column) == "jsonb"
+    )
+    fmt = "{}->>'en_US'" if is_jsonb else "{}"
```
Pirols, line 514:
```suggestion
    fmt = "{}->>'en_US'" if version_gte("17.0") or column_type(cr, table, column) == "jsonb") else "{}"
```

Smaller diff?

KangOl:

src/util/pg.py
KangOl: No. This function is not called only for the `arch_db` field.

aj-fuentes: > Smaller diff?

Not all columns are jsonb after 17 ;)

The point is that a column can be translated on and off. We know arch_db would hardly change.

Pirols: Oh, of course! 🤦

src/util/pg.py
```diff
@@ -498,8 +498,18 @@ def pg_text2html(s, wrap="p"):
     )
 
 
+_JSONB_COLUMNS = {}
+if version_gte("17.0"):  # switched to jsonb in 16.0
+    _JSONB_COLUMNS[("ir_ui_view", "arch_db")] = True
```
KangOl, line 503:
```suggestion
    _JSONB_COLUMNS[("ir_act_server", "name")] = True
    _JSONB_COLUMNS[("ir_model", "name")] = True
    _JSONB_COLUMNS[("ir_ui_view", "arch_db")] = True
```

aj-fuentes: Done

KangOl:
As I feared, the `_cache_domain_fields` patch causes more problems than it tries to solve: https://runbot.odoo.com/runbot/batch/2613901/build/115975836

The shortcut commits are OK.

aj-fuentes:
Dropped it.

KangOl (approved):
@robodoo rebase-ff r+

PR #460, opened 2026-07-03
[FIX] util.remove_group: handle rules linkde to only the removed group

KangOl:
The `ir.rule` model has a computed-stored `global` field that is set to True when they are not restricted to some group.

To avoid having rules becoming global, there is "ON DELETE RESTRICT" definied on the m2m table, forbidding to remove a group linked to a rule.

We already circumvent this 'ON DELETE RESTRICT' by explicitly removing the m2m entry (see the code below this patch hunk).

We however kept the linked rules. By chance, those rules had still the `global` flag set tot `False` even if there were no more group linked to them, forbidding them to be applied globally by the ORM.

Better explicitly removing them.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -709,6 +709,21 @@ def remove_group(cr, xml_id=None, group_id=None):
     # Get all fks from table res_groups
     fks = get_fk(cr, "res_groups", quote_ident=False)
 
+    if table_exists(cr, "rule_group_rel"):
+        # special case for the rules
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only mrp sale_management

aj-fuentes (approved):

rco-odoo (approved):

KangOl:
@robodoo r+

PR #461, opened 2026-07-07
[FIX] util/pg: model_name became non store

apan-odoo:
The field [`model_name`] has become `store=f` in `v18.1` and its column is [removed] from upgrade too.

Traceback:
```py
  File "/tmp/tmpd0m7olzw/migrations/util/fields.py", line 259, in remove_field
    adapt_domains(
  File "/tmp/tmpd0m7olzw/migrations/util/domains.py", line 479, in adapt_domains
    cr.execute(query, [target_model, match_old, dot_old])
  File "/home/odoo/src/odoo/19.0/odoo/sql_db.py", line 440, in execute
psycopg2.errors.UndefinedColumn: column "model_name" does not exist
LINE 2:                 SELECT id, (SELECT model_name FROM ir_act_se...
                                           ^
```

[`model_name`]: https://github.com/odoo/odoo/blob/68f258e99f42693131a5309b3606c3b95f93d824/odoo/addons/base/models/ir_actions.py#L638
[removed]: https://github.com/odoo/upgrade/blob/master/migrations/base/saas~18.1.1.3/pre-migrate.py#L6

### DIFF
--- a/src/util/domains.py
+++ b/src/util/domains.py
@@ -147,7 +147,7 @@ def _get_domain_fields(cr):
             )
         ]
 
-    if column_exists(cr, "base_automation", "action_server_id"):
+    if column_exists(cr, "base_automation", "action_server_id") and not version_gte("saas~18.1"):
         result += [
             DomainField(
                 "base_automation",


sagu-odoo:
i think this is  the same case as this was for https://github.com/odoo/upgrade-util/pull/366 not sure.

apan-odoo:
> i think this is the same case as this was for #366 not sure.

Maybe when `action_server_id (M2o)` [became](https://github.com/odoo/odoo/pull/114352/changes#diff-f456849127002cd01a9e2058290e40ae8e5c1d57e15785699f0172400766bcd4R69-R80) `action_server_ids (O2m)` the column was not dropped in upgrade. That's the reason the column still exists? For the dbs which have atleast moved to v17 before that patch

**Edit:** Additionally there are much failing dbs in past onto this.

aj-fuentes:
You should investigate why the column is there on the first place. Be in the lookout for new requests failing on this error.

src/util/domains.py
```diff
@@ -147,7 +147,7 @@ def _get_domain_fields(cr):
             )
         ]
 
-    if column_exists(cr, "base_automation", "action_server_id"):
+    if column_exists(cr, "base_automation", "action_server_id") and not version_gte("saas~18.1"):
```
aj-fuentes, line 150: Actually I think that you need to change the query when the column doesn't exist as in https://github.com/odoo/upgrade-util/pull/506/changes. 

The domain fields are still there, it's just that the way to get the model name is different. I would add a column check instead of a version check.

You can also short-circuit the column check in [`_COLUMNS`](https://github.com/odoo-dev/upgrade-util/blob/6d6fb3cf14836ab8eae5c35bfd4c7c8b91df7d40/src/util/pg.py#L548)

Finally, for the upgrade m2o->m2m, test it locally, if confirmed that the column was not removed then fix that (if you do a PR ensure it's linked here --same branch name)

apan-odoo: > Finally, for the upgrade m2o->m2m, test it locally, if confirmed that the column was not removed then fix that (if you do a PR ensure it's linked here --same branch name)

Although I already tried this locally, I believe it might be a bug in the ORM that has now been addressed. I tried digging into it as well, but had no luck. Maybe I’ll take another look :(

KangOl:
upgradeci retry with always only base_automation

PR #464, opened 2026-07-14
[FIX] util/records.py: add required visibilty in add_view

vksi-odoo:
The visibility field is now [required](https://github.com/odoo/odoo/commit/6fdb9a1a1215426a78a8824594db0c3c470b2d89)
 on the ir.ui.view model. Set the default
value to public in the add_view() helper when creating new views. https://github.com/odoo/upgrade-util/blob/743d75c99e13c26ec4a6e95619a5f6c24d3ee990/src/util/records.py#L310

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -2113,6 +2113,20 @@ class TestRecords(UnitTestCase):
         self.assertTrue(cat_2.exists())
         self.assertFalse(cat_3.exists())
 
+    def test_add_view(self):
+        cr = self.env.cr
+        arch = '<form><field name="name"/></form>'
+        view_id = util.add_view(cr, "test_add_view", "res.partner", "form", arch)
+        self.assertTrue(view_id)
+        cr.execute(
+            "SELECT name, model, type FROM ir_ui_view WHERE id = %s",
+            [view_id],
+        )
+        name, model, view_type = cr.fetchone()
+        self.assertEqual(name, "test_add_view")
+        self.assertEqual(model, "res.partner")
+        self.assertEqual(view_type, "form")
+
 
 class TestEditView(UnitTestCase):
     @parametrize(
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -38,6 +38,7 @@ from .inherit import direct_inherit_parents, for_each_inherit
 from .misc import AUTOMATIC, chunks, version_between, version_gte
 from .orm import env, flush
 from .pg import (
+    ColumnList,
     PGRegexp,
     SQLStr,
     _get_unique_indexes_with,
@@ -322,13 +323,24 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
     arch_col = "arch_db" if column_exists(cr, "ir_ui_view", "arch_db") else "arch"
     jsonb_column = column_type(cr, "ir_ui_view", arch_col) == "jsonb"
     arch_column_value = Json({"en_US": arch_db}) if jsonb_column else arch_db
+    extra = (
+        [arch_col]
+        + (["key"] if key_exist else [])
+        + (["visibility"] if column_exists(cr, "ir_ui_view", "visibility") and version_gte("saas~19.3") else [])
+    )
+    columns = ColumnList.from_unquoted(cr, extra)
+    params = ["%({})s".format(x) for x in extra]
     cr.execute(
-        """
-        INSERT INTO ir_ui_view(name, "type",  model, inherit_id, mode, active, priority, %s)
-        VALUES(%%(name)s, %%(view_type)s, %%(model)s, %%(inherit_id)s, %%(mode)s, 't', %%(priority)s, %%(arch_db)s %s)
-        RETURNING id
-    """
-        % (arch_col + (", key" if key_exist else ""), ", %(key)s" if key_exist else ""),
+        format_query(
+            cr,
+            """
+            INSERT INTO ir_ui_view(name, "type",  model, inherit_id, mode, active, priority{columns})
+            VALUES(%(name)s, %(view_type)s, %(model)s, %(inherit_id)s, %(mode)s, 't', %(priority)s{values})
+            RETURNING id
+            """,
+            columns=columns.using(leading_comma=True),
+            values=SQLStr(", " + ", ".join(params)),
+        ),
         {
             "name": name,
             "view_type": view_type,
@@ -337,7 +349,8 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
             "mode": "extension" if inherit_id else "primary",
             "priority": priority,
             "key": key,
-            "arch_db": arch_column_value,
+            "visibility": "public",
+            arch_col: arch_column_value,
         },
     )
     return cr.fetchone()[0]


sagu-odoo (commented):
Hello @vksi-odoo , Thanks for the fix!. Here are some points which should be handle. 

1. ``visiblity`` field belongs to website module and this util can be used if ``website`` module is not installed. 
2. selection valuation changed in ``19.3`` before that it was `''`

src/util/records.py
```diff
@@ -338,6 +338,7 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
             "priority": priority,
             "key": key,
             "arch_db": arch_column_value,
+            "visibility": "public",
```
aj-fuentes: Don't do this, just inline the value in the insert query.
To be done only if the column exists.

src/util/records.py
```diff
@@ -324,8 +324,8 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
     arch_column_value = Json({"en_US": arch_db}) if jsonb_column else arch_db
     cr.execute(
         """
-        INSERT INTO ir_ui_view(name, "type",  model, inherit_id, mode, active, priority, %s)
-        VALUES(%%(name)s, %%(view_type)s, %%(model)s, %%(inherit_id)s, %%(mode)s, 't', %%(priority)s, %%(arch_db)s %s)
+        INSERT INTO ir_ui_view(name, "type",  model, inherit_id, mode, active, priority, visibility, %s)
```
KangOl: This field doesn't exist in all versions (and not all databases, as it is added by the `website` module).

KangOl:
upgradeci retry with always only website

KangOl:
It also shows that this method is not covered by a test.

aj-fuentes:
@vksi-odoo please add a test in test_util for this method.

vksi-odoo:
Hello @aj-fuentes , @sagu-odoo 

I've updated the fix and added a test case for add_view().
Could you please take another look?

cc: @KangOl

src/base/tests/test_util.py
```diff
@@ -2113,6 +2113,38 @@ def test_delete_unused_include_m2m(self):
...
 
+    def test_add_view(self):
+        cr = self.env.cr
+        arch = """
+            <tree>
+                <field name="name"/>
+            </tree>
+        """
+        view_type = "list" if util.version_gte("saas~17.5") else "tree"
+        view_id = util.add_view(cr, "test_add_view", "res.partner", view_type, arch)
```
aj-fuentes: Just make it simpler
```suggestion
        arch = '<form><field name="name"/></form>'
        view_id = util.add_view(cr, "test_add_view", "res.partner", "form", arch)
```

src/base/tests/test_util.py
```diff
@@ -2113,6 +2113,38 @@ def test_delete_unused_include_m2m(self):
...
+        self.assertTrue(view_id)
+        if util.column_exists(cr, "ir_ui_view", "visibility"):
```
aj-fuentes: From this line down it becomes very specific. We could remove it. But it's still OK.

src/util/records.py
```diff
@@ -322,13 +322,23 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
...
+        arch_db_placeholder = "%(arch_db)s"
     cr.execute(
         """
         INSERT INTO ir_ui_view(name, "type",  model, inherit_id, mode, active, priority, %s)
-        VALUES(%%(name)s, %%(view_type)s, %%(model)s, %%(inherit_id)s, %%(mode)s, 't', %%(priority)s, %%(arch_db)s %s)
+        VALUES(%%(name)s, %%(view_type)s, %%(model)s, %%(inherit_id)s, %%(mode)s, 't', %%(priority)s, %s%s)
         RETURNING id
     """
-        % (arch_col + (", key" if key_exist else ""), ", %(key)s" if key_exist else ""),
+        % (
```
aj-fuentes: I'm not sure why the linter doesn't block this formatting, but we must use format_query. Since you are touching this code, better make it right.

aj-fuentes:
upgradeci retry with always base in all versions

KangOl:
upgradeci retry with always only website base in all versions

KangOl:
There is already the `key` column that is dynamic.
The extra columns can be accumulated and then use `ColumnList.from_unquoted` to inject then in the query.
```python
extra = [c for c in ["key", "visibility"] if column_exists(cr, "ir_ui_view", c)]
columns = ColumnList.from_unqoted(extra).using(leading_comma=True)
params = (", " + ", ".join("%({})s".format(c) for c in extra)) if extra else ""
```

vksi-odoo:
Hello @aj-fuentes @KangOl 

Thanks for the suggestion!

I've updated the implementation to use `ColumnList.from_unquoted()` for the dynamic columns as suggested.

I also included `arch_col` in the accumulated columns because the arch_db column is also dynamic.

src/util/records.py
```diff
@@ -322,13 +323,28 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
     arch_col = "arch_db" if column_exists(cr, "ir_ui_view", "arch_db") else "arch"
     jsonb_column = column_type(cr, "ir_ui_view", arch_col) == "jsonb"
     arch_column_value = Json({"en_US": arch_db}) if jsonb_column else arch_db
+    extra = (
+        ["arch_db"]
```
KangOl:
```suggestion
        [arch_col]
```

src/util/records.py
```diff
@@ -322,13 +323,28 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
...
+        + (["key"] if key_exist else [])
+        + (["visibility"] if column_exists(cr, "ir_ui_view", "visibility") and not version_gte("saas~19.3") else [])
+    )
+    columns = ColumnList.from_unquoted(cr, extra)
+    params = (
+        ["%(arch_db)s"]
+        + (["%(key)s"] if key_exist else [])
+        + (["''"] if column_exists(cr, "ir_ui_view", "visibility") and not version_gte("saas~19.3") else [])
+    )
```
KangOl: reuse `extra`
```suggestion
    params = ["%({})s".format(x) for x in extra]
```

src/util/records.py
```diff
@@ -322,13 +323,28 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
     arch_col = "arch_db" if column_exists(cr, "ir_ui_view", "arch_db") else "arch"
     jsonb_column = column_type(cr, "ir_ui_view", arch_col) == "jsonb"
     arch_column_value = Json({"en_US": arch_db}) if jsonb_column else arch_db
+    extra = (
+        ["arch_db"]
+        + (["key"] if key_exist else [])
+        + (["visibility"] if column_exists(cr, "ir_ui_view", "visibility") and not version_gte("saas~19.3") else [])
```
KangOl:
```suggestion
        + (["visibility"] if column_exists(cr, "ir_ui_view", "visibility") and version_gte("saas~19.3") else [])
```

src/util/records.py
```diff
@@ -322,13 +323,28 @@ def add_view(cr, name, model, view_type, arch_db, inherit_xml_id=None, priority=
...
+            INSERT INTO ir_ui_view(name, "type",  model, inherit_id, mode, active, priority{columns})
+            VALUES(%(name)s, %(view_type)s, %(model)s, %(inherit_id)s, %(mode)s, 't', %(priority)s{values})
+            RETURNING id
+            """,
+            columns=columns.using(leading_comma=True),
+            values=SQLStr(", " + ", ".join(params)),
+        ),
         {
             "name": name,
             "view_type": view_type,
```
KangOl, line 346:
```suggestion
            "view_type": view_type,
            "visibility": "public",
            arch_col: arch_column_value,
```

KangOl (changes requested):

vksi-odoo:
Hello @KangOl 
Thanks for the review
Suggested changes are done!!

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://pixelfed.social/p/tommyandtabitha/982991031995709369

PR #466, opened 2026-07-15
[FIX] util: fix migration report

sagu-odoo:
Skip generating the summary if no data:

This prevents summary from being added to the migration
report if there is nothing to add. If really wants to add
summary only use the "report_with_summary" method [example]

[example]: https://github.com/odoo/upgrade/blame/ed551c9fda560287ca3c39d47f152dce2e9564a0/migrations/base/0.0.0/pre-match-partner_company.py#L27

image

### DIFF
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -152,6 +152,9 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
     The entry consists of a category (title) and a summary (body).
     The entry displays a list of records previously returned by SQL query, or any list.
 
+    .. warning::
+            If the list to render would be empty, nothing is posted in the report.
+
     .. example::
 
         .. code-block:: python
@@ -170,8 +173,7 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
 
     :param str summary: description of a report entry.
     :param list(tuple) data: data to report, each entry would be a row in the report.
-                             It could be empty, in which case only the summary is rendered.
-    :param tuple(str) columns: columns in `data`, can be referenced in `row_format`.
+    :param tuple(str) columns: columns in `data`, can be referenced in `row_format`
     :param str row_format: format for rows, can use any name from `columns` or `links`, e.g.:
                            "Partner {partner_link} that lives in {city} works at company {company_link}."
     :param dict(str, tuple(str, str, str)) links: optional model/record links spec,
@@ -198,15 +200,14 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
             )
         return "<li>{}</li>".format(row_format.format(**row_dict))
 
-    if not data:
-        row_to_html(columns)  # Validate the format is correct, including links
-        return report_with_summary(summary=summary, details="", category=category)
-
     disclaimer = "The total number of affected records is {}. ".format(total) if total else ""
     total = len(data) if total is None else total
     limit = min(limit, total) if limit is not None else total
     if total > limit:
         disclaimer += "This list is showing the first {} records.".format(limit)
+    if not data[:limit]:
+        row_to_html(columns)  # Validate the format is correct, including links
+        return None  # there is nothing to show in the list
 
     rows = "<ul>\n" + "\n".join([row_to_html(row) for row in data[:limit]]) + "\n</ul>"
     return report_with_summary(summary, "<i>{}</i>{}".format(disclaimer, rows), category)
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -3068,19 +3068,8 @@ class TestAssertUpdated(UnitTestCase):
 
 
 class TestReportUtils(UnitTestCase):
-    def test_report_simple(self):
-        self.assertEqual(
-            util.report_with_list(
-                summary="Simple test with minimal arguments and no data.",
-                data=[],
-                columns=("id", "name"),
-                row_format="Partner {name} has id {id}.",
-            ),
-            "<summary>Simple test with minimal arguments and no data.</summary>",
-        )
-
     def test_report_links(self):
-        self.assertEqual(
+        self.assertFalse(
             util.report_with_list(
                 summary="Testing links.",
                 data=[],
@@ -3088,7 +3077,6 @@ class TestReportUtils(UnitTestCase):
                 row_format="Partner {partner_link}.",
                 links={"partner_link": ("res.partner", "id", "name")},
             ),
-            "<summary>Testing links.</summary>",
         )
 
     def test_report_data_minimal(self):


src/util/report.py
```diff
@@ -204,6 +204,9 @@ def row_to_html(row):
 
     disclaimer = "The total number of affected records is {}. ".format(total) if total else ""
     total = len(data) if total is None else total
+    if total is None:
```
Pirols:
```suggestion
    if not total:
```

This looks impossible: in the line right above if `total` was `None` it would be transformed into `len(data)` which can't be None. On the other hand it could be 0, if passed explicitely like in the example you posted. Which probably means we should change the condition above from `if not data` to `if not data or total == 0`

sagu-odoo: wait i think i mis read something. here if we don;t have data why we are forcing it report the summary. I think we should remove this
https://github.com/odoo/upgrade-util/blob/1ac1fbad13ddf18a21dc9e54bdc7e9249f318b3c/src/util/report.py#L201-L203

aj-fuentes: I think we can make this `return None`. It looks indeed weird to do anything in `report_with_list` if data is empty.

src/util/report.py
```diff
@@ -199,8 +199,7 @@ def row_to_html(row):
         return "<li>{}</li>".format(row_format.format(**row_dict))
 
     if not data:
-        row_to_html(columns)  # Validate the format is correct, including links
```
aj-fuentes, line 202: You should keep this line

Pirols: And make the condition `if not data or total == 0`.

Pirols: Or move this check to after `total = len(data) if total is None else total` and change it to `if not total`

aj-fuentes: Indeed, after that line
```py
    if not total:
        return  # there is nothing to show in the list
```

sagu-odoo: As i understand the objective,  introducing ``total`` to limit the ``list`` item according to data. So, i don't find any objective of doing as soon as we have ``data`` it fine we will get ``total``. So, moving the condition don't make sense too mee. 

Edit: i might be wrong but i just want to keep my point. If you still feel please do let me know. I will adapt it

KangOl:

src/util/report.py
Pirols: No, `total` - from the documentation - was supposed to be a way for the caller to relay the correct number of affected records; relevant for instance if `data` was partial/truncated (e.g. using `cr.fetchmany(n)`).

On the other hand, it's true that passing an explicit `limit` 0 (or `total` as pointed out already) would result in an empty list, just like with empty 'data`. All in all I'd propose the following patch:

```diff
diff --git i/src/util/report.py w/src/util/report.py
index e0a09fb..0817591 100644
--- i/src/util/report.py
+++ w/src/util/report.py
@@ -170,7 +170,6 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
 
     :param str summary: description of a report entry.
     :param list(tuple) data: data to report, each entry would be a row in the report.
-                             It could be empty, in which case only the summary is rendered.
     :param tuple(str) columns: columns in `data`, can be referenced in `row_format`.
     :param str row_format: format for rows, can use any name from `columns` or `links`, e.g.:
                            "Partner {partner_link} that lives in {city} works at company {company_link}."
@@ -198,13 +197,12 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
             )
         return "<li>{}</li>".format(row_format.format(**row_dict))
 
-    if not data:
-        row_to_html(columns)  # Validate the format is correct, including links
-        return report_with_summary(summary=summary, details="", category=category)
-
-    disclaimer = "The total number of affected records is {}. ".format(total) if total else ""
     total = len(data) if total is None else total
     limit = min(limit, total) if limit is not None else total
+    if not data or not limit:
+        return
+
+    disclaimer = "The total number of affected records is {}. ".format(total)
     if total > limit:
         disclaimer += "This list is showing the first {} records.".format(limit)
```

Notice this also changes the meaning of `total` to what I think was meant to be. @diagnoza can you shed some light here?

aj-fuentes: Please don't remove `row_to_html` in any version.

I think just returning when `data` is not truthy would be enough. That being said since the purpose of this util is to show a report **with a list** checking `total` right after we set its definitive value would be pretty clear.

src/util/report.py
```diff
@@ -200,7 +199,7 @@ def row_to_html(row):
 
     if not data:
         row_to_html(columns)  # Validate the format is correct, including links
-        return report_with_summary(summary=summary, details="", category=category)
+        return None  # there is nothing to show in the list
```
aj-fuentes: Maybe is data is set, but total was set to 0, for whatever reason.

I keep believing that it is better to return right before posting to be sure the logic is very close to the part that generates the list.
If you wish you can do `if not rows: return None  # there is nothing to show in the list` and it would be totally clear as well.

src/util/report.py
```diff
@@ -170,7 +170,6 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
 
     :param str summary: description of a report entry.
```
aj-fuentes, line 174: We should add a warning in the docstring before the example.
```
If the list to render would be empty, nothing is posted in the report
```

src/util/report.py
sagu-odoo: Using ``limit 0`` seems like a misuse. If data exists, does ``limit`` 0 really make sense?

aj-fuentes: Yes. It is valid.
When data is long we may want to limit to some max. Limit 0 may mean: we don't care.
One hypothetical example: `ODOO_UPG_MAX_REPORTED_CHAGES_IN_UNINPORTANT_MODEL=nothing`, could pass `limit=0` to disable the output for a particular DB.

aj-fuentes: To me, it's better to keep flexibility than tying up the parameters. The check should be in the very end, if `rows` is empty, for whatever reason (even extra reasons we may uncover later) we don't report.

sagu-odoo: yes i found this suggestion pretty good. look like test case also need to be adapted. Once i will be done will let you know.

src/util/report.py
```diff
@@ -198,17 +197,15 @@ def row_to_html(row):
             )
         return "<li>{}</li>".format(row_format.format(**row_dict))
 
-    if not data:
-        row_to_html(columns)  # Validate the format is correct, including links
```
aj-fuentes, line 202: This line is still being removed.

src/util/report.py
```diff
@@ -152,6 +152,9 @@ def report_with_list(summary, data, columns, row_format, links=None, total=None,
     The entry consists of a category (title) and a summary (body).
     The entry displays a list of records previously returned by SQL query, or any list.
 
+    .. warning::
+            If the list to render would be empty, nothing is posted in the report
```
aj-fuentes:
```suggestion
            If the list to render would be empty, nothing is posted in the report.
```

src/util/report.py
```diff
@@ -198,15 +200,14 @@ def row_to_html(row):
...
-    if not data:
-        row_to_html(columns)  # Validate the format is correct, including links
-        return report_with_summary(summary=summary, details="", category=category)
-
     disclaimer = "The total number of affected records is {}. ".format(total) if total else ""
     total = len(data) if total is None else total
     limit = min(limit, total) if limit is not None else total
     if total > limit:
         disclaimer += "This list is showing the first {} records.".format(limit)
+    if not data or not limit:
```
aj-fuentes:
```suggestion
    if not data[:limit]:
```

sagu-odoo:
thanks for review @aj-fuentes  and @Pirols  suggested changes done.

src/util/report.py
```diff
@@ -198,15 +200,14 @@ def row_to_html(row):
...
-        row_to_html(columns)  # Validate the format is correct, including links
-        return report_with_summary(summary=summary, details="", category=category)
-
     disclaimer = "The total number of affected records is {}. ".format(total) if total else ""
     total = len(data) if total is None else total
     limit = min(limit, total) if limit is not None else total
     if total > limit:
         disclaimer += "This list is showing the first {} records.".format(limit)
+    if not data[:limit]:
+        row_to_html(columns)
```
KangOl: Keep the comment.
```suggestion
        row_to_html(columns)  # Validate the format is correct, including links
```

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://ieji.de/@BlippyTheWonderSlug/116928959006185370

PR #467, opened 2026-07-21
[FIX] util/records: make xml_filname fully qualified

sagu-odoo:
Passing `xml_filename` as `module/file_path` does not work with
`get_resource_from_path()`, as it resolves the resource relative to
`addons.__path__`.

Since `addons.__path__` contains paths such as
`/home/odoo/src/enterprise/<version>` or `/home/odoo/src/odoo/<version>`,
the module name and data file must be resolved from the full module path
before looking up the resource.

See `odoo.modules.module.get_resource_from_path()`[1] for the corresponding
resource resolution logic.

[1]: https://github.com/odoo/odoo/blob/4b1077a5f8726f9139705521cc6e368fb5adba1a/odoo/modules/module.py#L347

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1866,6 +1866,21 @@ class TestRecords(UnitTestCase):
         cr.execute(query, [record_id])
         self.assertEqual(cr.fetchone()[0], original)
 
+    def test_update_record_from_xml_resets_arch_fs_noupdate(self):
+        cr = self.env.cr
+        xmlid = "base.contact_name"
+
+        view = self.env["ir.ui.view"].browse(util.ref(cr, xmlid))
+
+        original_arch_fs = view.arch_fs
+        view.arch_fs = False
+        util.flush(view)
+
+        util.update_record_from_xml(cr, xmlid)
+        util.invalidate(view)
+
+        self.assertEqual(original_arch_fs, view.arch_fs)
+
     def test_ensure_xmlid_match_record(self):
         cr = self.env.cr
         tx1 = self.env["res.currency"].create({"name": "TX1", "symbol": "TX1"})
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -67,6 +67,19 @@ try:
 except NameError:
     basestring = unicode = str
 
+if version_gte("15.0"):
+    from odoo.tools.misc import file_path
+elif version_gte("9.0"):
+    try:
+        from odoo.modules.module import get_resource_path
+    except ImportError:
+        from openerp.modules.module import get_resource_path
+
+    file_path = lambda path: get_resource_path(*path.split("/", 1))
+else:
+    # doesn't matter as `xml_filename` is not used in older version.
+    file_path = lambda path: path
+
 
 def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
     """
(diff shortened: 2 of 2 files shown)


sagu-odoo:
https://github.com/odoo/upgrade/pull/10823/files use case

sagu-odoo:
upgradeci retry with mail+sms

KangOl:
upgradeci retry with always only crm mail+sms

KangOl:
Does it work in all version? Is the `addons-path` always present? if not, we would have to use `odoo.tools.file_path`.

Also can you add a test to ensure we get the correct value?
The test would have to read the `arch_fs` of a view, reset it, then reread if after the call to `update_record_from_xml` to ensure it has been correctly reset.

sagu-odoo:
Hello @KangOl , 

I have checked regarding the `addons_path` it was added in [``15.2``](https://github.com/odoo/odoo/commit/2e29a9350306fd94fff19bc02821d7e7ab9c6f8b#diff-367842087e68d6811cf53f2ab3eb21103cd8e793b1648ef3e57f220ee0ccde4e)  and regarding the [``file_path``] it added in [14.3](https://github.com/odoo/odoo/commit/49429f986a98023aae282a9076597e55c61cb2e5). If we want to make it compatible with older version that's 8.0  in that case.
1. use ``file_open``  with ``module/name`` return file name can be used as path name
```py
 tools.file_open('crm/report/crm_lead_report_view.xml').name
u'/home/odoo/src/odoo/8.0/addons/crm/report/crm_lead_report_view.xml'
```
2. make new private method to do find a path for file as we don't have for older version
https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/tools/misc.py#L195-L202

KangOl:
For older versions, we can use `file_open(..., pathinfo=True)`

KangOl:
> For older versions, we can use `file_open(..., pathinfo=True)`

Actually, we can use `get_resource_path`, which avoid opening the file.
```python
if util.version_gte("15.0"):
    from odoo.tools.misc import file_path
elif util.verison_gte("9.0"):
    from odoo.modules.module import get_resource_path

    file_path = lambda path: get_resource_path(*path.split("/", 1))
else:
    # doesn't matter as `xml_filename` is not used in older version.
    file_path = lambda path: path
```

sagu-odoo:
upgradeci retry with always only crm mail+sms in all versions

sagu-odoo:
upgradeci retry with always only crm mail+sms in all versions

sagu-odoo:
@KangOl  thanks for suggestion 🐱

src/base/tests/test_util.py
```diff
@@ -1866,6 +1866,20 @@ def test_update_record_from_xml_cache(self):
...
         self.assertEqual(cr.fetchone()[0], original)
 
+    def test_update_record_from_xml_resets_arch_fs_noupdate(self):
+        cr = self.env.cr
+        xmlid = "base.contact_name"
+
+        view = self.env["ir.ui.view"].browse(util.ref(cr, xmlid))
+
+        original_arch_fs = view.arch_fs
+        view.arch_fs = False
```
KangOl, line 1876:
```suggestion
        view.arch_fs = False
        util.flush(view)
```

sagu-odoo: it's needed ?. we already do invalidate(that will eventually flush) just before parsing the target xml node. https://github.com/odoo/upgrade-util/blob/063f0afdf24ab53cbdb98165830315f150c822a3/src/util/records.py#L1275-L1280


Edit: nvm done

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://pixelfed.social/p/mam1hlap1natapa1/985804414041912772

PR #470, opened 2026-07-23
[FIX] util/report: make `details` optional

sagu-odoo:
`details` had no default, forcing every caller to pass it even when there was nothing to add in details. Default it to "" so callers can omit it.


KangOl:

KangOl:
What the point of using this function without `details`? Just use `report_message` directly, then. And no need to use the `<summary>` tag, a simple `<p>` is enough.

sagu-odoo:
indeed make sense

PR #472, opened 2026-07-24
[IMP] new model <-> table exception

KangOl:
See odoo/odoo#253064


KangOl:

KangOl:
@robodoo r+

KangOl:
@robodoo r+

PR #473, opened 2026-07-27
[IMP] util/report: Improve new reporting function

labo-odoo:
Description
Problem:-
Several small issues in the upgrade report helpers (report_with_summary / report_with_list) were producing incorrect or misleading output in the generated report UI.

1. ```<details>/<summary>``` nesting was inverted, causing the "Details" toggle to look odd
Before:
```
python"<summary>{}<details>{}</details></summary>".format(summary, details)
```
image
it will shown every version of field renamed
image
image

Per the HTML5 spec, ```<summary>``` only behaves as a clickable disclosure label when it is the first child of a ```<details> ``` element. Here, the outer ```<summary>``` had no ``` <details>``` parent, so it lost its toggle behavior entirely and rendered as plain, always-visible text. Meanwhile, the inner ```<details>``` had no ```<summary> ``` child of its own, so browsers auto-inserted their default fallback label — the literal word "Details" — as its toggle.

This is exactly why every "Fields renamed" entry showed a generic "▼ Details" toggle.

Fix: reverted to the correct, spec-compliant nesting — <details> as the parent, <summary> as its first child:
```
python"<details><summary>{}</summary>{}</details>".format(summary, details)
```
This restores the intended behavior already used consistently elsewhere in the migration reports: one clickable custom label, with the detailed content folded underneath it.

2. Disclaimer showed the configured limit instead of the actual number of rows displayed

When a caller fetches fewer rows than limit (e.g. data=cr.fetchmany(20) with total=cr.rowcount returning 1546, and the default limit=100), the old code compared total > limit and printed the disclaimer using limit:
```
"The total number of affected records is 1546. This list is showing the first 100 records."
```
image

This is inaccurate — only [20 records](https://github.com/odoo/upgrade/commit/2e47e5d37374a80f80c1611507e91f8d511c998d#diff-ecfa011cc2bfd6575bdf0ac7fd7466eb1ace63343ce83dd1a56d37e0de41e3a0R40) were actually rendered, not 100. The message described an internal config value instead of what was actually shown on screen the example is also given in the self report method [here](https://github.com/odoo/upgrade-util/blob/3663104f21e0dfde1bdbc47021335cd11d9f2826/src/util/report.py#L162-L172)

Fix: slice the data once (shown = data[:limit]), and base both the truncation check and the disclaimer text on the actual number of rows shown (shown_count = len(shown)) rather than the raw limit:
```
pythonif total > shown_count:
    disclaimer = "The total number of affected records is {}. This list is showing the first {} records.".format(total, shown_count)
```
Now the message always reflects reality, regardless of whether the caller under-fetched relative to limit.

3. Disclaimer appeared even for tiny lists where it added no value
image

Previously, the disclaimer's first sentence ("The total number of affected records is X.") was shown whenever total was truthy — even for a single record, where mentioning a "total" and "showing first N" is meaningless noise.

Fix: the disclaimer (both sentences, as one unit) is now only generated when the list was genuinely truncated, i.e. total > shown count. If everything fetched is being shown (1 record, 5 records, 20 out of 20, etc.), no disclaimer is added at all.

Improve:-  we can set the default limit 50 if we want to show minimal record. but showing for the single record disclaimer look not good.

With the current patch after migration

image
image
image

UPG:- 4467262

### DIFF
--- a/src/util/report.py
+++ b/src/util/report.py
@@ -138,9 +138,9 @@ def report_with_summary(summary, details, category="Other"):
     :param str category: Title of a report entry.
     """
     msg = (
-        "<summary>{}<details>{}</details></summary>".format(summary, details)
+        "<details><summary>{}</summary>{}</details>".format(summary, details)
         if details
-        else "<summary>{}</summary>".format(summary)
+        else "<p>{}</p>".format(summary)
     )
     report_message(message=msg, category=category, format="html")
     return msg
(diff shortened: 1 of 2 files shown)


KangOl:

src/util/report.py
```diff
@@ -138,7 +138,7 @@ def report_with_summary(summary, details, category="Other"):
     :param str category: Title of a report entry.
     """
     msg = (
-        "<summary>{}<details>{}</details></summary>".format(summary, details)
+        "<details><summary>{}</summary>{}</details>".format(summary, details)
         if details
         else "<summary>{}</summary>".format(summary)
```
KangOl: > A `<summary>` element may only be used as the first child of a `<details>` element

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/summary#usage_notes

So when no detail is provided, we can't just use `<summary>` tag.
```suggestion
        else "<p>{}</p>".format(summary)
```

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://piaille.fr/@CatEveryHour/117001735782690523

src/util/report.py
labo-odoo: @KangOl Sorry I haven't seen your comment thank you for the commit.

PR #474, opened 2026-07-29
[FIX] test_util: fixup cf2ab86

Pirols:
If a currency has already been used for accounting entries, trying to decrease its decimal places fails: https://upgradeci.odoo.com/upgradeci/run/331343.

The test on `pubuser.active` is a better check to ensure that `<record>` values are preserved anyways. So we can just remove the check on eur.rounding.


KangOl:
You can use the `symbol` column instead.

KangOl:
upgradeci retry with always only base account

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://piaille.fr/@CatEveryHour/117003623179995726

PR #475, opened 2026-07-30
[IMP] records: validate models/tables

Pirols:
Add basic `_validate_{model,table}` sanity checks.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -583,7 +583,7 @@ def _remove_import_export_paths(cr, model, field=None):
     else:
         export_q += " WHERE el.name IS NOT NULL"
 
-    impex_data = [(export_q, "ir_exports_line")]
+    impex_data = [(export_q, "ir.exports.line")]
     if table_exists(cr, "base_import_mapping"):
         import_q = """
             SELECT id,
@@ -595,7 +595,7 @@ def _remove_import_export_paths(cr, model, field=None):
             import_q = cr.mogrify(import_q + " WHERE field_name ~ %s ", [r"\y{}\y".format(field)]).decode()
         else:
             import_q += " WHERE field_name IS NOT NULL "
-        impex_data.append((import_q, "base_import_mapping"))
+        impex_data.append((import_q, "base_import.mapping"))
 
     for query, impex_model in impex_data:
         cr.execute(query)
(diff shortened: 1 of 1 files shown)


Pirols:
upgradeci retry with all modules in all versions

cc: @KangOl

src/util/records.py
```diff
@@ -516,6 +518,7 @@ def remove_records(cr, model, ids):
 
 
 def _rm_refs(cr, model, ids=None):
+    _validate_model(model)
```
KangOl, line 521: I'm not sure about using it in private functions. They are supposed to be called internally, from other functions where the parameters have already been validated.

KangOl:
```
;  rg -l 'util\.(add_view|remove_records|ensure_xmlid_match_record|replace_in_all_jsonb_values|ensure_mail_alias_mapping|update_parent_path)\(' | cut -d/ -f1 | sort | uniq | xargs echo 'upgradeci retry with always only in all versions'
upgradeci retry with always only in all versions account_asset account_reports ai analytic appointment auth_signup base data_merge documents documents_hr documents_hr_payroll documents_project event fleet helpdesk hr hr_fleet hr_payroll hr_payroll_account hr_recruitment iap im_livechat industry_fsm l10n_ae l10n_be_hr_payroll l10n_de l10n_dk l10n_eg l10n_es_edi_tbai l10n_fi_reports l10n_in_reports l10n_it l10n_jo l10n_lb_account l10n_nl l10n_si l10n_th l10n_vn mail mass_mailing mass_mailing_sms mrp payment point_of_sale portal_anonymous product project purchase sale sale_amazon sale_subscription sales_team stock stock_account training_base uom utm web web_studio website website_blog website_event website_sale website_slides
```

src/util/records.py
Pirols: I'm not sure either. However it seems that we have hardcoded errors even in the utils 2cf5657a6d8a82d89cbd91fac0bc54bffa5881df. In this case the models were passed on to a public function (`remove_records`), had it been a private one without validation it would have slipped through. Still I am not against applying this to public functions only if you prefer.

KangOl (approved):
@robodoo rebase-ff r+

[^source]

[^source]: https://mastodon.social/@pia_ranha/117010142306843321

PR #476, opened 2026-07-30
[FIX] util/pg: Add optional parameter

sagu-odoo:
add optional parameter too run parallel
execute serial wise if needed.

### DIFF
--- a/src/util/pg.py
+++ b/src/util/pg.py
@@ -153,7 +153,7 @@ else:
     _parallel_execute_threaded = _parallel_execute_serial
 
 
-def parallel_execute(cr, queries, logger=_logger, qualifier="queries"):
+def parallel_execute(cr, queries, logger=_logger, qualifier="queries", allow_parallel=AUTO):
     """
     Execute queries in parallel.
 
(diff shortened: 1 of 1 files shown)


sagu-odoo:
see https://github.com/odoo/upgrade/pull/10908/changes

src/util/pg.py
```diff
@@ -181,13 +181,19 @@ def parallel_execute(cr, queries, logger=_logger, qualifier="queries"):
...
-    )
-    return parallel_execute_impl(cr, queries, logger=_logger, qualifier=qualifier)
+
+    if allow_parallel is AUTO:
+        allow_parallel = not (
+            getattr(threading.current_thread(), "testing", False)
+            or (odoo_module is not None and getattr(odoo_module, "current_test", False))
+        )
+
+    if allow_parallel:
```
Pirols, line 190: I don't think you ever need a parameter for this. The `_serial` is sort of a fallback when we know we can't run in parallel.

EDIT: As discussed in DM it's really anti idiomatic to call `parallel_execute` while asking it **not** to run in parallel. On the caller side you can just write a for loop over exploded queries, and it is to be seen that's even better than a single bigger query.

KangOl:

src/util/pg.py
KangOl: OTOH, tests for this method should then be changed to explicitly call `_parallel_execute_threaded` as it's the one that is actually tested.

sagu-odoo:
https://github.com/odoo/upgrade/pull/10908/changes not needed. As the case was tooo specific

PR #477, opened 2026-07-30
[IMP] util/specific: add overwrite_arch

elct-odoo:
## util/specific.py: overwrite_arch

Adds `util.overwrite_arch(cr, filepath=None, xmlid=None, arch=None, view_id=None, active="auto")` to replace a view's arch from a file or raw string, avoiding the need to reimplement this pattern in individual migration scripts. I think this is a worthwhile addition to upgrade-util. It's a small utility that removes duplicated SQL and provides a reusable helper for a pattern that comes up regularly, especially for normalizing reports.

### Description

Reads the content of `filepath` (or accepts a raw `arch` string) and writes it directly to the view's arch column. Unlike `edit_view`, it does not parse the XML; it is intended for cases where the entire arch is being replaced from a maintained file.

### Usage

> **Note:** `MIG_SCRIPT_DIR` is not provided by this library. It is a common convention in migration scripts to reference files relative to the script itself. Define it at the top of your migration script so that file paths resolve correctly regardless of where the upgrade is run from.

```python
from pathlib import Path
MIG_SCRIPT_DIR = Path(__file__).parent
from odoo.upgrade import util

def migrate(cr, version):
       util.overwrite_arch(cr, MIG_SCRIPT_DIR / "arch/my_view.xml", "web_studio.studio_report")
```
```python
from odoo.upgrade import util
def migrate(cr, version):
       util.overwrite_arch(cr, xmlid="module.my_view_id", arch="<data>...</data>")
```
Hierarchy e.g.
```
web_studio/
└─ 19.0.1.0/
   - pre-migrate.py
   - arch/my_view.xml
   ```

### DIFF
--- a/src/util/specific.py
+++ b/src/util/specific.py
@@ -1,13 +1,24 @@
 # -*- coding: utf-8 -*-
 import logging
 
+from psycopg2.extras import Json
+
 from .exceptions import UpgradeError
 from .helpers import _validate_table
 from .misc import _cached, version_gte
 from .models import rename_model
 from .modules import rename_module
 from .orm import env
-from .pg import column_exists, format_query, parallel_execute, remove_constraint, rename_table, table_exists
+from .pg import (
+    column_exists,
+    column_type,
+    format_query,
+    parallel_execute,
+    remove_constraint,
+    rename_table,
+    table_exists,
+)
+from .records import ref
 from .report import add_to_migration_reports
 
 try:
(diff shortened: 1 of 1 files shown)


elct-odoo:
Moving this back to draft after getting some good feedback from DENI and SBK.

1. Implement `active="auto"` like `edit_view`.
2. Make `template_id` or `arch` optional parameters. This would let us overwrite a view's arch with a small XML snippet (e.g., one or two xpths) instead of requiring a full xml file for a simple change.

KangOl:
Move it to `specific.py`

KangOl:

KangOl:
Note that before being promoted to `upgrade-util`, such helper functions can be added to [local_util.py](https://github.com/odoo/upgrade-specific/blob/master/local_util.py)

elct-odoo:
> Moving this back to draft after getting some good feedback from DENI and SBK.
> 
> 1. Implement `active="auto"` like `edit_view`.
> 2. Make `template_id` or `arch` optional parameters. This would let us overwrite a view's arch with a small XML snippet (e.g., one or two xpths) instead of requiring a full xml file for a simple change.

Added both of these functionalities

elct-odoo:
@KangOl I made the requested changes, thanks!

PR #478, opened 2026-08-03
[FIX] util/fields: allow child model's field company dependent

hpr-odoo:
- it's raising an error for child model if `company` field is related to their parent model
- because of parent model field is there, but column is not exist
- to just overcome the error column does not exist for child model, added inherits model to have the vale from parent model's table
- [example](https://github.com/odoo/odoo/blob/6b53ca6d9e1d8bd926b5b64c894d1f5c2d25feb4/addons/product/models/product_product.py#L54)

- Error:
```py
File "/tmp/tmp0gpoh58w/migrations/base/saas~17.5.1.3/pre-company-dependent-jsonb.py", line 120, in update_db_column
    util.make_field_company_dependent(
  File "/tmp/tmp0gpoh58w/migrations/util/fields.py", line 917, in _convert_field_to_company_dependent
    alter_column_type(cr, table, field, "jsonb", using=using, where=where_condition)
  File "/tmp/tmp0gpoh58w/migrations/util/pg.py", line 846, in alter_column_type
    explode_execute(
  File "/tmp/tmp0gpoh58w/migrations/util/pg.py", line 429, in explode_execute
    return parallel_execute(
  File "/tmp/tmp0gpoh58w/migrations/util/pg.py", line 190, in parallel_execute
    return parallel_execute_impl(cr, queries, logger=_logger, qualifier=qualifier)
  File "/tmp/tmp0gpoh58w/migrations/util/pg.py", line 137, in _parallel_execute_threaded
    tot_cnt += future.result() or 0
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/tmp0gpoh58w/migrations/util/pg.py", line 115, in execute
    tcr.execute(query)
  File "/home/odoo/src/odoo/18.0/odoo/sql_db.py", line 371, in execute
psycopg2.errors.UndefinedColumn: column "company_id" does not exist
LINE 1: ...upg") WHERE "_standard_price_upg" IS NOT NULL AND "company_i...
```

- UPG-[4075838]

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -53,6 +53,7 @@ from .inherit import for_each_inherit
 from .misc import AUTO, log_progress, safe_eval, version_gte
 from .orm import env, invalidate
 from .pg import (
+    ColumnList,
     PGRegexp,
     SQLStr,
     alter_column_type,
@@ -872,12 +873,21 @@ m2o2m2m = convert_m2o_field_to_m2m
 
 
 def _convert_field_to_company_dependent(
-    cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
+    cr,
+    model,
+    field,
+    type,
+    target_model=None,
+    default_value=None,
+    default_value_ref=None,
+    company_field="company_id",
+    inherits=None,
 ):
     _validate_model(model)
     if target_model:
         _validate_model(target_model)
 
+    inherits = inherits or {}
     type2field = {"char", "float", "boolean", "integer", "text", "many2one", "date", "datetime", "selection", "html"}
     assert type in type2field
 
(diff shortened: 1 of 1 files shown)


KangOl:
upgradeci retry with always only product

KangOl:
We could simply check if the `company_id` field is not a related. If so, just follow it.

src/util/fields.py
```diff
@@ -872,12 +873,21 @@ def convert_m2o_field_to_m2m(cr, model, field, new_name=None, m2m_table=None, co
...
-    cr, model, field, type, target_model=None, default_value=None, default_value_ref=None, company_field="company_id"
+    cr,
+    model,
+    field,
+    type,
+    target_model=None,
+    default_value=None,
+    default_value_ref=None,
+    company_field="company_id",
+    inherits=None,
```
aj-fuentes, line 884: I agree with @KangOl 

Instead of passing inherits, just check `related` property in `ir_model_fields`. For example for `product.product.company_id` it is `product_tmpl_id.company_id`

PR #480, opened 2026-08-05
[FIX] base/0.0.0: compute field in SQL

labo-odoo:
Recomputing **commercial_partner_id** causes the ORM to load huge number of records raising an memory error when computing commercial_partner_id which is fetching is_company for every record Loading millions of records causes a MemoryError. [odoo fix
](https://github.com/odoo/odoo/pull/275295)
As till now the the compute was same in all stable version so we will only run the file once while in migration. may be in future if the computation was changes there we need to update the script accordingly.

```
  File "/home/odoo/src/odoo/15.0/odoo/modules/registry.py", line 88, in new
  File "/home/odoo/src/odoo/15.0/odoo/modules/loading.py", line 415, in load_modules
  File "/home/odoo/src/odoo/15.0/odoo/modules/loading.py", line 227, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/home/odoo/src/odoo/15.0/odoo/modules/migration.py", line 186, in migrate_module
  File "/tmp/tmptz8vmnt0/migrations/base/0.0.0/post-commercial_partner_id.py", line 10, in migrate
    util.recompute_fields(
  File "/tmp/tmptz8vmnt0/migrations/util/orm.py", line 267, in wrapper
    return f(*args, **kwargs)
  File "/tmp/tmptz8vmnt0/migrations/util/orm.py", line 345, in recompute_fields
    cr.commit()
  File "<decorator-gen-9>", line 2, in commit
  File "/home/odoo/src/odoo/15.0/odoo/sql_db.py", line 90, in check
  File "/home/odoo/src/odoo/15.0/odoo/sql_db.py", line 470, in commit
    self.flush()
  File "/home/odoo/src/odoo/15.0/odoo/sql_db.py", line 109, in flush
    self.transaction.flush()
  File "/home/odoo/src/odoo/15.0/odoo/api.py", line 846, in flush
    env_to_flush['base'].flush()
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 5692, in flush
    self.recompute()
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 6165, in recompute
    process(field)
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 6149, in process
    field.recompute(recs)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1265, in recompute
    self.compute_value(record)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1295, in compute_value
    records._compute_field_value(self)
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 4277, in _compute_field_value
    fields.determine(field.compute, self)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 88, in determine
    return needle(*args)
  File "/home/odoo/src/odoo/15.0/odoo/addons/base/models/res_partner.py", line 291, in _compute_display_name
    names = dict(self.with_context(**diff).name_get())
  File "/home/odoo/src/odoo/15.0/odoo/addons/base/models/res_partner.py", line 791, in name_get
    name = partner._get_name()
  File "/home/odoo/src/odoo/15.0/odoo/addons/base/models/res_partner.py", line 769, in _get_name
    name = self._get_contact_name(partner, name)
  File "/home/odoo/src/odoo/15.0/odoo/addons/base/models/res_partner.py", line 758, in _get_contact_name
    return "%s, %s" % (partner.commercial_company_name or partner.sudo().parent_id.name, name)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1080, in __get__
    self.recompute(record)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1273, in recompute
    self.compute_value(recs)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1295, in compute_value
    records._compute_field_value(self)
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 4277, in _compute_field_value
    fields.determine(field.compute, self)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 88, in determine
    return needle(*args)
  File "/home/odoo/src/odoo/15.0/odoo/addons/base/models/res_partner.py", line 351, in _compute_commercial_company_name
    p = partner.commercial_partner_id
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 2620, in __get__
    return super().__get__(records, owner)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1080, in __get__
    self.recompute(record)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1265, in recompute
    self.compute_value(record)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1295, in compute_value
    records._compute_field_value(self)
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 4277, in _compute_field_value
    fields.determine(field.compute, self)
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 88, in determine
    return needle(*args)
  File "/home/odoo/src/odoo/15.0/odoo/addons/base/models/res_partner.py", line 343, in _compute_commercial_partner
    if partner.is_company or not partner.parent_id:
  File "/home/odoo/src/odoo/15.0/odoo/fields.py", line 1109, in __get__
    recs._fetch_field(self)
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 3292, in _fetch_field
    self._read(fnames)
  File "/home/odoo/src/odoo/15.0/odoo/models.py", line 3389, in _read
    self.env.cache.update(fetched, field, values)
  File "/home/odoo/src/odoo/15.0/odoo/api.py", line 913, in update
    field_cache.update(zip(records._ids, values))
MemoryError
```

As the odoo generic will not merge in 15.0 as its not stable version :- https://github.com/odoo/odoo/pull/275295#issuecomment-5105265513
so we need to handle in migration only

Some commits which confirm:- https://github.com/odoo/odoo/pull/279976#issuecomment-5167372081
checkd the scirpt 

I have also test the soln with random record 

UPG:- 4467262

### DIFF
--- a/src/base/0.0.0/post-commercial_partner_id.py
+++ b/src/base/0.0.0/post-commercial_partner_id.py
@@ -1,15 +1,49 @@
-# -*- coding: utf-8 -*-
 from odoo.addons.base.maintenance.migrations import util
 
 
+@util.once("10.0", "20.0")
 def migrate(cr, version):
     # The `commercial_partner_id` field is expected to always be set. Although the column is not marked as `NOT NULL`.
     # Fight the Murphy's Law, and recompute the value on partners with a NULL value.
-    util.recompute_fields(
+    """Ref: https://github.com/odoo/odoo/blob/4cb3090ee829084c2a9260b6d38095cd818678f5/odoo/addons/base/models/res_partner.py#L515
+        if partner.is_company or not partner.parent_id:
+           partner.commercial_partner_id = partner
+        else:
+            partner.commercial_partner_id = partner.parent_id.commercial_partner_id
+    The `is_company` check is gone as of saas~19.1, see odoo/odoo@a0ee79e6."""
+    query = util.format_query(
         cr,
-        "res.partner",
-        ["commercial_partner_id"],
-        query="SELECT id FROM res_partner WHERE commercial_partner_id IS NULL",
-        chunk_size=10000,
-        strategy="commit",
+        """
+        WITH RECURSIVE _commercial_partners_to_set AS (
+              SELECT p.id,
+                     CASE WHEN p.parent_id IS NULL {0} THEN p.id
+                          ELSE p.parent_id
+                     END AS target_id,
+                     p.parent_id IS NULL {0} AS found
+                FROM res_partner p
+               WHERE p.commercial_partner_id IS NULL
+                 AND {{parallel_filter}}
+
+           UNION ALL
+
+              SELECT c.id,
+                     CASE WHEN p.commercial_partner_id IS NOT NULL
+                          THEN p.commercial_partner_id
+                          WHEN p.parent_id IS NULL {0} THEN p.id
+                          ELSE p.parent_id
+                     END AS target_id,
+                     p.commercial_partner_id IS NOT NULL OR p.parent_id IS NULL {0} AS found
+                FROM _commercial_partners_to_set c
+                JOIN res_partner p
+                  ON p.id = c.target_id
+               WHERE NOT c.found
+        )
+        UPDATE res_partner p
+           SET commercial_partner_id = c.target_id
+          FROM _commercial_partners_to_set c
+         WHERE c.id = p.id
+           AND c.found
+        """,
+        util.SQLStr("" if util.version_gte("saas~19.1") else "OR p.is_company IS TRUE"),
     )
+    util.explode_execute(cr, query, table="res_partner", alias="p")


src/base/0.0.0/post-commercial_partner_id.py
```diff
@@ -1,15 +1,39 @@
...
-        chunk_size=10000,
-        strategy="commit",
+        """
+        WITH RECURSIVE partner_commercial AS (
+            SELECT root.id,
+                   root.id AS commercial_partner_id,
+                   ARRAY[root.id] AS path
+              FROM res_partner root
+             WHERE (root.is_company IS TRUE OR root.parent_id IS NULL)
+               AND {parallel_filter}
```
hpr-odoo:
```suggestion
```
because if the child is in one range and its parent is outside that range, the recursion won't find the correct `commercial_partner_id`, leading to incorrect results.

And outside of CTE leads every worker builds the entire partner_commercial, even though it updates only a subset of rows.
E.x: If you have 8 workers, the recursive traversal is performed 8 times. For a large res_partner table, that's expensive.

hpr-odoo (commented):
also need to consider this comment: https://github.com/odoo/upgrade-util/pull/26#issuecomment-1857684281

KangOl:

src/base/0.0.0/post-commercial_partner_id.py
labo-odoo: Hello @hpr-odoo,

Thank you for the review. I think I understand your point now.

In the original recursive CTE approach, if one worker processes the range 1-10 and another processes 11-20, then a child in one range may not be able to reach its parent in another range, which can produce incorrect results.

Switching to a bottom-up approach would align much more closely with the ORM logic and would address that correctness issue. However, it still wouldn't solve the other concern you mentioned:
```
If you have 8 workers, the recursive traversal is performed 8 times. For a large res_partner table, that's expensive.
```

The key point of the old fix
- It is time-consuming on large databases (as shown in the PR).
- It can lead to MemoryError.

can we do instead?

```
util.explode_execute(
    cr,
    """
    UPDATE res_partner p
       SET commercial_partner_id = p.id
     WHERE p.commercial_partner_id IS NULL
       AND (p.is_company IS TRUE OR p.parent_id IS NULL)
       AND {parallel_filter}
    """,
    table="res_partner",
)


while util.explode_execute(
    cr,
    """
    UPDATE res_partner p
       SET commercial_partner_id = parent.commercial_partner_id
      FROM res_partner parent
     WHERE parent.id = p.parent_id
       AND p.commercial_partner_id IS NULL
       AND parent.commercial_partner_id IS NOT NULL
       AND {parallel_filter}
    """,
    table="res_partner",
):
    pass
```

For example, suppose there are around 1 million partners with a NULL commercial_partner_id.

- the first query processes those unresolved partners and assigns the value for companies and root partners.
- each subsequent iteration processes only the remaining unresolved partners, propagating the parent's already computed value one level deeper, following the same logic as the ORM compute.
- Since there is no recursive CTE, each worker performs only a simple UPDATE ... FROM instead of recursively traversing the hierarchy.

We can also decrease the iteration create a temporary index after the first update

- assume in the first query it resolves, around 500k partners, then only the remaining ~500k unresolved rows need to be considered in the another iterations. once the work done we can simply drop the temporary index.

I haven't tested this approach yet—it's just an idea open for optimised suggestion.

cc @KangOl

labo-odoo:
Hello @hpr-odoo 

I have tested and updated my soln in case if you need the flow of query in each iteration i have mention in detail please look in to the spefic fix of description :- https://github.com/odoo/upgrade-specific/pull/38964

cc @KangOl

src/base/0.0.0/post-commercial_partner_id.py
```diff
@@ -1,15 +1,38 @@
-# -*- coding: utf-8 -*-
 from odoo.addons.base.maintenance.migrations import util
 
 
+@util.once("10.0", "20.0")
```
KangOl, line 4: Why not after 20?
```suggestion
```

labo-odoo: @KangOl Since the logic is same till current version [19.4](http://github.com/odoo/odoo/blob/saas-19.4/odoo/addons/base/models/res_partner.py#L514) if in case the logic is change then it will configure wrong values
As i am not sure with this that's why i have put the 20.0.

src/base/0.0.0/post-commercial_partner_id.py
```diff
@@ -1,15 +1,38 @@
...
 def migrate(cr, version):
     # The `commercial_partner_id` field is expected to always be set. Although the column is not marked as `NOT NULL`.
     # Fight the Murphy's Law, and recompute the value on partners with a NULL value.
-    util.recompute_fields(
+    # Ref: https://github.com/odoo/odoo/blob/4cb3090ee829084c2a9260b6d38095cd818678f5/odoo/addons/base/models/res_partner.py#L515
+    # `if partner.is_company or not partner.parent_id: partner.commercial_partner_id = partner`
```
Pirols: This is no longer true in 19.1+. At least the `is_company` check is gone: https://github.com/odoo/odoo//blob/a0ee79e67a509685a5b362ae48c9ec9361b1d010/odoo/addons/base/models/res_partner.py#L503

src/base/0.0.0/post-commercial_partner_id.py
```diff
@@ -1,15 +1,38 @@
...
+    util.explode_execute(
```
Pirols:
```suggestion
    query = util.format_query(
        cr
        """
        WITH RECURSIVE _commercial_partners_to_set AS (
            SELECT p.id,
                   CASE WHEN p.parent_id IS NULL {} THEN p.id ELSE p.parent_id END AS target_id,
                   p.parent_id IS NULL {} AS found,
              FROM res_partner p
             WHERE p.commercial_partner_id IS NULL
               AND {{parallel_filter}}

         UNION ALL

            SELECT id,
                   CASE 
                       WHEN p.commercial_partner_id IS NOT NULL THEN p.commercial_partner_id 
                       WHEN p.parent_id IS NULL {} THEN p.id 
                       ELSE p.parent_id 
                   END AS target_id,
                   p.commercial_partner_id IS NOT NULL OR p.parent_id IS NULL {} AS found
              FROM _commercial_partners_to_set c
              JOIN res_partner p
                ON p.id = c.target_id
             WHERE NOT c.found
        )
        UPDATE res_partner p
           SET commercial_partner_id = c.target_id
          FROM _commercial_partners_to_set c
         WHERE c.id = p.id
           AND c.found
        """,
        util.SQLStr("OR p.is_company" if not util.version_gte("saas~19.1") else ""),
    )
    util.explode_execute(cr, query, table="res_partner", alias="p")
```

I think we can do this with a recursive cte.

Note this accounts for my other comment too (https://github.com/odoo/upgrade-util/pull/480/changes#r3749911622)

Pirols (commented):
I agree we should not limit this to 20.0

aj-fuentes:
From discussion in DM.

I'm not fan of this SQL code to be run in 0.0.0 _forever_. This code runs outside our control --public repo. Even though it looks OK because the code has been stable for many versions. I would feel more compelled if it is fully proved that we cannot keep up with `recompute_fields`. Why is it breaking now? In which case? Is the breaking case too specific? 

Alternatively, would adding ORDER BY to the query that fetches the list of IDs help? (For better chances of processing parents before children)

Maybe we can do the SQL version up to 20.0, and keep the `recompute_fields` version for higher (future) versions. We can always come back here and increase the max version for the SQL approach if we know nothing changed.

OTOH I'm not sure how impactful a slightly wrong computed commercial partner --due to some "future" misalignment-- could be. Probably not that much...

labo-odoo:
@aj-fuentes, Thank you for your revivew 
Is the breaking case too specific?
No — sagu hit the same class of issue before and tried to solve it in [SQL](https://github.com/odoo/upgrade-util/pull/26), which was closed because no more case

Why it breaks
Customer database has: 4449419 partners, 1252548 with a NULL commercial_partner_id.
recompute_fields  [invalidate](https://github.com/odoo/upgrade-util/blob/be32312146f198639ad6584fa7205933dff09781/src/util/orm.py#L348) after every chunk. A single chunk is enough, because of records.modified(fields)](https://github.com/odoo/upgrade-util/blob/be32312146f198639ad6584fa7205933dff09781/src/util/orm.py#L343): it collects the dependent records through _modified_triggers and marks them for recomputation, so the set being processed is no longer the chunk.

As i already face such type of case before with another recomputation  [ref](https://docs.google.com/presentation/d/11CcRk9MqlnaA67RbfHvyCqMH7VRzr06HU7SwjM9IiKw/edit?slide=id.g35a07119ce0_0_0#slide=id.g35a07119ce0_0_0)
The flow for the any recompute write will follow  in deeper like this. 
- Flow: write or update → called to modified →  and then traverse with traverse _modified_triggers → recompute the dependent records as part of process.
- The traversal walks out of the chunk through both direct dependencies (fields of the same model) and inverse ones (fields of related models reached through many2one/one2many):

commercial_partner_id → commercial_company_name → display_name → name_get() → _get_contact_name(), which reads name / company_name / is_company / parent_id of ancestors that are not in the chunk.

With this volume the cache fills with those dependent records and the MemoryError lands
same thing only mention in easy way [here](https://github.com/odoo/odoo/pull/275295) 

Would adding ORDER BY help?
- FYI if your pointing this then it already did. [recompute_fields](https://github.com/odoo/upgrade-util/blob/be32312146f198639ad6584fa7205933dff09781/src/util/orm.py#L312) fetches through [query_ids](https://github.com/odoo/upgrade-util/blob/be32312146f198639ad6584fa7205933dff09781/src/util/pg.py#L2052), which ends with '''SELECT id FROM {tmp} ORDER BY id''', and the docstring states the processing is always ascending.

Version bound
- Agreed, done. The SQL runs only within [10.0, 20.0] (util.once), and from saas~20.1 the original recompute_fields call is used unchanged — if the compute changes again we adapt the version rather than silently computing wrong values. 
- I've reworked and tested the @Pirols' suggestion — bottom-up, which is the direction the compute itself follows. I tested the patched result against the ORM logic and every record matches correctly. Proof below.

```
labo@upg-test-m2 ~/logs> psql labo_4449419_19.0 -Atc \
                                                         "SELECT id, commercial_partner_id FROM res_partner ORDER BY id" \
                                                         > /tmp/orm-commercial-partner_orm_labo.txt
labo@upg-test-m2 ~/logs> psql labo_4449419_copy_19.0 -Atc \
                                                     "SELECT id, commercial_partner_id FROM res_partner ORDER BY id" \
                                                     > /tmp/sql-commercial-partner_sql_labo.txt
labo@upg-test-m2 ~/logs> diff -u /tmp/orm-commercial-partner_orm_labo.txt /tmp/sql-commercial-partner_sql_labo.txt
labo@upg-test-m2 ~/logs> 
```

aj-fuentes:
In regard to what you posted about the triggering fields.
1. If the update of this field triggers the update of other stored fields it means that updating in SQL will be wrong (https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/odoo/addons/base/models/res_partner.py#L161)
2. Can we disable prefetch? Or make the chunks smaller?
3. Can we do the processing in a smarter way? Select first records without parent. Then records whose parents are in the previous set, then repeat (could be done with an extra column). A loop calling recompute_fields would work (we usually don't have a very deep structure).
4. Is the _specific_ DB you are dealing with just setting `commercial_partner_id = id` for most of the records? This case could be simpler. In the query you use to check before/after results, compare the other recomputed fields as well to see if the SQL is missing them.

The main reason why we have been relying on `recompute_fields` is precisely the triggers. If they cause a memory error we need to analyze first if we can simplify the processing. Otherwise whatever we do in SQL needs to update all triggered fields, which could be very hard to track across all versions :/

labo-odoo:
@aj-fuentes 
chunk_size only limits the records that recompute_fields() starts with. After that records.modified(fields) marks the dependent fields and then commit() → flush() → recompute() processes the full pending recomputation set.

At the commit point I can see these fields pending:
res.partner.commercial_partner_id --> res.partner.commercial_company_name ---> res.partner.display_name

Why smaller chunk_size does not solve it recompute_fields() basically does:
add_to_compute() -- recompute(records) -- records.modified(fields) -- commit()

chunk_size only limits the first recompute(records) part.
After modified() the ORM can add many more records because of the dependencies.

In this case:
```
commercial_partner_id
    depends on parent_id.commercial_partner_id

display_name
    depends on parent_id.display_name
```
So one partner can mark its children, then those children can mark their children, etc.
I tested with chunk_size=500 and saw: you can check the memory here
```
2026-08-12 17:35:14,290 23 INFO labo_4449419_copy_testing_19.0 odoo.upgrade.util.orm: Computing fields ['commercial_partner_id'] of 'res.partner' on 1252548 records 
> /tmp/tmpblz42q82/migrations/util/orm.py(344)recompute_fields()
-> recompute(records)
(Pdb) n
> /tmp/tmpblz42q82/migrations/util/orm.py(346)recompute_fields()
-> records.modified(fields)
(Pdb) import os, psutil
(Pdb) process = psutil.Process(os.getpid())
(Pdb) process.memory_info().rss / 1024 / 1024
130.82421875
(Pdb) n
> /tmp/tmpblz42q82/migrations/util/orm.py(347)recompute_fields()
-> import pdb; pdb.set_trace()
(Pdb) process.memory_info().rss / 1024 / 1024
395.4140625
(Pdb) n
> /tmp/tmpblz42q82/migrations/util/orm.py(348)recompute_fields()
-> if strategy == "commit":
(Pdb) n
> /tmp/tmpblz42q82/migrations/util/orm.py(349)recompute_fields()
-> import pdb; pdb.set_trace()
(Pdb) n
> /tmp/tmpblz42q82/migrations/util/orm.py(350)recompute_fields()
-> cr.commit()
(Pdb) n
MemoryError
> /tmp/tmpblz42q82/migrations/util/orm.py(350)recompute_fields()
-> cr.commit()
(Pdb) process.memory_info().rss / 1024 / 1024
2431.875
```
So the main problem seems to be during commit() → flush() → recompute(), not the initial chunk itself.

Disabling prefetch
I also tried that like this:
```
Partner = util.env(cr)["res.partner"].with_context(prefetch_fields=False)
util.recompute_fields(
    cr,
    Partner,
```
This can help with the _fetch_field() / prefetch part, but it does not solve the pending recomputation problem.

The ORM still gets the dependent fields from modified() and processes them during commit().

What I am trying now

With the SQL approach, we already calculate and store the correct commercial_partner_id directly in the database.
So we don't really need the ORM to calculate commercial_partner_id again.
What we still need is to make sure the ORM knows that the field was changed and that the required dependent fields are handled correctly.

The problem is that doing: `records.modified(["commercial_partner_id"])` and then: `cr.commit()`

creates the large pending recomputation set and we hit the MemoryError while updating the cache.

So I am trying to find if there is another ORM way to handle this part without going through the huge recomputation during commit(). may be not sure this [help](https://github.com/odoo/odoo/blob/12a66c6931d81e1cce18f676836aca0b32090d16/odoo/orm/environments.py#L435-L499) 
for the specific i tried a monkey patch based on the upstream 15.0 fix and it works completely perfect. [here](https://github.com/odoo/upgrade-specific/pull/38964/changes)

I am still trying to find the safest way to handle this in the upgrade
If yo have any suggestion or bettter way to handle the modified() / pending recomputation part when commercial_partner_id is already correctly stored by SQL or any other way.
cc @Pirols @KangOl

labo-odoo:
Hello @aj-fuentes,

Could you please review my last comment.

As a temporary workaround, i use a monkey patch? I have tested this approach, and it works correctly. It simply overrides the method and applies the Odoo fix of hpr.

Could you please review this specific fix, as the ticket has been open for quite long time.
Specific fix:- https://github.com/odoo/upgrade-specific/pull/38964/changes

aj-fuentes:
@labo-odoo I'd suggest we work with https://github.com/odoo/upgrade-util/pull/493 and see if we still face the issue.

labo-odoo:
@aj-fuentes The issue will be resolved with your fix
so clsoing current in favour of new fix  #493

PR #481, opened 2026-08-07
[IMP] util.remove_records

KangOl:
Delete the indirect references using `explode_execute`


KangOl:
upgradeci retry with always only base payment

aj-fuentes (approved):
DELETE has been safe for a while for parallel runs --defaults to serial execution on failure.

LGTM

KangOl:
@robodoo r+

PR #482, opened 2026-08-10
[IMP] remove_records

KangOl:
Using `explode_execute` for deleting indirect references is actually counter-productive when the list of references (`res_id` column) is small. We can just split per id as we can assume there will have an index on the res_model/res_id columns.


KangOl:

aj-fuentes (approved):
As discussed in DM, the original slow query had `ids` with multiple (many) elements.

KangOl:
@robodoo r+

PR #484, opened 2026-08-11
[IMP] util.pg.get_index_on

KangOl:
Add option to match indexes that use *at least* the given columns (in the given order).

It will allow to not create useless indexes of there is already one that match the given criteria.

--
Used by odoo/upgrade#10993


KangOl:
upgradeci retry with always only mail base

KangOl:
@robodoo r+

PR #486, opened 2026-08-13
[IMP] util.remove_views

mattismegevand:
Removing views one by one repeats t-call scans and reference cleanup.

Add remove_views to collect descendants and COW copies before removing t-calls and records in batch. Keep remove_view backward-compatible and use the batch helper when uninstalling or merging modules.

### DIFF
--- a/src/util/records.py
+++ b/src/util/records.py
@@ -99,111 +99,141 @@ def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
     .. warning::
        Either `xml_id` or `view_id` must be set. Specifying both will raise an error.
     """
-    assert bool(xml_id) ^ bool(view_id)
-    if xml_id:
-        view_id = ref(cr, xml_id)
-        if view_id:
-            module, _, name = xml_id.partition(".")
-            cr.execute("SELECT model FROM ir_model_data WHERE module=%s AND name=%s", [module, name])
+    remove_views(cr, xml_ids=[xml_id], view_ids=[view_id], silent=silent, keys={xml_id or view_id: key})
 
-            [model] = cr.fetchone()
-            if model != "ir.ui.view":
-                raise ValueError("%r should point to a 'ir.ui.view', not a %r" % (xml_id, model))
-    else:
-        # search matching xmlid for logging or renaming of custom views
-        xml_id = "?"
-        if not key:
-            cr.execute("SELECT module, name FROM ir_model_data WHERE model='ir.ui.view' AND res_id=%s", [view_id])
-            if cr.rowcount:
-                xml_id = "%s.%s" % cr.fetchone()
 
-    # From given or determined xml_id, the views duplicated in a multi-website
-    # context are to be found and removed.
-    if xml_id != "?" and column_exists(cr, "ir_ui_view", "key"):
-        cr.execute("SELECT id FROM ir_ui_view WHERE key = %s AND id != %s", [xml_id, view_id])
-        for [v_id] in cr.fetchall():
-            remove_view(cr, view_id=v_id, silent=silent, key=xml_id)
+def remove_views(cr, xml_ids=None, view_ids=None, silent=False, keys=None):
+    """
+    Remove several views and all their descendants in batch.
 
-    if not key and column_exists(cr, "ir_ui_view", "key"):
-        cr.execute("SELECT key FROM ir_ui_view WHERE id = %s and key != %s", [view_id, xml_id])
-        [key] = cr.fetchone() or [None]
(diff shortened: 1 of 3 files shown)


KangOl:
upgradeci retry with always only base website

src/util/records.py
```diff
@@ -99,6 +99,40 @@ def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
     .. warning::
        Either `xml_id` or `view_id` must be set. Specifying both will raise an error.
     """
+    to_remove = set()
+    tcalls = set()
+    _remove_view(cr, to_remove, tcalls, xml_id=xml_id, view_id=view_id, silent=silent, key=key)
+    _remove_redundant_tcalls(cr, tcalls)
+    remove_records(cr, "ir.ui.view", to_remove)
```
aj-fuentes: We don't need to duplicate code here.
```suggestion
    remove_views(cr, xml_ids=[xml_id], view_ids=[view_id], silent=silent, keys={xml_id or view_id: key})
```

src/util/records.py
```diff
@@ -99,6 +99,40 @@ def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
...
+
+
+def remove_views(cr, xml_ids=None, view_ids=None, silent=False):
```
aj-fuentes: `keys` could be a mapping `xmlid/id->key`. There is no (sane) way an XMLid could collide with a numeric ID so a unified mapping should be fine.
```suggestion
def remove_views(cr, xml_ids=None, view_ids=None, silent=False, keys=None):
    keys = keys or {}
```
Note that this allows to `batch-remove` in an upgrade script directly _and_ specify some keys for some views.

src/util/records.py
```diff
@@ -99,6 +99,40 @@ def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
...
+    .. warning::
+       Either `xml_ids` or `view_ids` must be set. Specifying both will raise an error.
+    """
+    assert (xml_ids is None) ^ (view_ids is None)
+    to_remove = set()
+    tcalls = set()
+    for xml_id in xml_ids or ():
+        _remove_view(cr, to_remove, tcalls, xml_id=xml_id, silent=silent)
+    for view_id in view_ids or ():
+        _remove_view(cr, to_remove, tcalls, view_id=view_id, silent=silent)
```
aj-fuentes:
```suggestion
        _remove_view(cr, to_remove, tcalls, xml_id=xml_id, silent=silent, keys.get(xml_id))
    for view_id in view_ids or ():
        _remove_view(cr, to_remove, tcalls, view_id=view_id, silent=silent, keys.get(view_id))
```

src/util/records.py
```diff
@@ -99,6 +99,40 @@ def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
...
+    _remove_redundant_tcalls(cr, tcalls)
+    remove_records(cr, "ir.ui.view", to_remove)
+
+
+def _remove_view(cr, to_remove, tcalls, xml_id=None, view_id=None, silent=False, key=None):
```
aj-fuentes: I'd prefer to not have an externally visible function with mutable "output" parameters.
I'd move this inside `remove_views` (right before the two loops) to make it clear this is a closure.
```suggestion
    to_remove, tcalls = set(), set()  # used by the recursive function below
    def _remove_view(xml_id=None, view_id=None, key=None):
```
`cr`, and `silent` are invariant for this inner function.

This diff would be also small with "Hide whitespace" option.

aj-fuentes (commented):
Some ideas geared towards my taste.

I'll let @KangOl give extra feedback here.

KangOl:

src/util/records.py
```diff
@@ -99,111 +99,141 @@ def remove_view(cr, xml_id=None, view_id=None, silent=False, key=None):
...
+                        "Disabled views",
+                    )
+        if not silent:
+            _logger.info("remove deprecated %s view %s (ID %s)", (key and "COWed") or "built-in", key or xml_id, view_id)
 
-    remove_records(cr, "ir.ui.view", [view_id])
+    for xml_id in xml_ids or ():
+        _remove_view(xml_id=xml_id, key=keys.get(xml_id))
+    for view_id in view_ids or ():
+        _remove_view(view_id=view_id, key=keys.get(view_id))
```
KangOl, line 234: Processing record one by one doesn't really improve performance.
Can't it be rewritten to process all the views all together?

mattismegevand: I think it could be doable I just wanted to not have a huge diff but I'll try do you have an idea in mind?

PR #487, opened 2026-08-13
[ADD] util.get_modules

KangOl:
Cached version of `odoo.modules.module.get_modules()`

As it is used a lot and don't change during the life of the odoo server, we can cache it result.

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -19,11 +19,6 @@ from psycopg2 import sql
 from psycopg2.extras import Json
 
 try:
-    from odoo import modules
-except ImportError:
-    from openerp import modules
-
-try:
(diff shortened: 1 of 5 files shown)


KangOl:

rco-odoo (approved):
Because it fixes the issue in all versions, I think that it is a good solution.

Xavier-Do:
I would be pleased to see this pr merged asap since it has the potential to speedup the upgrade on runbot significantly :)

aj-fuentes (approved):

aj-fuentes:
@robodoo r+

We still need to adapt the few scripts that do not use the cached (utils) version.

PR #488, opened 2026-08-13
[IMP] util.remove_view 

KangOl:
Limit search for redundant t-call on and from qweb views.


KangOl:

KangOl:
`util.remove_view` has been rewritten. Doesn't apply anymore.

PR #489, opened 2026-08-13
[FIX] util.misc._cached

KangOl:
Functions that take `cr` as only argument are cached per dbname.

They were previously gloablly cached.

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -12,6 +12,7 @@ import os
 import re
 import sys
 import textwrap
+import threading
 import uuid
 from contextlib import contextmanager
 from itertools import chain, islice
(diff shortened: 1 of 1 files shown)


KangOl:

src/util/misc.py
```diff
@@ -75,17 +78,48 @@ def __repr__(self):
...
+        sig_args, sig_kwargs = sig[0], sig[1:]
+        if not (sig_args in ([], ["cr"]) and sig_kwargs == (None, None, None)):
+            raise TypeError("Invalid function signature; expected `()` or `(cr)`.")
 
-    return wrapper
+        lru = _LRU(16 if sig_args else 1)
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            key = args[0].dbname if args else ()
```
aj-fuentes: Calling `func(cr=cr)` would invalidate this. I think this is the advantae functools variant below has.
```suggestion
            key = args[0].dbname if args else kwargs.get("cr", ())
```

PR #491, opened 2026-08-13
[FIX] util/misc: allow None in util.once

diagnoza:
`None` as source or target version in `util.once` is a legitimate value that corresponds to `util.version_gte(..)`.
The [previous block](https://github.com/odoo/upgrade-util/blob/cb9ef521af323981f2363b94faa2679c2a639b6a/src/util/misc.py#L200-L201) already treats it as such.

Previous implementation failed the assertion: for instance an upgrade from 16.0 to 19.0, running `util.once('saas~19.4', None)` at the `17.0` step of the upgrade would mean `a=saas~19.4` and `b=17.0`, thus failing `assert a <= b`.

Required for https://github.com/odoo/upgrade/pull/11027, as well as a fix for https://github.com/odoo/upgrade/blob/cdfe501f2fe8b49edcd1a7e9377eb55801a62e1e/migrations/accountant_knowledge/0.0.0/end-article-knowledge.py#L10.

### DIFF
--- a/src/util/misc.py
+++ b/src/util/misc.py
@@ -208,9 +208,10 @@ class once(object):
         s, t = parse_version(_SOURCE_VERSION), parse_version(_TARGET_VERSION)
         assert s <= t, "check source and target env vars"
 
+        if lower is not None and upper is not None and parse_version(lower) > parse_version(upper):
+            raise SleepyDeveloperError("invalid interval: lower={!r} > upper={!r}".format(lower, upper))
         a = s if lower is None else parse_version(lower)
         b = t if upper is None else parse_version(upper)
-        assert a <= b, "invalid interval"
 
         s_maj = int(_SOURCE_VERSION.replace("saas~", "").replace("saas-", "").split(".", 1)[0])
         t_parts = _TARGET_VERSION.replace("saas~", "").replace("saas-", "").split(".", 1)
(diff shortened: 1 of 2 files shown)


KangOl:
upgradeci retry with always only hr_skills l10n_tr_nilvera_einvoice account_reports account_reports hr_expense sale_subscription accountant_knowledge point_of_sale stock base

src/util/misc.py
```diff
@@ -208,9 +208,10 @@ def __init__(self, lower, upper, logger=None):
         s, t = parse_version(_SOURCE_VERSION), parse_version(_TARGET_VERSION)
         assert s <= t, "check source and target env vars"
 
+        if lower is not None and upper is not None and parse_version(lower) > parse_version(upper):
```
aj-fuentes, line 211: I'd say that if `lower == upper` we can wake up the dev. In such case we just need a single script in the relevant version.
```suggestion
        if lower is not None and upper is not None and parse_version(lower) >= parse_version(upper):
```

aj-fuentes (commented):
Could you please add a case in the test that covers the failure you encountered?

src/util/misc.py
aj-fuentes: Well, for some reason we allow this even in tests. Like
https://github.com/odoo/upgrade-util/blob/31340eb0e305350ae9388f274cca00af1cde56c8/src/base/tests/test_util.py#L2606

So I guess it's `>` in the end. My bad.

aj-fuentes (approved):
Once fixed the > vs >= mismatch we can merge it. I'd say we keep `>` since it was already allowed (avoid breaking code relying on it).

aj-fuentes (approved):
@robodoo r+

PR #492, opened 2026-08-14
[FIX] util/pg: fix obtaining m2m

diagnoza:
Fixes the following issues with the query returning m2m:
- Duplicate constraints responsible for returning the same m2m more than once.
Now we first gather fk constraints in a CTE that ensures there are no dupes, and only then we do the two `JOIN`s on `pg_attribute` to obtain their columns.
- Typo in the a2 `JOIN` referencing `c1.conkey`, instead of `c2.conkey`.
Made irrelevant after refactoring. Now there is only one place where we check `AND array_length(c.conkey, 1) = 1`.
- Add `AND NOT a.attisdropped` to exclude dropped columns.
Dropped columns remain in `pg_attribute` flagged as `attisdropped=True`, leaving room for erroneously fetching them when `JOIN`ining `pg_attribute`.

### DIFF
--- a/src/base/tests/test_util.py
+++ b/src/base/tests/test_util.py
@@ -1258,6 +1258,50 @@ class TestPG(UnitTestCase):
         self.assertEqual("res_groups_res_users_rel", auto_generated_m2m_table_name)
         self.assertTrue(util.table_exists(cr, auto_generated_m2m_table_name))
 
+    def test_get_m2m_on(self):
+        cr = self.env.cr
+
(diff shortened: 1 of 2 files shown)


KangOl:

KangOl:
Could you add a test to show what it fix?

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://hachyderm.io/@ggggbbybby/117109643861577866

PR #494, opened 2026-08-19
A little bit faster

KangOl:


KangOl:

src/util/misc.py
```diff
@@ -98,6 +98,7 @@ def str2bool(s, default=None):
     return s in y
 
 
+@getattr(functools, "lru_cache", lambda f: f)
```
aj-fuentes: This fails for [Python <3.8](https://docs.python.org/3.8/library/functools.html#functools.lru_cache:~:text=Changed%20in%20version%203%2E8%3A%20Added%20the%20user%5Ffunction%20option)

```suggestion
# lru_cache before Python3.8 generates a decorator, and cannot be used directly as such
@getattr(functools, "lru_cache", lambda: lambda f: f)()
```
```
  File "/tmp/tmpkuuxebkp/odoo/12.0/odoo/addons/base/maintenance/migrations/util/misc.py", line 101, in <module>
    @getattr(functools, "lru_cache", lambda f: f)
  File "/home/upgradeci/.pyenv/versions/3.6/lib/python3.6/functools.py", line 477, in lru_cache
    raise TypeError('Expected maxsize to be an integer or None')
TypeError: Expected maxsize to be an integer or None
```

src/util/misc.py
```diff
@@ -114,6 +115,7 @@ def version_gte(version):
     return parse_version(release.serie) >= parse_version(version)
 
 
+@getattr(functools, "lru_cache", lambda f: f)
```
aj-fuentes: same

src/util/misc.py
KangOl: I saw that in the UpgradeCI error.
Already fixed.

aj-fuentes (approved):
Looks good now.
Let's wait if there are no more errors in CI.

KangOl:
@robodoo rebase-ff r+

PR #495, opened 2026-08-19
[FIX] util/fields: skip stale related path when removing field

apan-odoo:
`remove_field` blindly reused a field's `related` path to redirect domain adaptation, even if the path's target had already been removed by an earlier `remove_field` call (e.g. removing a related field before its target across two models). This produced domains referencing non-existing fields, breaking at runtime.

Validate the path with `resolve_model_fields_path` before using it; fall back to plain domain cleanup if it no longer resolves.

usecase: https://github.com/odoo/upgrade/pull/11066

### DIFF
--- a/src/util/fields.py
+++ b/src/util/fields.py
@@ -246,6 +246,10 @@ def remove_field(
             cr.execute("SELECT related FROM ir_model_fields WHERE model=%s AND name=%s", [model, fieldname])
             if cr.rowcount:
                 related = cr.fetchone()[0]
+            if related:  # `related` could be using stale references to removed/invalid fields
+                path = related.split(".")
+                if len(resolve_model_fields_path(cr, model, path)) != len(path):
+                    related = None
 
         if related:
             update_field_usage(cr, model, fieldname, related, skip_inherit=skip_inherit)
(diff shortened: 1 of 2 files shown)


src/util/fields.py
```diff
@@ -246,6 +251,10 @@ def remove_field(
             cr.execute("SELECT related FROM ir_model_fields WHERE model=%s AND name=%s", [model, fieldname])
             if cr.rowcount:
                 related = cr.fetchone()[0]
+            if related:
```
```suggestion
            if related:   # `related` could be using stale references to removed/invalid fields
```

KangOl:
upgradeci retry with always only base planning_field_service

aj-fuentes (commented):
Looks good by why don't you include a test to be sure it works for both bad and good field paths.

apan-odoo:
> Looks good by why don't you include a test to be sure it works for both bad and good field paths.

Added the test for both cases, valid related path rewrites the domain, stale one falls back to `TRUE_LEAF`. 
Thanks for the review!

aj-fuentes (approved):

aj-fuentes:
@robodoo r+

PR #497, opened 2026-08-24
[FIX] util.remove_field

KangOl:
Avoid adapting related and models multiple time.
The processing of the inherited models will be done by the recursive call.


KangOl:

aj-fuentes (approved):
Looks obvious in retrospect.

Let's see how the CI ends.

aj-fuentes:
@robodoo r+

PR #498, opened 2026-08-24
[FIX] get_m2m_on: include self m2m and speedup

Pirols:


src/util/pg.py
KangOl, line 1836: Skipping the m2m on itself are indeed a bug. But not returning twice is actually dependent on what the caller will do with the result. Even if all our current usages expect unique tables, I would let the caller decide the behavior
```suggestion
def get_m2m_on(cr, table, dedup=True):
```

(variable name is subject of debate).

src/util/pg.py
```diff
@@ -1856,7 +1856,7 @@ def get_m2m_on(cr, table):
                AND array_length(c.conkey, 1) = 1
           GROUP BY c.conrelid, c.conkey[1], c.confrelid
         )
-        SELECT t.relname, a1.attname, a2.attname, other_table.relname
+        SELECT t.relname, MIN(a1.attname), MIN(a2.attname), other_table.relname
```
KangOl: This is wrong. It will return the same attribute twice.

Example with `res_groups`
```
res_groups_implied_rel                           | gid           | gid                                  | res_groups
```

```suggestion
        SELECT t.relname, (array_agg(a1.attname ORDER BY a1.attnum))[1], (array_agg(a2.attname ORDER BY a2.attnum))[2], other_table.relname
```

KangOl (changes requested):

src/util/pg.py
KangOl: Ignore the suggestion. It's still wrong.

Pirols: I fixed it with `MIN` and `MAX`.

src/util/pg.py
```diff
@@ -1835,53 +1835,34 @@ def get_m2m_on(cr, table):
...
+            ON a2.attrelid = rel.oid
+           AND a2.attnum = f2.conkey[1]
+         WHERE rel.relkind = 'r'
+           AND (
+                   SELECT COUNT(*)
+                     FROM pg_attribute a
+                    WHERE a.attrelid = rel.oid
+                      AND NOT a.attisdropped
+                      AND a.attnum > 0
+               ) = 2
```
KangOl, line 1882:
```suggestion
           AND rel.relnatts = 2
```

Pirols: Nice! Unfortunately it does not result in further speedups. Though the plan is also reduced.

KangOl:
upgradeci retry with always only hr crm_iap_mine

KangOl:
Note that on bigger databases, the gain is way more than 5x.

src/util/pg.py
KangOl: The fix should also be included in the first commit.

src/util/pg.py
```diff
@@ -1834,55 +1834,35 @@ def get_m2m_on(cr, table):
...
-           AND other_table.relkind = 'r'
-    """
+            ON a2.attrelid = rel.oid
+           AND a2.attnum = f2.conkey[1]
+         WHERE rel.relkind = 'r'
+           AND rel.relnatts = 2
+    {group_by}
+        """,
+        attribute_names=SQLStr("MIN(a1.attname), MAX(a2.attname)" if dedup else "a1.attname, a2.attname"),
+        group_by=SQLStr("GROUP BY 1, 4" if dedup else ""),
```
```suggestion
        group_by=SQLStr("GROUP BY rel.relname, other_table.relname" if dedup else ""),
```

KangOl (changes requested):

src/util/pg.py
KangOl: It actually doesn't work for dropped columns. :/

src/util/pg.py
```diff
@@ -1834,55 +1834,35 @@ def get_m2m_on(cr, table):
...
+        """
+        SELECT rel.relname, {attribute_names}, other_table.relname
+          FROM pg_class rel
+          JOIN pg_constraint f1
+            ON f1.conrelid = rel.oid
+           AND f1.confrelid = to_regclass(%s)
+           AND ARRAY_LENGTH(f1.conkey, 1) = 1
+          JOIN pg_constraint f2
+            ON f2.conrelid = rel.oid
+           AND f2.oid != f1.oid
```
```suggestion
           AND f2.oid != f1.oid
           -- Avoid matching a duplicated FK. The second FK should target another column
           AND f2.conkey != f1.conkey
```

src/util/pg.py
```diff
@@ -1834,55 +1834,35 @@ def get_m2m_on(cr, table):
...
-            ON f2.confrelid = other_table.oid
-         WHERE the_table.relkind = 'r'
-           AND the_table.relname = %s
-           AND other_table.relname != the_table.relname
-           AND other_table.relkind = 'r'
-    """
+            ON a2.attrelid = rel.oid
+           AND a2.attnum = f2.conkey[1]
+         WHERE rel.relkind = 'r'
+           AND rel.relnatts = 2
```
```suggestion
           AND (
               -- we cannot use `rel.relnatts` as dropped columns are counted in
               SELECT COUNT(*)
                     FROM pg_attribute a
                    WHERE a.attrelid = rel.oid
                      AND NOT a.attisdropped
                      AND a.attnum > 0
               ) = 2
```

src/util/pg.py
```diff
@@ -1824,7 +1824,7 @@ def get_m2m_tables(cr, table):
     return [r[0] for r in get_m2m_on(cr, table)]
 
 
-def get_m2m_on(cr, table):
+def get_m2m_on(cr, table, dedup=True):
```
KangOl: Actually, we need the `GROUP BY` to handle the duplicated FKs.

So we can't (easily) provide a sane `dedup` option to return m2m on the same table twice.

This can be removed and documented that "m2m with the two columns pointing to the same table are only returned once".
```suggestion
def get_m2m_on(cr, table):
```

KangOl (changes requested):
Can you also amend the test to create a m2m with the two columns pointing to `_upg_test_m2m_main`?

src/util/pg.py
Pirols: Yes, I will fixup the first commit once we have all details sorted.

KangOl (approved):
@robodoo rebase-ff r+

[^source]

[^source]: https://mastodon.social/@chopps/117161134215478206

aj-fuentes:
Nice

PR #505, opened 2026-08-31
[IMP] */saas~19.5: remove Field.size from python

kmagusiak:
The `Field.size` attribute is deprecated for years. Use it only for UI as this silently trucates values.

odoo/odoo#284290


KangOl:

src/base/tests/test_util.py
```diff
@@ -848,7 +848,7 @@ def test__normalize_pg_type(self, type_, expected):
         [
             ("res_country", "name", False, "jsonb" if util.version_gte("16.0") else "varchar"),  # translated field
             ("res_country", "code", False, "varchar"),
-            ("res_country", "code", True, "varchar(2)"),
+            ("res_country", "code", True, "varchar"),
```
KangOl: No. You broke the test for older versions.

PR #509, opened 2026-09-03
[FIX] util/models: wrap removed fields list in a details tag

labo-odoo:
- Wrapped the list of removed fields inside a `<details>` block.
- Aligns the message structure with standard reporting formats.
(worst case) when a db has a large number of remove fields

UPG-4632046

Current behaviour:
image

After Migration:
image

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -37,7 +37,7 @@ from .pg import (
 # avoid namespace clash
 from .pg import rename_table as pg_rename_table
 from .records import _remove_import_export_paths, _rm_refs, remove_records, remove_view, replace_record_references_batch
-from .report import add_to_migration_reports
+from .report import add_to_migration_reports, html_escape, report_with_list
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


aj-fuentes (approved):

src/util/models.py
```diff
@@ -211,13 +211,14 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
                         remove_field(cr, field_model, field, drop_column=False)
-                    msg = "The following fields have been removed because their related model {} ({}) was removed:\n{}".format(
-                        mod_label, model, "\n".join(" - field {} of model {}".format(*r) for r in affected_fields)
-                    )
-                    add_to_migration_reports(
-                        message=msg,
+                    report_with_list(
+                        summary="The following fields have been removed because their related model {} ({}) was removed.".format(
+                            mod_label, model
+                        ),
```
aj-fuentes: Actually this needs to be sanitized (import escape_html from utils)
```suggestion
                    msg = "The following fields have been removed because their related model {} (<pre>{}</pre>) was removed.".format(
                            html_escape(mod_label), html_escape(model)
                        ),
                    report_with_list(
                        summary=msg,
```

src/util/models.py
```diff
@@ -211,13 +211,14 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+                        data=affected_fields,
+                        columns=("name", "model"),
+                        row_format="field <kbd>{name}</kbd> of model <kbd>{model}</kbd>",
```
aj-fuentes: Does `pre` make this less "highlighted"?
```suggestion
                        row_format="field <pre>{name}</pre> of model <pre>{model}</pre>",
```

aj-fuentes (changes requested):

src/util/models.py
labo-odoo: @aj-fuentes with your suggested changes it comes like this while using ```<pre>```

image

if not ```<kbd>``` we can use the ```<code>``` it will higlight in red colour like this 
image

let me know in which way should i go ?

aj-fuentes: `code` is better indeed.

KangOl:

src/util/models.py
labo-odoo: @aj-fuentes 
Updated the patch with the ```<code>``` tag and for the model name instead of the ```<pre>``` i just use bold it looks good better than last one let me know if still need to do anything.

it looks like this with current patch
image

aj-fuentes (approved):

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://mastodon.social/@brainpilgrim/117207769060970824

PR #512, opened 2026-09-09
[FIX] util/models: convert outgoing restrict FKs to SET NULL

apan-odoo:
`remove_model()` can crash with a FK violation if the model being removed has its own restrict foreign key to something that gets deleted along the way.

**Example:**
- model A has a field `model_id` -> `ir.model`, `ondelete='cascade'`
- model B has a field `a_id` -> model A, `ondelete='restrict'`
- a record of A exists, pointing at model B (via `model_id`)
- a record of B exists, pointing at that same record of A (via `a_id`)

Now call `remove_model(cr, "B")`. To remove model B, it deletes the `ir.model` row for B. Since `A.model_id` is `ondelete='cascade'`, PG automatically deletes the record of A that pointed at B. But that record of A is still referenced by the record of B, through B's own restrict FK `a_id`. PG refuses to delete the A record while that reference still exists.

**Steps to reproduce:**
- Define this in a custom module `mymodule`.
```py
from odoo import fields, models

class A(models.Model):
    _name = "mymodule.a"
    _description = "mymodule.a"

    model_id = fields.Many2one("ir.model", required=True, ondelete="cascade")

class B(models.Model):
    _name = "mymodule.b"
    _description = "mymodule.b"

    a_id = fields.Many2one("mymodule.a", ondelete="restrict")
```
- after installing `mymodule` run this in odoo shell
```py
env = self.env
a = env["mymodule.a"].create({"model_id": env.ref("mymodule.model_mymodule_b").id})
b = env["mymodule.b"].create({"a_id": a.id})

cr = env.cr
util.remove_model(cr, "mymodule.b")
```
- We get the below error:
```sql
2026-09-09 07:34:41,840 150270 ERROR fk_issue_19 odoo.sql_db: bad query: b'DELETE FROM ir_model WHERE id=124'
ERROR: update or delete on table "mymodule_a" violates foreign key constraint "mymodule_b_a_id_fkey" on table "mymodule_b"
DETAIL:  Key (id)=(1) is still referenced from table "mymodule_b".

  File "/home/odoo/src/odoo/19.0/odoo-bin", line 6, in <module>
    odoo.cli.main()
  File "/home/odoo/src/odoo/19.0/odoo/cli/command.py", line 133, in main
    command().run(args)
  File "/home/odoo/src/odoo/19.0/odoo/cli/shell.py", line 162, in run
    self.shell(dbnames[0])
  File "/home/odoo/src/odoo/19.0/odoo/cli/shell.py", line 148, in shell
    self.console(local_vars)
  File "/home/odoo/src/odoo/19.0/odoo/cli/shell.py", line 82, in console
    exec(sys.stdin.read(), local_vars)
  File "<string>", line 8, in <module>
  File "/home/odoo/src/upgrade/migrations/util/models.py", line 258, in remove_model
    cr.execute("DELETE FROM ir_model WHERE id=%s", (mod_id,))
  File "/home/odoo/src/odoo/19.0/odoo/sql_db.py", line 440, in execute
psycopg2.errors.ForeignKeyViolation: update or delete on table "mymodule_a" violates foreign key constraint "mymodule_b_a_id_fkey" on table "mymodule_b"
DETAIL:  Key (id)=(1) is still referenced from table "mymodule_b".
```

**Note:** The above case is not reproducible if model B inherits from model A, as that case is already handled correctly.

### DIFF
--- a/src/util/models.py
+++ b/src/util/models.py
@@ -86,6 +86,46 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
     chunk_size = 1000
     notify = False
     unk_id = _unknown_model_id(cr)
+    table = table_of_model(cr, model)
+    cr.execute(
+        """
+        SELECT con.conname,
+               att.attname,
+               cl2.relname,
+               fatt.attname
+          FROM pg_constraint con
+          JOIN pg_class cl2
+            ON cl2.oid = con.confrelid
+          JOIN pg_attribute att
+            ON att.attrelid = con.conrelid
+           AND att.attnum = con.conkey[1]
+          JOIN pg_attribute fatt
+            ON fatt.attrelid = con.confrelid
+           AND fatt.attnum = con.confkey[1]
+         WHERE con.conrelid = to_regclass(%s)
+           AND con.contype = 'f'
+           AND con.confdeltype IN ('r', 'a')
+           AND array_length(con.conkey, 1) = 1
+        """,
+        [table],
+    )
+    for conname, column, ref_table, ref_col in cr.fetchall():
+        cr.execute(
+            format_query(
+                cr,
+                """
+                ALTER TABLE {table}
+                ALTER COLUMN {column} DROP NOT NULL,
+                 DROP CONSTRAINT {conname},
+                  ADD CONSTRAINT {conname} FOREIGN KEY ({column}) REFERENCES {ref_table}({ref_col}) ON DELETE SET NULL
+                """,
+                table=table,
+                column=column,
+                conname=conname,
+                ref_table=ref_table,
+                ref_col=ref_col,
+            )
+        )
 
     # remove references
     for ir in indirect_references(cr):


KangOl:

src/util/models.py
```diff
@@ -86,6 +88,34 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+                       att.attname,
+                       cl2.relname
+                  FROM pg_constraint con
+                  JOIN pg_class cl2
+                    ON cl2.oid = con.confrelid
+                  JOIN pg_attribute att
+                    ON att.attrelid = con.conrelid
+                   AND att.attnum = con.conkey[1]
+                 WHERE con.conrelid = '{table}'::regclass
+                   AND con.contype = 'f'
```
KangOl: Inject as argument
```suggestion
                 WHERE con.conrelid = %s::regclass
                   AND con.contype = 'f'
```

src/util/models.py
```diff
@@ -86,6 +88,34 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+                """,
+                table=table,
+            )
+        )
+        for conname, column, ref_table in cr.fetchall():
+            cr.execute(
+                format_query(cr, "ALTER TABLE {table} ALTER COLUMN {column} DROP NOT NULL", table=table, column=column)
+            )
+            remove_constraint(cr, table, conname)
+            create_fk(cr, table, column, ref_table, on_delete_action="SET NULL")
```
KangOl: I would do the recreation in one single query. This will also allow to keep the same constraint name.
```suggestion
            cr.execute(
                format_query(
                    cr, 
                    """
                      ALTER TABLE {table}
                      ALTER COLUMN {column} DROP NOT NULL,
                      DROP CONSTRAINT {conname},
                      ADD CONSTRAINT {conname} FOREIGN KEY ({column}) REFERENCES {ref_table}(id) ON DELETE SET NULL
                     """,
                     table=table,
                     column=column,
                     conname=conname,
                     ref_table=ref_table,
                )
            )
```

apan-odoo: I thought of doing the same, but decided to use the utils. Anyway, I’ve pushed the suggestions. Thanks!

src/util/models.py
```diff
@@ -86,6 +86,44 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
     chunk_size = 1000
     notify = False
     unk_id = _unknown_model_id(cr)
+    table = table_of_model(cr, model)
+    if table_exists(cr, table):
```
aj-fuentes:
```suggestion
```
Use `to_regclass(%s)` instead `%s::regclass` below. It returns NULL if the table doesn't exist.

src/util/models.py
```diff
@@ -86,6 +86,44 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+        cr.execute(
+            format_query(
```
KangOl, line 114: Not needed anymore.
```suggestion
```

src/util/models.py
apan-odoo: Didn't know about that :) Thanks!

src/util/models.py
```diff
@@ -86,6 +86,40 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+               cl2.relname
+          FROM pg_constraint con
+          JOIN pg_class cl2
+            ON cl2.oid = con.confrelid
+          JOIN pg_attribute att
+            ON att.attrelid = con.conrelid
+           AND att.attnum = con.conkey[1]
+         WHERE con.conrelid = to_regclass(%s)
+           AND con.contype = 'f'
+           AND con.confdeltype = 'r'
```
aj-fuentes:
```suggestion
           AND con.confdeltype IN ('r', 'a')
```
https://www.postgresql.org/docs/18/sql-createtable.html#SQL-CREATETABLE-PARMS-REFERENCES-REFACT-NO-ACTION

src/util/models.py
```diff
@@ -86,6 +86,40 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
     chunk_size = 1000
     notify = False
     unk_id = _unknown_model_id(cr)
+    table = table_of_model(cr, model)
+    cr.execute(
+        """
+        SELECT con.conname,
+               att.attname,
+               cl2.relname
```
KangOl:
```suggestion
               cl2.relname,
               fatt.attname
```

src/util/models.py
```diff
@@ -86,6 +86,40 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+          FROM pg_constraint con
+          JOIN pg_class cl2
+            ON cl2.oid = con.confrelid
+          JOIN pg_attribute att
+            ON att.attrelid = con.conrelid
+           AND att.attnum = con.conkey[1]
```
KangOl, line 101:
```suggestion
           AND att.attnum = con.conkey[1]
          JOIN pg_attribute fatt
            ON fatt.attrelid = con.confrelid
           AND fatt.attnum = con.confkey[1]
```

src/util/models.py
```diff
@@ -86,6 +86,40 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+         WHERE con.conrelid = to_regclass(%s)
+           AND con.contype = 'f'
+           AND con.confdeltype = 'r'
+        """,
+        [table],
+    )
+    for conname, column, ref_table in cr.fetchall():
```
KangOl:
```suggestion
    for conname, column, ref_table, ref_col in cr.fetchall():
```

src/util/models.py
```diff
@@ -86,6 +86,40 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+        cr.execute(
+            format_query(
+                cr,
+                """
+                ALTER TABLE {table}
+                ALTER COLUMN {column} DROP NOT NULL,
+                 DROP CONSTRAINT {conname},
+                  ADD CONSTRAINT {conname} FOREIGN KEY ({column}) REFERENCES {ref_table}(id) ON DELETE SET NULL
```
KangOl:
```suggestion
                  ADD CONSTRAINT {conname} FOREIGN KEY ({column}) REFERENCES {ref_table}({ref_col}) ON DELETE SET NULL
```

src/util/models.py
```diff
@@ -86,6 +86,40 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+                """,
+                table=table,
+                column=column,
+                conname=conname,
+                ref_table=ref_table,
```
KangOl, line 125:
```suggestion
                ref_table=ref_table,
                ref_col=ref_col,
```

KangOl (changes requested):
Don't hardcode `id` as the column on the referenced table.

src/util/models.py
```diff
@@ -86,6 +86,45 @@ def remove_model(cr, model, drop_table=True, ignore_m2m=()):
...
+            ON cl2.oid = con.confrelid
+          JOIN pg_attribute att
+            ON att.attrelid = con.conrelid
+           AND att.attnum = con.conkey[1]
+          JOIN pg_attribute fatt
+            ON fatt.attrelid = con.confrelid
+           AND fatt.attnum = con.confkey[1]
+         WHERE con.conrelid = to_regclass(%s)
+           AND con.contype = 'f'
+           AND con.confdeltype IN ('r', 'a')
```
aj-fuentes, line 107: To be aligned with the re-added FK. We want single column FKs -- anything custom-made we let it as is.
```suggestion
           AND con.confdeltype IN ('r', 'a')
           AND array_length(con.conkey, 1) = 1
```

apan-odoo: I was checking `get_fk` w.r.t the above suggestions. Was the [aim of](https://github.com/odoo/upgrade-util/blob/bcb57d01dcca86d27a6e9639eacab6cbe9ad93ed/src/util/pg.py#L1061-L1062):
```sql
              AND array_lower(con.conkey, 1) = 1
              AND array_lower(con.confkey, 1) = 1
```
also to check for single column FKs? The current condition only excludes `NULL` values and still considers FKs with more than one column.

Indeed, your suggestion of using `array_length` for single column FKs is correct.

aj-fuentes: It is an old bug. Already discussed in DM.

aj-fuentes (approved):

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://piaille.fr/@CatEveryHour/117245686868105210

PR #513, opened 2026-09-09
[IMP] pg: add explode_on kwarg to explode_{query_range,execute}

cawo-odoo:
This will allow for eliminating unnecessary and hurtful joins on a FK when the join is only used to add a filter on the joined table. Instead, it is now possible to explode on the FK column.


KangOl:
upgradeci retry with always only stock

src/util/pg.py
```diff
@@ -369,14 +369,14 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
         sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
         query += sep_kw + "{parallel_filter}"
 
-    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
+    cr.execute(format_query(cr, "SELECT min({id_col}), max({id_col}) FROM {table}", id_col=id_col, table=table))
     min_id, max_id = cr.fetchone()
     if min_id is None:
         # empty table
         if on_CI():
             # Even if there are any records, return one query to be executed to validate its correctness and avoid
```
```suggestion
            # Even if there aren't any records, return one query to be executed to validate its correctness and avoid
```

KangOl:
Bikesheddding: I don't like `id` being in the parameter name. I would either go with `split_on="id"` or `explode_on="id"`, or maybe just `on="id"`.

Anyway, this new parameter should be documented in the docstring.

KangOl:
Also, a test should be added.

src/util/pg.py
```diff
@@ -417,18 +418,25 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
         # only two buckets and the second would have at most 10% of bucket_size records.
         # Still, since the query may only be valid if there is no split, we force the usage of `prefix` in the query to
         # validate its correctness and avoid scripts that pass the CI but fail in production.
-        parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
+        parallel_filter = format_query(cr, "{}.{} IS NOT NULL", alias, id_col)
```
jjmaksoud: using format_query here with the alias will add the double quote identifiers, making the alias case sensitive, in the failing CI the query uses the alias `A` unquoted so it defaults to a lower case `a` and then the parallel filter clause alias does not match

cawo-odoo:
What's happening? :-D This *Draft* is fully untested (and docstring and test are on the todo list), I just wanted to let the CI run once. 

Anyway, thanks for the reviews ;-)

src/base/tests/test_util.py
```diff
@@ -1011,12 +1011,23 @@ def test_explode_query_range(self):
         count += 1
         cr.execute("UPDATE res_partner_category SET id = 10000000 WHERE id = %s", [tid])
 
+        for i in range(3):
+            cr.execute("UPDATE res_partner SET country_id={id} WHERE id={id}".format(id=i + 1))
```
KangOl: What?
1/ `.format` usage ???
2/ This is fragile. Don't hardcode record ids.

src/util/pg.py
```diff
@@ -379,14 +379,16 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
...
+    )
     min_id, max_id = cr.fetchone()
     if min_id is None:
         # empty table
         if on_CI():
-            # Even if there are any records, return one query to be executed to validate its correctness and avoid
+            # Even if there are no records, return one query to be executed to validate its correctness and avoid
             # scripts that pass the CI but fail in production.
-            parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
+            parallel_filter = "{alias}.{explode_on} IS NOT NULL".format(alias=alias, explode_on=explode_on)
```
KangOl: Add a comment about why we still use `.format` here.

KangOl: Actually, we should use `.format` only for `alias`. `explode_on` should be injected by `format_query`.

src/util/pg.py
```diff
@@ -401,15 +403,16 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
                 cr,
                 """
                 WITH t AS (
-                    SELECT id,
-                           mod(row_number() OVER(ORDER BY id) - 1, %s) AS g
+                    SELECT {explode_on},
+                           mod(row_number() OVER(ORDER BY {explode_on}) - 1, %s) AS g
```
KangOl: Does it still works as expected now that `explode_on` is not necessarily unique?

src/util/pg.py
KangOl: Allowing usage of `format_query`
```suggestion
    alias = SQLStr(alias) if alias else table
```

src/util/pg.py
aj-fuentes: Doesn't look correct but `dense_rank` may be what we need here.
```suggestion
                           mod(dense_rank() OVER(ORDER BY {explode_on}) - 1, %s) AS g
```
```sql
=> SELECT id,
        row_number() OVER (ORDER BY id) ,
        dense_rank() OVER (ORDER BY id)
   FROM (VALUES (1), (1), (2)) a(id)
+----+------------+------------+
| id | row_number | dense_rank |
|----+------------+------------|
| 1  | 1          | 1          |
| 1  | 2          | 1          |
| 2  | 3          | 2          |
+----+------------+------------+
```

aj-fuentes: > dense_rank () → bigint
>
> Returns the rank of the current row, without gaps; this function effectively counts peer groups.

src/util/pg.py
cawo-odoo: I added the SQLStr, but only after enforcing it contains only word characters.

src/util/pg.py
cawo-odoo: Good catch, thanks. The use-case I had in mind did still guarantee uniqueness, but it's better to not assume.

aj-fuentes: I think we need a test that illustrates this case --fails with row_number and passes with dense_rank.

src/util/pg.py
```diff
@@ -370,6 +370,10 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
         alias = prefix.rstrip(".")
 
     alias = alias or table
+    if re.search(r"\W", alias, re.ASCII):
+        raise ValueError("The `alias` argument can only contain ASCII word characters")
+    # for backwards compatibility, `alias` can not be quoted by `format_query` to keep it case-insensitive
+    alias = SQLStr(alias)
```
KangOl: No. It should only be done if passed as argument, not when it is the fallback `table`

src/util/pg.py
```diff
@@ -379,14 +383,21 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
         sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
         query += sep_kw + "{parallel_filter}"
 
-    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
+    cr.execute(
+        format_query(cr, "SELECT min({explode_on}), max({explode_on}) FROM {table}", explode_on=explode_on, table=table)
```
KangOl, line 386:
```suggestion
        format_query(cr, "SELECT min({explode_on}), max({explode_on}) FROM {table} WHERE {explode_on} IS NOT NULL", explode_on=explode_on, table=table)
```

src/util/pg.py
```diff
@@ -379,14 +383,21 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
...
     if min_id is None:
         # empty table
         if on_CI():
-            # Even if there are any records, return one query to be executed to validate its correctness and avoid
+            # Even if there are no records, return one query to be executed to validate its correctness and avoid
             # scripts that pass the CI but fail in production.
-            parallel_filter = "{alias}.id IS NOT NULL".format(alias=alias)
+            parallel_filter = format_query(
+                cr,
+                "{alias}.{explode_on} IS NOT NULL",
```
KangOl, line 396: We want an always-false condition.
```suggestion
                "({alias}.{explode_on} IS NOT NULL AND false)",
```

src/util/pg.py
```diff
@@ -379,14 +383,21 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
         sep_kw = " AND " if re.search(r"\sWHERE\s", query, re.M | re.I) else " WHERE "
         query += sep_kw + "{parallel_filter}"
 
-    cr.execute(format_query(cr, "SELECT min(id), max(id) FROM {}", table))
+    cr.execute(
+        format_query(cr, "SELECT min({explode_on}), max({explode_on}) FROM {table}", explode_on=explode_on, table=table)
+    )
     min_id, max_id = cr.fetchone()
     if min_id is None:
         # empty table
```
```suggestion
        # no matching rows.
```

src/util/pg.py
cawo-odoo: You mean because `table` has been used in format_query already before? OK, then.

src/util/pg.py
cawo-odoo: IDGI. What is it supposed to change?

KangOl: Nothing, indeed.

src/util/pg.py
```diff
@@ -401,15 +411,16 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
...
-                    SELECT id,
-                           mod(row_number() OVER(ORDER BY id) - 1, %s) AS g
+                    SELECT {explode_on},
+                           mod(dense_rank() OVER(ORDER BY {explode_on}) - 1, %s) AS g
                       FROM {table}
-                     ORDER BY id
-                ) SELECT array_agg(id ORDER BY id) FILTER (WHERE g=0),
-                         min(id),
-                         max(id)
+                     ORDER BY {explode_on}
```
```suggestion
                        WHERE {explode_on} IS NOT NULL
                     ORDER BY {explode_on}
```

cawo-odoo: yes, I arrived at this a few min ago while working on a test case, but it's not there, yet.

src/util/pg.py
```diff
@@ -401,15 +411,17 @@ def explode_query_range(cr, query, table, alias=None, bucket_size=DEFAULT_BUCKET
...
+                    SELECT {explode_on},
+                           mod(dense_rank() OVER(ORDER BY {explode_on}) - 1, %s) AS g
                       FROM {table}
-                     ORDER BY id
-                ) SELECT array_agg(id ORDER BY id) FILTER (WHERE g=0),
-                         min(id),
-                         max(id)
+                     WHERE {explode_on} IS NOT NULL
+                     ORDER BY {explode_on}
+                ) SELECT array_agg({explode_on} ORDER BY {explode_on}) FILTER (WHERE g=0),
```
aj-fuentes: Don't we need this?
```suggestion
                ) SELECT array_agg(DISTINCT {explode_on} ORDER BY {explode_on}) FILTER (WHERE g=0),
```
Following this example:
```
=> SELECT id,
        row_number() OVER (ORDER BY id),
        dense_rank() OVER (ORDER BY id)
   FROM (VALUES (1), (2), (2)) a(id)
+----+------------+------------+
| id | row_number | dense_rank |
|----+------------+------------|
| 1  | 1          | 1          |
| 2  | 2          | 2          |
| 2  | 3          | 2          |
+----+------------+------------+
```
if bucket size=2, we get 2 % 2 = 0 for the last two rows.

KangOl: Yes.

aj-fuentes (approved):

KangOl (approved):
@robodoo r+

[^source]

[^source]: https://app.wafrn.net/fediverse/post/01a0aa73-c051-719a-bf67-2717506e2454

xavierbol:
robodoo priority

(I don't know if I can but this PR blocked my key feature 😇 )

PR #515, opened 2026-09-16
[FIX] base/17.0: sanitize embedded \n in attrs before ast.parse

bsra-odoo:
Causing SyntaxError while parsing `attrs` values in attr_domains2expr.py:

    File "attr_domains2expr.py", line 336, in ast_parse
        return ast.parse(val.strip(), mode="eval").body
    File "<unknown>", line 1
        {"invisible": ["&","&",[...],["x_studio_source_of_booking","=","Online
    SyntaxError: unterminated string literal

Root cause: the stored attrs string contained a raw/literal newline character inside a quoted string value (e.g. "Online \nMessaging Platforms"). Python string literals cannot contain unescaped newlines, so ast.parse() failed [here](https://github.com/odoo/upgrade-util/blame/c67ac262b546c16937ba20c187cec92d6655a384/src/base/17.0.1.3/attr_domains2expr.py#L336) before the domain could even be evaluated.

attrs is meant to hold conditional domain expressions (short field/operator/value tokens), not free text. A raw newline inside it has no valid use case — it's corrupted data

Faulty node example from customer DB:
```
{"invisible": ["&","&",["x_studio_source_of_booking_2","=","Onsite"],["x_studio_source_of_booking_2","=","Online\n            Messaging Platforms"],["x_studio_source_of_booking_2","=","Abenson Stores"]], "readonly": [["state","not in",["draft","sent"]]]}
```

**Shell output:**
```python
In [13]: import ast

In [14]: attrs = '{"invisible": ["&","&",["x_studio_source_of_booking_2","=","Onsite"],["x_studio_source_of_booking_2","=","Online\n            Messaging Platforms"],["x_studio_source_
    ...: of_booking_2","=","Abenson Stores"]], "readonly": [["state","not in",["draft","sent"]]]}'
    ...: 

In [15]: ast.parse(attrs.strip(), mode="eval").body

  File "/usr/lib/python3/dist-packages/IPython/core/interactiveshell.py", line 3457, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  File "<ipython-input-15-dce14de5d456>", line 1, in <module>
    ast.parse(attrs.strip(), mode="eval").body

  File "/usr/lib/python3.10/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,

  File "<unknown>", line 1
    {"invisible": ["&","&",["x_studio_source_of_booking_2","=","Onsite"],["x_studio_source_of_booking_2","=","Online
                                                                                                             ^
SyntaxError: unterminated string literal (detected at line 1)
```

Found more req with similar issue: [3705846](https://upgrade.odoo.com/odoo/request/3705846), [2453310](https://upgrade.odoo.com/odoo/request/2453310)

- OPW: 6558799


aj-fuentes:
The "solution" here looks wrong. We cannot just replace like that.
You need to (in order):
1. Check how such string could be saved in `attrs`. If we cannot reproduce this in the frontend then it means that something was edited by other means. So the error is valid and should be addressed specifically.
2. When posting the root issue ensure you post the raw arch data from the table column (shortened to the interesting part). To be sure we clearly see what was set.

`parse` can fail in many ways, trying to make it never fail is not a good strategy. The string literal in the failing case is clearly wrong.

bsra-odoo:
> The "solution" here looks wrong. We cannot just replace like that. You need to (in order):
> 
> 1. Check how such string could be saved in `attrs`. If we cannot reproduce this in the frontend then it means that something was edited by other means. So the error is valid and should be addressed specifically.
> 2. When posting the root issue ensure you post the raw arch data from the table column (shortened to the interesting part). To be sure we clearly see what was set.
> 
> `parse` can fail in many ways, trying to make it never fail is not a good strategy. The string literal in the failing case is clearly wrong.

Yes, you are right that string literals can fail for many reasons. my thought was to perform the replacement only for the attrs case, since attrs generally contains conditions and domains. Having \n in domains is not a valid case. so The idea is to avoid the failure in a case that we already know can fail.


>Check how such string could be saved in attrs. If we cannot reproduce this in the frontend then it means that something was >edited by other means. So the error is valid and should be addressed specifically.

If we manually add a line break in the view from the UI, this case is triggered. I will check further in versions <= v15 to see if we can reproduce it using Studio itself.

Thank you for the review and input @aj-fuentes

aj-fuentes:
> my thought was to perform the replacement only for the attrs case, since attrs generally contains conditions and domains. Having \n in domains is not a valid case. so The idea is to avoid the failure in a case that we already know can fail.

So let them fail. Replacing newline in something like `'1\n2'` would give `'12'` which has a very different meaning syntaxtically. I don't think there is a safe way to replace those. They are truly a valid failure. IDK how the JS side handled it, probably the whole value was ignored.

bsra-odoo:
> > my thought was to perform the replacement only for the attrs case, since attrs generally contains conditions and domains. Having \n in domains is not a valid case. so The idea is to avoid the failure in a case that we already know can fail.
> 
> So let them fail. Replacing newline in something like `'1\n2'` would give `'12'` which has a very different meaning syntaxtically. I don't think there is a safe way to replace those. They are truly a valid failure. IDK how the JS side handled it, probably the whole value was ignored.

got it. thank you :)  @aj-fuentes

PR #516, opened 2026-09-17
[FIX] util/pg: fix module not found re.ASCII Python 2

apan-odoo:
`re.ASCII` doesn't exist in Python 2's `re` module, breaking matt running for `v9.0` -> `v10.0`

failing matt: https://upgradeci.odoo.com/upgradeci/run/358094


aj-fuentes (approved):
@robodoo r+

KangOl:

PR #517, opened 2026-09-21
[ADD] util/specific.py: new helper to update cowed view

hpr-odoo:
- every time when update the Cowed View, we need to write a query to have the ids
- after this PR, we only need to provide the key/s
- the function will then find the cowed views and modify based on callback
- you can also give the extra filters to narrow down the search results
- ex:
```py
In [1]: def update_cowed_view(arch):
   ...:     arch.insert(0, etree.Element("new"))
   ...:

In [3]: util.process_cowed_views(cr, "website.contactus", update_cowed_view)
Out[3]: {4328}

In [4]: util.process_cowed_views(cr, "website.contactus", update_cowed_view, "id != 4328")
Out[4]: set()

In [5]: util.process_cowed_views(cr, "website.contactus", update_cowed_view, "arch_db->>'en_US' ilike '%%contact%%'")
Out[5]: {4328}

In [6]: util.process_cowed_views(cr, ["website.record_cover", "website.footer_custom"], update_cowed_view)
Out[6]: {3713, 4788}
```


KangOl:

src/util/specific.py
```diff
@@ -163,6 +165,29 @@ def rename_custom_column(cr, table_name, col_name, new_col_name, custom_module=N
...
+    if isinstance(key, str):
+        key = [key]
+    cr.execute(
+        """
+        SELECT id
+          FROM ir_ui_view
+         WHERE key IN %s
+           AND website_id IS NOT NULL
+           AND ({})
+        """.format(extra_where),
```
KangOl: Don't use `.format()` on queries.

src/util/specific.py
```diff
@@ -163,6 +165,29 @@ def rename_custom_column(cr, table_name, col_name, new_col_name, custom_module=N
...
+        [tuple(key)],
+    )
+    views = cr.fetchall()
+    for (view_id,) in views:
+        with skippable_cm(), edit_view(cr, view_id=view_id) as arch:
```
KangOl: Why `skippable_cm`?

hpr-odoo: > Why `skippable_cm`?

I get it. That was to handle the ContextManager, but due to the iteration on results, if the view isn't found, it won't be there :saluting_face:

src/util/specific.py
hpr-odoo: done

PR #519, opened 2026-09-23
[FIX] util/records: handle t-call tail on view removal

aj-fuentes:
Removing a stale t-call node also drops its tail -- the text that follows
it. When the tail is the whole body of a `t-else`, the branch is left
empty and rendering the view fails with an `IndentationError`.

Even if no error is raised dropping the tail content is wrong.


aj-fuentes:
Example rendering error:
```
  File "/home/odoo/src/odoo/20.0/odoo/addons/base/models/ir_qweb.py", line 891, in _render_iterall
    raise QWebError(qweb_error_info) from error
odoo.addons.base.models.ir_qweb.QWebError: Error while rendering the template:
    IndentationError: expected an indented block after 'else' statement on line 21 (<526>, line 23)
    Template: 526
    Reference: 526
```

aj-fuentes:
upgradeci retry with always website in versions 18.0 19.0 20.0

